PR #60, opened 2016-07-23
Don't require server.sh to be launched from standalone directory

bagage:
Instead of doing `cd standalone/ && sh server.sh`, one can directly do `sh standalone/server.sh`.


bagage:
Up?

Phyks:
@abrensch This should be fully backward compatible and ease a lot getting started with BRouter :)

bagage:
@abrensch could it be merged please?

PR #72, opened 2016-12-02
handling url encoded parameters

markshiz:


nrenner:
Thanks for the PR!

These separators ("|" and ",") were my mistake; what I really wanted was a human-readable URL that doesn't need to be encoded. But I just copied what some other service was using without checking the spec. This has been bugging me since, because the unencoded URL doesn't work everywhere, e.g. with link detection in mails, and obviously your use-case.

So my suggestion would rather be to introduce new separators that don't need encoding, while keeping the old ones for backwards compatibility.

One idea might be using "_" and "~", which are [unreserved characters](https://tools.ietf.org/html/rfc3986#page-13) and don't get encoded in JavaScript encodeURI/encodeURIComponent, and add those as alternative delimiters to the split call. But I would want to do further research and check again what other services are using.

What is your use-case? Would changing your URL to use "_"/"~" instead of "|"/"," work for you?

markshiz:
In my case, I'm connecting to an iOS app that forces proper URI encoding.  This patch seemed to work sufficiently for my needs.  I will admit, the pipe separator did make me cringe a little, but it's a minor annoyance, really. :smile:

As far as array data in a query string is concerned, I think there are a number of valid ways to go.  One of them seems to be commas --which of course, still need to be encoded.  See this posting for some other ideas: http://stackoverflow.com/a/11890080

nrenner:
> First off, thanks for open sourcing this project, it's a great resource!

Just to clarify, this goes to @abrensch, the owner of this project.

Thanks for your comment. Agree that there would have been better options, but not sure I would want to change the parameter structure while still supporting the old one.

I made a [comparison of URL formats](https://github.com/nrenner/brouter-web/wiki/Route-URL-comparison) used by some other routing services. The only common thing seems to be using "," as lat,lon separator, so I guess we will stick with that.

And while "," does not need to be encoded, it may still be, e.g. with form-encoding, so my idea of introducing characters that always will be unencoded instead of handling encoding is obsolete. Therefore I'm fine with applying your fix now and thinking about further improvements later.

markshiz:
Thanks for the analysis of the URL formats @nrenner.  Interesting to see how other people approach the problem, for sure.

I agree, I have no qualms with applying the patch as is.  It seems like an easy win for now.

PR #75, opened 2017-03-29
NegativeArraySizeException during map creation

Mincol:
I do not thing this is cure, but during map creation I got

  Exception in thread "main" java.lang.NegativeArraySizeException
        at btools.mapcreator.NodeData.<init>(NodeData.java:32)

which is strange, because array length (description.length) cannot be negative. So either description length was > 127 or ... ?

This pull request contains sanity check during serialization so description lengt cannot be bigger than 1 byte.
Also changes data format. Patch changes signed description lenght to unsigned byte. This change MAY be backward compatitble, but I did not bothered to check. I appologize for that.

### DIFF
--- a/brouter-map-creator/src/main/java/btools/mapcreator/NodeData.java
+++ b/brouter-map-creator/src/main/java/btools/mapcreator/NodeData.java
@@ -29,7 +29,7 @@ public class NodeData extends MapCreatorBase
     ilon = (int)dis.readDiffed( 1 );
     ilat = (int)dis.readDiffed( 2 );
     int mode = dis.readByte();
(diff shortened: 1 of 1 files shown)


abrensch:
Should be save to just change byte -> short here. Backward compatibility is no issue here, because this serialisation format is just temporary during map-processing.

Later in the pipe there are either no such length prefixes ( in rd5 files ) or variable encoding for the prefix (in intermediate serialisation format after decoding md5).

Did you encounter that with the original lookups.dat or an extended one? If original, I should get that as well when processing next planet?

I will hav a look shortly - thanks.

Mincol:
No I do thing it is needed to be merged before next map run.

PR #86, opened 2017-10-07
Added flag avoid_tolls = true

poutnikl:
I have created the new profile **car-tollfree**, based on your car-fast profile. It may not be so compelling in DE, but for many drivers/countries is option to avoid toll roads important.

Global context : Added flag avoid_tolls = true
Way context_Sets CF to 10000 if toll=yes
Node context: Sets initialcost to 1000000 if barrier=toll_booth

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ target/
 .classpath
 .project
 *.class
+.idea/
+*.iml
--- a/README.md
+++ b/README.md
@@ -1,16 +1,21 @@
-brouter
+BRouter
 =======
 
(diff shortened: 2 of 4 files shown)


poutnikl:
It may be worthy to change CF 10000 to 9999 to avoid navigation hint issues. e.g at the highway exit before the toll highway section.

poutnikl:
I originally have put the toll stuff into a separate car-tollfree profile, but you may want to consider to include it into car-fast and car-eco profiles as well, with avoid_toll = false. I was thinking about a pull request for that, but by the second thought I let it on you.

poutnikl:
I have messed up with updates of this fork, so I close it.

PR #97, opened 2018-02-26
Nogo polygons / polylines

ntruchsess:
This pull-request adds support for nogo-polygons and polylines. A nogo-polygon is being used as a nogo-area being defined by an (arbitrary shaped) polygon. A nogo-polyline is simmilar to this with the difference that the polygon is not taken as closed, so it defines a border that is not to be crossed when routing.
I currently can be utilized via the standalone server (new query-parameters 'polygons' and 'polylines') and on Android from OsmAnd (by saving tracks with the filename beginning with 'nogo').
(The web-frontend does not yet support this, but QMapshack already has full support for this feature).

### DIFF
--- a/brouter-core/pom.xml
+++ b/brouter-core/pom.xml
@@ -32,5 +32,10 @@
             <artifactId>brouter-expressions</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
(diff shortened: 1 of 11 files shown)


ntruchsess:
see
https://bitbucket.org/maproom/qmapshack/wiki/AdvRoutes#markdown-header-using-no-go-areas-and-lines
on how this works in qmapshack

PR #107, opened 2018-06-28
Fix typo

saraedum:
this doesn't really matter, but I just stumbled upon this one so I thought I can as well propose a fix…


Phyks:
@abrensch Could you take this one? Thanks!

PR #117, opened 2018-10-30
Allow custom nogo weights

Phyks:
Hi,

This PR should allow user to specify custom weights/penalty for each nogo circle. The motivation behind this is that so far the nogos are an all or nothing feature whereas they could potentially offer more control to the user by slightly biasing some areas.

Typically, in France, a lot of cities provide OpenData about road works (a full list of the files I found so far is available at https://framagit.org/phyks/cyclassist/blob/34e0fec0165d3633e1d254f360fcd8c2cdfac78c/scripts/opendata/works.py#L304-389). Some road works might be a complete road closure (nogo as currently implemented) while other might just mean having to deal with some extra pain (only some lanes closed, traffic jam expected etc). The latter ones should be avoided if possible, but not at all cost.

I tried to take care of making it retro-compatible. Please note however that this is my first contact with BRouter code (and Java!) so it might be far from being perfect. If you think this is worth a merge, I'll gladly take any review and fix it to make it mergeable!

The code was tested with a modified version of brouter-web, so it should be working fine.

Thanks!

### DIFF
--- a/brouter-core/src/main/java/btools/router/OsmNodeNamed.java
+++ b/brouter-core/src/main/java/btools/router/OsmNodeNamed.java
@@ -28,9 +28,8 @@ public class OsmNodeNamed extends OsmNode
   public double distanceWithinRadius(int lon1, int lat1, int lon2, int lat2, double totalSegmentLength) {
     double[] lonlat2m = CheapRulerSingleton.getLonLatToMeterScales( (lat1 + lat2) >> 1 );
 
-    double realRadius = radius * 110984.;
-    boolean isFirstPointWithinCircle = CheapRulerSingleton.distance(lon1, lat1, ilon, ilat) < realRadius;
-    boolean isLastPointWithinCircle = CheapRulerSingleton.distance(lon2, lat2, ilon, ilat) < realRadius;
+    boolean isFirstPointWithinCircle = CheapRulerSingleton.distance(lon1, lat1, ilon, ilat) < radius;
+    boolean isLastPointWithinCircle = CheapRulerSingleton.distance(lon2, lat2, ilon, ilat) < radius;
     // First point is within the circle
     if (isFirstPointWithinCircle) {
       // Last point is within the circle
@@ -62,7 +61,7 @@ public class OsmNodeNamed extends OsmNode
     double initialToCenter = CheapRulerSingleton.distance(ilon, ilat, lon1, lat1);
     // Half length of the segment within the circle
     double halfDistanceWithin = Math.sqrt(
-      realRadius*realRadius - (
+      radius*radius - (
         initialToCenter*initialToCenter -
         initialToProject*initialToProject
       )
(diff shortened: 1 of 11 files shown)


Phyks:
Hi,

Thanks for the review!

> the nogo-weigth is added to the path-cost for every "link" that hits the nogo. A link is a part of a way that connects two "non-trivial-network nodes". The partitioning of real-life "ways" into links is completely arbitrary, and that should not have an effect on the calculated cost of a path. If it has an effect, you get strange artefacts.

Indeed, thanks for catching this! I think with the original motivation in mind (avoiding work areas for instance), it would make more sense to have a cost-penalty proportional to the distance travelled inside the nogo. This would also allow to cross a way with nogo areas along it (way under work) without penalty (while taking the way would mean crossing the nogo and account for a huge penalty).

I'll try to rework it like this :)


> The other objection I have is more technical: using 100k as the "no not ever go" value is fine, but it should still result in a cost=-1 path, because only cost=-1 paths are discarded directly by the routing engine and not feeded into the priority queue. Have too many high-cost-paths in the priority queue would have a memory impact

Oh, ok, I missed this special case with a cost=-1. I just took 100k as it seemed to mean "discarded" in the code. I'll rework this!

Phyks:
Cost of `-1` should be fixed at the moment. I'll work on computing the cost penalty properly based on distance soon.

Phyks:
This one should be ready for review. This one or https://github.com/abrensch/brouter/pull/130 should be updated according to which one is the first one to get merged.

I included the computation of distance within the polygon / nogo circle in the last commit and so far, it seems to work nice. I pushed some unittests for this as coding geometry stuff can get tricky :)

Best,

brouter-core/src/main/java/btools/router/OsmNodeNamed.java
```diff
@@ -6,27 +6,89 @@
...
+    } else {
+      return ilon + "," + ilat + "," + name + "," + nogoWeight;
+    }
   }
-  
+
+  public double distanceWithinRadius(int lon1, int lat1, int lon2, int lat2, double totalSegmentLength) {
+    double[] lonlat2m = CheapRulerSingleton.getLonLatToMeterScales( (lat1 + lat2) >> 1 );
+
+    double realRadius = radius * 110984.;
```
Phyks: This should be updated depending whether https://github.com/abrensch/brouter/pull/130 gets merged before. Same below.

brouter-core/src/main/java/btools/router/OsmPath.java
```diff
@@ -143,7 +143,7 @@ protected void addAddionalPenalty(OsmTrack refTrack, boolean detailMode, OsmPath
 
     boolean recordTransferNodes = detailMode || rc.countTraffic;
 
-    rc.nogomatch = false;
+    rc.nogoCost = 0.;
```
Phyks: Instead of a boolean `nogomatch`, I'm now storing the cost due to crossing nogos instead. The previous `nogoMatch` can then be inferred by this variable as `Math.abs(rc.nogoCost) > 0`.

Phyks:
Thanks for the merge!

PR #120, opened 2018-11-08
Complete doc for map creation

Phyks:
Hi,

I was having a look at map creation and the best way to create segments on one's own and felt like this part could use a bit more documentation (especially the `pbfparser` building process which took me some time to figure out).

I reworked a bit the `pbfparser` README with a bit more detailed instructions.

Also, there seems to be a missing argument to `OsmCutter` in the `process_pbf_planet.sh` script.

That being said, the `OsmCutter` line in this script works perfectly fine for me with `bzip -dc *.osm.b2 | java …` but it does not extract any ways (nor gives any error) when I provide it with a pbf extract from https://download.geofabrik.de/. Am I missing something in the `pbfparser` build process?

Best,

### DIFF
--- a/misc/pbfparser/README.txt
+++ b/misc/pbfparser/README.txt
@@ -6,12 +6,18 @@ done in the process_pbf_planet.sh script included in
 the git-repo), you have to build yourself the "pbfparser.jar"
 by doing the following:
 
--> get osmosis.jar
--> get protobuf.jar
--> copy the brouter...with_dependencies.jar and name it brouter.jar
+-> get osmosis from https://bretth.dev.openstreetmap.org/osmosis-build/osmosis-latest.zip
+-> copy lib/default/osmosis-osm-binary-*.jar in the archive to osmosis.jar in
+this folder
+-> copy lib/default/protobuf-java-*.jar in the archive to protobuf.jar in this
+folder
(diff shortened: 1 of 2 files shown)


abrensch:
Thanx, and sorry for the messed-up script...

The answer to your question is:

The pbf-parsing variant of the "OmsParser" class (the one you compiled and packed into pbfparser.jar) cannot read from stdin, it needs a file. The filename is expeced as an additional argument to the call to OsmCutter. Here's the line from the production-script:

java -Xmx2600m -Xms2600m -Xmn32m -cp ../pbfparser.jar:../brouter.jar btools.mapcreator.OsmCutter ../lookups.dat nodetiles ways.dat relations.dat restrictions.dat ../all.brf ../planet-latest.osm.pbf

Please note that for rd5-creation you also need SRTM Data files. The script will not fail if you don't have any, but your rd5's will miss elevation. Here's the corresponding production line with the path to my srmt-data:

java -Xmx2600M -Xms2600M -Xmn32M -cp ../brouter.jar -Ddeletetmpfiles=true -DuseDenseMaps=true btools.mapcreator.PosUnifier nodes55 unodes55 bordernids.dat bordernodes.dat ../../srtm3

All flavor's of 90m SRTM should work, but only the one's from CGIAR will gibe you the same results...

PR #121, opened 2018-11-08
Better process_pbf_planet.sh script

Phyks:
Hi again,

Thanks for merging https://github.com/abrensch/brouter/pull/120! I think I got the issue and the `jar` command to build the `pbfparser.jar` was not working actually I think. That's why it was not working with pbf files I think.

I updated the scripts a bit to ease building segments:

* I defined variables in the script to help change the locations more easily for users.
* I added a doc about the SRTM data sources and which files to use.
* I fixed a few typos (STRM vs SRTM).

The script should be working the same way, assuming it is running from the `misc/scripts/mapcreation` folder.

I tested it on a pbf geographical extract from Geofabrik and it works very well except for a few questions I have left:

1. I don't get the `car-traffic_analysis.brf` lines. What are they for (are they actually generating some output or just here for debugging)? It's using coordinates which are outside my geographical extract, so it's failing at the moment :)

2. Why is the `-DuseDenseMaps=true btools.mapcreator.WayLinker unodes55 waytiles55 bordernodes.dat restrictions.dat ../lookups.dat ../all.brf segments rd5` duplicated, once before the car traffic analysis and once after? (is it expected or is it a bad copy/paste or some other error)

Thanks!


abrensch:
Hi Phyks,

thanx for thsi PR.

for your questions: this is a "car-traffic-simulation". It computes traffic estimates for some 5*5 squares (not the entire planet), and the second call to "WayLinker" re-creates these rd5's including the "estimated_traffic_class" tags. Only the profile "fastbike-lowtraffic" is using these tags.

Phyks:
Hi,

Thanks for the feedback and the merge!

> for your questions: this is a "car-traffic-simulation". It computes traffic estimates for some 5*5 squares (not the entire planet), and the second call to "WayLinker" re-creates these rd5's including the "estimated_traffic_class" tags. Only the profile "fastbike-lowtraffic" is using these tags.

If I understand correctly, it means the current script computes traffic in 5x5 latitude / longitude squares around the 6 specified points, right? So the `fastbike-lowtraffic` with this script should be usable as is only around these locations?

Thanks!

Issue #116, opened 2018-10-11
Thoughts on profile development and testing


wetneb:
I find it quite frustrating that #58 is not merged, but I understand the underlying issue: as a maintainer, how can you be confident that changes to your profiles will not badly screw up the results in some cases?

So, do we need a sort of test-driven profile development? Has this been done anywhere else? Here a few random thoughts of how this could work.

A test case could be an origin, destination and one or more test point(s) through which the computed route would need to pass to validate the test. To ease maintenance, test cases could come with a short description of why the route is expected to pass there rather than elsewhere. Each profile would have its own test suite, but ideally it should be possible for these test suites to share common tests.

Should the tests be run against a frozen state of OSM? That would have the advantage of making the testing process more reliable and would lower the maintenance costs for the test suite. Or should they be run continuously as the map evolves? That would have the advantage of matching the real state of the map better, and potentially the maintainer of any important tagging changes.

If we had a nice test suite for a few profiles, we could even try to derive profile weights from test cases themselves! Convert all the test cases to a big bunch of linear inequations, and let the solver come up with something.

I'm sure plenty of people have done similar things before (I am not familiar with the field at all) so it would be important to research what is already out there before.

poutnikl:
IMHO, the best test environment is area the profile developer/tester is familiar with, so it cannot be the coming one.

It is on such a person familiar with area roads/ways to consider, what route the profile with given priorities should take and to check what it really chooses.

Also, concerning surface quality data, there is need to compare OSM data with reality, in case they do not match (well).

wetneb:
Yes I agree that it helps a lot to know an area to design test cases. But I do not understand what you mean by "it cannot be the coming one"? Which "coming one"?

poutnikl:
It was an Android typo, I have immediately corrected it as common, but you replied to the original

wetneb:
Okay, I still don't understand with "common" instead of "coming", so I have also edited my post!

poutnikl:
In my understanding, you asked for some common test cases and test scenarios.

wetneb:
I think a lot of basic tests can be understood without knowing the place… I think they should only be added in places where the decision can be made based on the OSM data (otherwise there is no point trying to figure out a routing profile that somehow manages to pass the test for an unrelated reason).

More specifically I am thinking of the following workflow:
* I am using some profile to plan my trips;
* One day I find out that one route was bad - I update the map to indicate why (steps, bad road surface, narrow corner, whatever);
[...]

PR #123, opened 2018-11-10
Improve the trekking profile

Phyks:
**EDIT**: I'll split this large pull request into smaller ones to make it easier to debug and merge into BRouter. Just leaving this one open here for anyone interested in the full PR at the moment.

Hi,

Here is an attempt to improve the trekking profile while modifying it as less as possible. The idea was to really keep the way it is working at the moment but slightly improve some routes using all the available tags. This profile requires #122 first to work fully.

I'll comment in the diff to better explain what the differences are and why I introduced these.

I tried to find minimal tests cases for the features I wanted to improve, a bit like what was thought of in #116. My tests are written in a iPython notebook available [in this repo](https://github.com/Phyks/BrouterTesting) (please feel free to reuse / merge / write PR on this one if you think that's something worth it).

A static copy of my tests cases with detailed descriptions of the expected behaviors is available at https://pub.phyks.me/tmp/BrouterTesting.html (blue is my profile, grey is the original trekking profile, green is what I am likely to do as a human). The tests are, of course, biased towards the areas I am familiar with, but cover a bit of cities / cycleroutes / countries.

Best,

EDIT: The logic described in this profile is probably a bit lengthy and could likely be written in a shorter way, but I think it was easier to read this way, at least for a first review.

### DIFF
--- a/misc/profiles2/trekking.brf
+++ b/misc/profiles2/trekking.brf
@@ -87,19 +87,26 @@ assign onewaypenalty =  # Penalty associated with going against traffic in a one
        # Note: Can be zero if opposite traffic is allowed for bikes
        if ( badoneway ) then
        (
-         if ( cycleway=opposite|opposite_lane|opposite_track ) then 0
-         # No penalty if there is any cycleway in opposite direction
-         # Note: We are already contraflow, so no need for more checks.
+         # Not a oneway street for bikes
+         if ( oneway:bicycle=no ) then 0
+         # No penalty if there is any cycleway going against traffic
+         else if ( cycleway=opposite|opposite_lane|opposite_track ) then 0
+         # Otherwise, check the value of cycleway:oneway
+         else if ( and reversedirection=yes cycleway:oneway=-1|no ) then 0
+         else if ( and ( not reversedirection=yes ) cycleway:oneway=yes|no ) then 0
+         # No penalty if there is any left cycleway in opposite direction
          else if ( cycleway:left=opposite|opposite_lane|opposite_track ) then 0
+         else if ( and reversedirection=yes cycleway:left:oneway=-1|no ) then 0
+         else if ( and ( not reversedirection=yes ) cycleway:left:oneway=yes|no ) then 0
+         # No penalty if there is any right cycleway in opposite direction
          else if ( cycleway:right=opposite|opposite_lane|opposite_track ) then 0
-         else if ( cycleway:left:oneway=-1|no ) then 0
-         else if ( cycleway:right:oneway=-1|no ) then 0
+         else if ( and reversedirection=yes cycleway:right:oneway=-1|no ) then 0
+         else if ( and ( not reversedirection=yes ) cycleway:right:oneway=yes|no ) then 0
          # No penalty if shared_busway and busway in opposite direction
          else if ( and cycleway=share_busway ( or busway=opposite ( or busway:left=opposite busway:right=opposite ) ) ) then 0
          else if ( and cycleway:left=share_busway ( or busway:left=opposite busway=opposite ) ) then 0
          else if ( and cycleway:right=share_busway ( or busway:right=opposite busway=opposite ) ) then 0
          # Other cases
-         else if ( oneway:bicycle=no                         ) then 0
          else if ( highway=primary|primary_link              ) then 50
          else if ( highway=secondary|secondary_link          ) then 30
          else if ( highway=tertiary|tertiary_link            ) then 20
--- a/misc/profiles2/trekking.brf
+++ b/misc/profiles2/trekking.brf
@@ -12,7 +12,9 @@ assign validForBikes = true
 assign allow_steps              = true   # %allow_steps% | Set to false to disallow steps | boolean
 assign allow_ferries            = true   # %allow_ferries% | set to false to disallow ferries | boolean
 assign ignore_cycleroutes       = false  # %ignore_cycleroutes% | Set to true for better elevation results | boolean
+assign ignore_access_tags_on_cycleroutes = false  # %ignore_access_tags_on_cycleroutes% | Set to true to ignore access restrictions on cycle routes | boolean
 assign stick_to_cycleroutes     = false  # %stick_to_cycleroutes% | Set to true to just follow cycleroutes | boolean
+assign is_right_side_driving    = true   # %is_right_side_driving% | Set to false if using in a left-side driving country | boolean
 assign avoid_unsafe             = false  # %avoid_unsafe% | Set to true to avoid standard highways | boolean
 assign considerTurnRestrictions = false  # %considerTurnRestrictions% | Set to true to take turn restrictions into account | boolean
 assign processUnusedTags        = false  # %processUnusedTags% | Set to true to output unused tags in data tab | boolean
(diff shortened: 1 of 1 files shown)


misc/profiles2/trekking.brf
```diff
@@ -52,7 +54,146 @@ assign is_ldcr =
...
+#
+assign badoneway =  # True if we are going against traffic in a one-way street
+       if reversedirection=yes then
+         if oneway:bicycle=yes then true
+         else if oneway= then junction=roundabout
+         else oneway=yes|true|1
+       else oneway=-1
+
+assign goodoneway = if reversedirection=yes then oneway=-1  # True if we are going along traffic in a one-way street
+                      else if oneway= then junction=roundabout else oneway=yes|true|1
```
Phyks: These lines were pre-existent and simply moved here for convenience purpose as I am using them afterwards.

misc/profiles2/trekking.brf
```diff
@@ -52,7 +54,146 @@ assign is_ldcr =
...
+       (
+         if ( cycleway=opposite|opposite_lane|opposite_track ) then 0
+         # No penalty if there is any cycleway in opposite direction
+         # Note: We are already contraflow, so no need for more checks.
+         else if ( cycleway:left=opposite|opposite_lane|opposite_track ) then 0
+         else if ( cycleway:right=opposite|opposite_lane|opposite_track ) then 0
+         # No penalty if shared_busway and busway in opposite direction
+         else if ( and cycleway=share_busway ( or busway=opposite ( or busway:left=opposite busway:right=opposite ) ) ) then 0
+         else if ( and cycleway:left=share_busway ( or busway:left=opposite busway=opposite ) ) then 0
+         else if ( and cycleway:right=share_busway ( or busway:right=opposite busway=opposite ) ) then 0
```
Phyks: I refactored the oneway penalty to correctly take into account the possible contraflow cycleways, in particular the shared busways.

misc/profiles2/trekking.brf
```diff
@@ -52,7 +54,146 @@ assign is_ldcr =
...
+        )
+        else (
+            or
+            ( and ( reversedirection=yes ) ( not cycleway:right:oneway=-1 ) )
+            ( and ( not reversedirection=yes ) ( cycleway:right:oneway=-1|no ) )
+        )
+    )
+
+# TODO: isbike could be a float value dependending on the quality of the infrastructure
+assign isbike =
```
Phyks: The `isbike` variable is now split into different checks for left/right cycleways, done above.

misc/profiles2/trekking.brf
```diff
@@ -66,7 +207,7 @@ assign probablyGood = or ispaved and isbike not isunpaved
 #
 assign turncost = if is_ldcr then 0
                   else if junction=roundabout then 0
-                  else 90
+                  else 80
```
Phyks: I slightly diminished this turncost so that cycleways around roundabout (and parallel to roads without cycleways) would not get too penalized. There is such an exemple in test suite.

misc/profiles2/trekking.brf
```diff
@@ -52,7 +54,146 @@ assign is_ldcr =
...
+
+# TODO: isbike could be a float value dependending on the quality of the infrastructure
+assign isbike =
+    if bicycle=yes|permissive|designated then true
+    else if lcn=yes then true
+    # Take into account cycleways
+    else if or has_cycleway ( or has_left_cycleway has_right_cycleway ) then true
+    else false
+
+# Add a penalty based on traffic speed, if there are no infrastructures
```
Phyks: I added an extra penalty for roads without cycleways based on the `maxspeed`, considering any `maxspeed` above 30 is way faster than a bike and not really safe so should be penalized.

misc/profiles2/trekking.brf
```diff
@@ -90,22 +231,26 @@ assign initialcost =
 # implicit access here just from the motorroad tag
 # (implicit access rules from highway tag handled elsewhere)
 #
-assign defaultaccess =
-       if access= then not motorroad=yes
-       else if access=private|no then false
-       else true
+assign defaultaccess = not ( access=private|no )
```
Phyks: This (and below) is a fix for https://github.com/nrenner/brouter-web/issues/31. This new behavior can be disabled by a variable at the beginning on the profile.

misc/profiles2/trekking.brf
```diff
@@ -170,6 +291,12 @@ assign costfactor
   else if ( highway=motorway|motorway_link ) then   10000
   else if ( highway=proposed|abandoned     ) then   10000
 
+  #
+  # put a penalty on roads under constructions
+  # See https://github.com/abrensch/brouter/issues/106 and https://github.com/abrensch/brouter/issues/62
+  #
+  else if ( highway=construction     ) then   100
```
Phyks: Try to avoid (but not completely) roads under construction.

misc/profiles2/trekking.brf
```diff
@@ -307,6 +433,10 @@ assign footaccess =
        else if foot= then defaultaccess
        else not foot=private|no
 
+assign trafficsignalcost =
```
Phyks: I added an extra cost for each traffic signal. It does not change much the routes found by BRouter (as traffic signals repartition is rather homogeneous and isotropic I guess), but helps in some particular cases where the cycleway is poorly designed and forces to wait at several traffic lights in a row.

There is such an example in the test suite where the cycleway forces to go downhill, pass 3 traffic signals and go back uphill while you could actually just use the pedestrian crossing (with a foot penalty) and only wait for one traffic signal.

misc/profiles2/trekking.brf
Phyks: Hmmm, it seems that `construction=yes` is deprecated (see https://wiki.openstreetmap.org/wiki/Key:construction#Usage).

`construction=minor` is mutually exclusive with `highway=construction` apparently but could be a good addition indeed (with a small penalty). In my experience (France), roads which would qualify for `construction=minor` are likely to have no usable cycleway (even if the road has one) so I guess the cost factor could be the one of the `highway` without `isbike` for these (if this local experience can be extended worldwide).

This would require to add the `construction` tag in `lookups.dat` file though.

abrensch:
Hi Lucas,

thanx for this PR, and I finally found some time for review.

So we can divide it into the following changes:

1) presence of cycleway tagging qualifies for isbike/probablyGood
2) no more overruling of access tags by cycle-routes
3) 20+30 speed-limits reduces costfactor for probablyGood=false
4) traffic-light penalty
5) reduced turncost
6) some fuzzy changes on the access logic

I think I like 1), am passionless for 2)-5), but I don't like 6)

I like 1) because there has always been complaints that brouter/trekking completely ignores cycleway tagging and thus has a bias for separately mapped cycleways. I sometimes argued that most of these cycleways are not what bikers are looking for, bad surfaces, shared with footways etc, But I see that this is probably differnt in big cities like paris, where you better look for cycleways.

I don't like 6) it' a loss of clarity and systematics. The "bikeaccess" completely ignores a "bicycle=no". It's only evaluated if it's redundant because there's also an access=no. The access logic is build on the assumption that if there is explicit access tagging, then this is evaluated according to the access hirarchy, and only if access tagging is not present, then some implicit rules apply. Only exception so far was motorways, where a bicycle=yes does not enables access (Which does not work in Australia, which is why there is a "fastbike_asia_pacific" profile...). But now there's a new exception for motorroad=yes. And, if I read it correctly, a "access=no bicycle=yes" can lead to access-denied, if there's a cycleway=share_busway but no bus=yes.  For me that looks like it needs a second thought.

I really think that trekking needs un update and it needs more international input. However, it should be evolution, not revolution, so maybe we can start with something less impacting?

regards, Arndt

Phyks:
Thanks for the feedback on this one! I'm ok with your comments and feedback (see below). I would add a

> 7) Handle a bit better the opposite oneways (such as shared busway in opposite direction etc).

(but you might already include this one in another item?). 

> I like 1) because there has always been complaints that brouter/trekking completely ignores cycleway tagging and thus has a bias for separately mapped cycleways. I sometimes argued that most of these cycleways are not what bikers are looking for, bad surfaces, shared with footways etc, But I see that this is probably differnt in big cities like paris, where you better look for cycleways.

My personal opinion with respect to this (but you are probably more experienced with bike routing than I am) is that BRouter should take cycleways whenever available, as cycleways are, by default, safer and adapted for bikes. However, it's true that many cycleways are not really good, with bad surfaces or issues with pedestrians. These problems could be tagged by dedicated OSM tags (like [segregated](https://wiki.openstreetmap.org/wiki/Key:segregated) or [surface](https://wiki.openstreetmap.org/wiki/Key:surface)) and could then be avoided based on these parameters. So, I'd be more in favor of using them by default and adding higher penalties based on surface etc if problems arise.


> am passionless for 2)-5)

Most of these are just minor corrections to get slightly better results in very specific cases :)


> I don't like 6)

You are right, I probably messed up a bit with the logic here. It's totally fine for me to discard this for the moment if the rest can be merged, and I might rework it a bit.


> I really think that trekking needs un update and it needs more international input. However, it should be evolution, not revolution, so maybe we can start with something less impacting?

I'm not sure to understand what you mean by something less impacting (is it 6 only or rather this much changes)?

Feel free to merge only a fragment of this PR (or ask me to open separate PRs) if only parts of this one could be fine for a merge. If you think 1-5 (everything except 6) is too much changes at a time (but still worth it), maybe an option could be to have a `trekking-rework` profile or something like this, to explicitly emphasize this is beta and includes a lot of merge?

Best,

misc/profiles2/trekking.brf
```diff
@@ -52,7 +54,147 @@ assign is_ldcr =
...
+assign onewaypenalty =  # Penalty associated with going against traffic in a one-way stret
+       # Note: Can be zero if opposite traffic is allowed for bikes
+       if ( badoneway ) then
+       (
+         if ( cycleway=opposite|opposite_lane|opposite_track ) then 0
+         # No penalty if there is any cycleway in opposite direction
+         # Note: We are already contraflow, so no need for more checks.
+         else if ( cycleway:left=opposite|opposite_lane|opposite_track ) then 0
+         else if ( cycleway:right=opposite|opposite_lane|opposite_track ) then 0
+         else if ( cycleway:left:oneway=-1|no ) then 0
```
Mincol: Should not this be dependent on reversedirection=yes?
```suggestion
          else if ( and reversedirection=yes cycleway:left:oneway=-1|no ) then 0
          else if ( and not reversedirection=yes cycleway:left:oneway=yes ) then 0
```

Phyks: This is enclosed within a `if ( badoneway) `: https://github.com/abrensch/brouter/pull/123/files#diff-2bbe87cbb0ab11e03987f2ebbd823c31R75

which should test for `reversedirection` https://github.com/abrensch/brouter/pull/123/files#diff-2bbe87cbb0ab11e03987f2ebbd823c31R63.

So, if I'm not mistaking, this part should only consider `cycleway:*=-1|no` which means against traffic (`reversedirection`).

If we are going same way as traffic, there should not be any particular `onewaypenalty` since we are not going against traffic in any way.

Mincol: Example:
badoneway = true because reversedirection=<null> and oneway=-1.
In other words:
- We are going in direction as way was mapped
- direction of way does not agree with direction of one way route. This is discouraged, but can happen.

The one way for cyclist agrees with direction of way in map so cycleway:left:oneway=yes

Is it clear now? Or is it me who get it wrong?

Phyks: Oh, indeed, thanks for the details. I guess this should be fixed in https://github.com/abrensch/brouter/pull/123/commits/5a1f7dccedc0c97f21166f0b661d9f562b8f7829. Could you check please?

Parts of this PR has already been included in https://github.com/abrensch/brouter/pull/136. I'll make smaller (atomic) commits to help merge the rest of this large one afterwards :)

Mincol: I have additional comments, but reading @abrensch comment, might be good to start new pull request and close this one. (Subject to style and opinion)

It might be worth to go through all conditions again and check them. Like line [80](https://github.com/abrensch/brouter/blob/5a1f7dccedc0c97f21166f0b661d9f562b8f7829/misc/profiles2/trekking.brf#L80) was not fixed. Maybe worth refactoring a bit.

Also I do not understand line [133](https://github.com/abrensch/brouter/blob/5a1f7dccedc0c97f21166f0b661d9f562b8f7829/misc/profiles2/trekking.brf#L133).

Phyks: First, thanks for your feedback and review on this! Editing BRouter profiles can be tricky and having mulitple pairs of eyes looking at the code is definitely great!

> I have additional comments, but reading @abrensch comment, might be good to start new pull request and close this one. (Subject to style and opinion)

This is what I had in mind when I started #136 with atomic edits, one edit per commit to ease review and discussion. :) Plus there are some duplicates and introducing intermediate variables would help a lot here I think.

If you have any comments on what is already included in #136, please feel free to review on this one, so this one can get fixed and eventually merged :) Then, I think we can move on to the following edits.

> It might be worth to go through all conditions again and check them.

Sure! Additionnally, if you know any real life examples for these conditions (OSM data), please feel free to report them to me or add them to https://github.com/Phyks/BrouterTesting. I used this to ensure my edits were not completely absurd and working on some features, but I don't have tests for all the conditions.

> Like line 80 was not fixed.

Actually, I was thinking this one was right. The issue here is about the meaning of `cycleway=opposite*` in OSM I think. My understanding was that in a oneway street, `cycleway:*=opposite*` means that the cycleway goes against traffic in a one-way street, which is always what we should be looking for when within a `badoneway`.

> Also I do not understand line 133.

Yes, I guess this should be written the same way as line 80 anyways. This part is quite difficult to read and definitely needs refactoring.

misc/profiles2/trekking.brf
Phyks: https://groups.google.com/forum/#!msg/osm-android-bikerouting/plmpWPPcqqQ/QqfpEKr5AwAJ is a discussion of interest about traffic signals penalty.

Totorrr:
@Phyks Here is my take on "considering the presence of a usable cycleway tagged on the highway": 

after the `assign onewaypenalty` block, insert: 

```
#
# re-assign isbike considering the presence of a usable cycleway on the highway
#
assign isbike = 
       or isbike
       or cycleway=lane|track|shared_lane|share_busway
       or and badoneway cycleway=opposite|opposite_lane|opposite_track
       or 
         and cycleway:left=lane|track|shared_lane|share_busway
           switch badoneway
             cycleway:left:oneway=no|-1
             true
         and cycleway:right=lane|track|shared_lane|share_busway
           switch badoneway
             cycleway:right:oneway=no|-1
             true
```

I tried to keep it concise. 

Re-assigning `isbike` may not be the best integration, but for my use case (evolution without revolution to stay close to this upstream repository) it limits too deep changes to the rest of the file by just being an independant new block. 

This can/might be inspiration for considering tagged cycleways.

Totorrr:
@Phyks Here is my take on "considering the presence of a usable cycleway tagged on the highway": 

after the `assign onewaypenalty` block, insert: 

```
#
# re-assign isbike considering the presence of a usable cycleway on the highway
#
assign isbike = 
       or isbike
       or cycleway=lane|track|shared_lane|share_busway
       or and badoneway cycleway=opposite|opposite_lane|opposite_track
       or 
         and cycleway:left=lane|track|shared_lane|share_busway
           switch badoneway
             cycleway:left:oneway=no|-1
             true
         and cycleway:right=lane|track|shared_lane|share_busway
           switch badoneway
             cycleway:right:oneway=no|-1
             true
```

I tried to keep it concise. 

Re-assigning `isbike` may not be the best integration, but for my use case (evolution without revolution to stay close to this upstream repository) it limits too deep changes to the rest of the file by just being an independant new block. 

This can/might be inspiration for considering tagged cycleways.

afischerdev:
Please feel free to reopen if needed.

PR #125, opened 2018-11-18
Travel time computation for bikes and foot

Phyks:
Hi,

Here is a first implementation of the basic model (Tobler's hiking function for foot, constant power for bike) from #6.

This is still a WIP:
* It does not work yet at the moment, not exactly sure why.
* The code could probably be improved and constants moved to the profiles.

Best,


brouter-core/src/main/java/btools/router/StdPath.java
```diff
@@ -139,7 +143,45 @@ else if ( ehbu < 0 )
...
+      double C_r = 0.01;
+      // Biker power
+      double BIKER_POWER = 100;  // in W
+
+      // Uphill angle
+      double alpha = Math.atan2(delta_h, distance);
+
+      // Travel speed
+      double speed = Double.NaN;
+      if (rc.footMode) { // TODO: || tags['way'].search('bicycle=dismount') !== -1) {
```
Phyks: Is there an easy way to access the `bicycle=dismount` tag at this point? Thanks!

Phyks:
Should be working fine now and brouter-web is already managing its output and showing it. The constants might need some minor adjustments (they can be overloaded in the profiles). On a typical route I'm doing daily, this simple model gives me 25 minutes. Google Maps gives 35 minutes but is clearly too large and Geovelo (French based) gives me 28 minutes.

The reality is between 25 and 30 minutes, depending a lot on the traffic signals.

There is the handling of `bicycle=dismount` tag which I don't know how to handle from within the Java code left to do. But otherwise, should be usable.

brouter-core/src/main/java/btools/router/StdPath.java
```diff
@@ -139,7 +145,31 @@ else if ( ehbu < 0 )
...
+        // Use Tobler's hiking function for walking sections
+        speed = 6 * Math.exp(-3.5 * Math.abs(delta_h / distance + 0.05)) / 3.6;
+      } else {
+        // Compute the speed assuming a basic kinematic model with constant
+        // power.
+        Cubic speedEquation = new Cubic();
+        speedEquation.solve(rc.S_C_x, 0.0, (rc.bikeMass * GRAVITY * (rc.defaultC_r + Math.sin(alpha))), -1.0 * rc.bikerPower);
```
Phyks: I'm still not very well versed into the BRouter routing, so I'm not sure whether this is done for *every* segment which is even considered or only for the segments which will be part of the final track.

If that's the latter case, I guess this is rather performance-sensitive and solving the Cubic equation there (although the code could be optimized a bit) might not be the best idea :/

abrensch: Hi Lucas,

Yes, StdPath.processWaySection is called for every segment evaluated by thr router. However, in the calling class OsmPath there's a flag "detailMode", and if you make that available in processWaySection  you can use that to do this calculation only for the final pass, which touches far less way sections.

Guess you could replace the cubic formula by newton approximation. Just need a good starting value.

One more remark: the whole idea I think is to catch the fact that going up and down is slower on average than going straight. But to achieve that you should use the filtered elevation-delta, because other wise the elevation noise made by buildings will have the same effect than real hills.

regards, Arndt

PR #126, opened 2018-11-23
Fix travel time computation

Phyks:
Hi Arndt,

Thanks for the merge! I saw you pushed an extra commit (https://github.com/abrensch/brouter/commit/6e6c4a18b21db90e9106bafe0be03fcb9b51d4fc), which is totally fine for me. There is just one slight issue, which is fixed in this PR, in that the `computeKinematic` method was not called :/

Best,

### DIFF
--- a/brouter-core/src/main/java/btools/router/OsmPath.java
+++ b/brouter-core/src/main/java/btools/router/OsmPath.java
@@ -369,7 +369,7 @@ abstract class OsmPath implements OsmLinkHolder
         }
       }
 
-      
+
 
 
       double elevation = ele2 == Short.MIN_VALUE ? 100. : ele2/4.;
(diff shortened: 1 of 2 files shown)


brouter-core/src/main/java/btools/router/OsmPath.java
```diff
@@ -396,68 +396,66 @@ protected void addAddionalPenalty(OsmTrack refTrack, boolean detailMode, OsmPath
...
+      {
+        message.turnangle = (float)angle;
+        message.time = (float)getTotalTime();
+        message.energy = (float)getTotalEnergy();
+        message.priorityclassifier = priorityclassifier;
+        message.classifiermask = classifiermask;
+        message.lon = lon2;
+        message.lat = lat2;
+        message.ele = ele2;
+        message.wayKeyValues = rc.expctxWay.getKeyValueDescription( isReverse, description );
```
Phyks: I reverted this change from my PR, as this is no longer needed.

brouter-core/src/main/java/btools/router/OsmPath.java
```diff
@@ -396,68 +396,66 @@ protected void addAddionalPenalty(OsmTrack refTrack, boolean detailMode, OsmPath
...
-          message.time = (float)getTotalTime();
-          message.energy = (float)getTotalEnergy();
-          message.priorityclassifier = priorityclassifier;
-          message.classifiermask = classifiermask;
-          message.lon = lon2;
-          message.lat = lat2;
-          message.ele = ele2;
-          message.wayKeyValues = wayKeyValues;
-        }
+      if ( message != null )
```
Phyks: Github highlights a lot of diff here, but this is just fixing the indent which was one extra indent.

brouter-core/src/main/java/btools/router/OsmPath.java
```diff
@@ -396,68 +396,66 @@ protected void addAddionalPenalty(OsmTrack refTrack, boolean detailMode, OsmPath
         traffic += dist*rc.expctxWay.getTrafficSourceDensity()*Math.pow(cost2/10000.f,rc.trafficSourceExponent);
       }
 
-      String wayKeyValues = "";
-      if ( message != null ) {
-        wayKeyValues = rc.expctxWay.getKeyValueDescription( isReverse, description );
-      }
+      // compute kinematic
+      computeKinematic( rc, dist, delta_h, detailMode );
```
Phyks: Added a call to `computeKinematic`.

abrensch:
Hi Lucus,

thanx for your work,

was late yesterday, so I didn't test my commit at all... I did test the convergence of the solveCubic method and found out that  convergence is 100% reliable if the start-value fo speed is above the expected result, so no need to handle the non-convergence case. I added another elevation-filter, just to smooth the elevation a little bit - otherwise when leaving a bridge o tunnel it could happen that you calculate a "free fall" section. But still the question is if it's better to start with th eresult of the "old" elevation filter that completely eliminates the small height  variations that you encounter when passing near buildings.

The "dismount" hack I removed for now. Maybe it's better to use just "costfactor >= 5" as the criterium to switch to walking mode, because this includes pedestrain areas and wrong oneways as well?
 
regards, Arndt

Phyks:
No problem, thanks for the merge!

> The "dismount" hack I removed for now. Maybe it's better to use just "costfactor >= 5" as the criterium to switch to walking mode, because this includes pedestrain areas and wrong oneways as well?

Indeed, this seems like a really nice and convenient way of handling all these cases without worrying too much about logic and performance. Definitely :+1: for this.

Issue #3, opened 2014-01-19
cleanup distance calculation


abrensch:
3 issues with distance calculation in BRouter:
- duplicate code ( 3 implemantions of public in calcDistance )
- significant distance mismatch compared to other tools evaluating the tracks (earth radius?)
- fails at lon = 180 degree 

However, as distance calucation is called frequently within brouter, any cleanup also must obey the performance constraints

Phyks:
Hi,

I ran a quick test on this one. I'm not exactly sure where the `calcDistance` formula is coming from in BRouter (there are some magical constants, especially the `0.0…01234134` and the `57.3`). 

I tested with a simple segment of about 880 meters and I get:
* 881.6690494652235 from BRouter `calcDistance`. If I don't take any `cos` approximation, I get 889.7434180976005.
* 880.5645655193547 using the [Haversine formula](https://en.wikipedia.org/wiki/Haversine_formula) (no approximations made in the computation of cosine etc).

Concerning the performance constraint, a possible idea would be to precompute a table of values for cosine and sine functions and use it, rather than using a Taylor expansion which might not be optimal.

I think the Haversine formula is the best one for distance computation, but I might be wrong / missing something.

nrenner:
Here the Leaflet implementation:

[LatLng.distanceTo](https://leafletjs.com/reference-1.3.4.html#latlng-distanceto) ([CRS.Earth.distance](
https://github.com/Leaflet/Leaflet/blob/736bd586975987e7429ed0f737dc1c9b51927b03/src/geo/crs/CRS.Earth.js#L17-L32)):

```js
// Mean Earth Radius, as recommended for use by
// the International Union of Geodesy and Geophysics,
// see http://rosettacode.org/wiki/Haversine_formula
R: 6371000,

// distance between two geographical points using spherical law of cosines approximation
distance: function (latlng1, latlng2) {
    var rad = Math.PI / 180,
        lat1 = latlng1.lat * rad,
        lat2 = latlng2.lat * rad,
        sinDLat = Math.sin((latlng2.lat - latlng1.lat) * rad / 2),
        sinDLon = Math.sin((latlng2.lng - latlng1.lng) * rad / 2),
        a = sinDLat * sinDLat + Math.cos(lat1) * Math.cos(lat2) * sinDLon * sinDLon,
        c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a));
    return this.R * c;
}
```

Phyks:
[...]

PR #128, opened 2018-12-02
Better distance computation

Phyks:
Hi,

Here is a first step toward improving on #3. I'll add some comments in this PR for possible next steps and questions.

Thanks,

**EDIT** Computation of cosine of latitude now relies on a precomputed table of cosine values, which should be much more efficient when repeatedly calling it. I am computing distance through a code (under FOSS license) from Mapbox, see https://github.com/abrensch/brouter/issues/3 for more background on this.


brouter-core/src/main/java/btools/router/RoutingContext.java
```diff
@@ -226,6 +227,7 @@ public static void prepareNogoPoints( List<OsmNodeNamed> nogos )
         try { ir = Integer.parseInt( s.substring( 4 ) ); }
         catch( Exception e ) { /* ignore */ }
       }
+      // TODO[Phyks]
       nogo.radius = ir / 110984.; //  6378000. / 57.3;
```
Phyks: Is it expected to be Earth radius over (180 / PI) ? The value of the constant does not match with the comment :/

brouter-core/src/main/java/btools/router/RoutingContext.java
```diff
@@ -236,13 +238,14 @@ public void cleanNogolist( List<OsmNodeNamed> waypoints )
     List<OsmNodeNamed> nogos = new ArrayList<OsmNodeNamed>();
     for( OsmNodeNamed nogo : nogopoints )
     {
+      // TODO[Phyks]
       int radiusInMeter = (int)(nogo.radius * 111894.);
```
Phyks: Is it expected to be Earth radius over (180 / PI) ? The value of the constant does not match with the usual values at other places in the code.

brouter-core/src/main/java/btools/router/RoutingContext.java
```diff
@@ -261,14 +264,15 @@ public void cleanNogolist( List<OsmNodeNamed> waypoints )
...
     {
-    	OsmNodeNamed nogo = nogopoints.get(i);
-    	cs[0] += nogo.ilon;
-    	cs[1] += nogo.ilat;
-    	cs[2] += (long) ( nogo.radius*111894.*10.);
+      OsmNodeNamed nogo = nogopoints.get(i);
+      cs[0] += nogo.ilon;
+      cs[1] += nogo.ilat;
+      // TODO[Phyks]
+      cs[2] += (long) ( nogo.radius*111894.*10.);
```
Phyks: Same question for the value of the constant. What is the `10` here for as well?

brouter-core/src/main/java/btools/router/RoutingContext.java
```diff
@@ -286,14 +290,13 @@ public void unsetWaypoint()
...
   {
-    double l = (lat2 - 90000000) * 0.00000001234134;
-    double l2 = l*l;
-    double l4 = l2*l2;
-    coslat = 1.- l2 + l4 / 6.;
-    double coslat6 = coslat*0.000001;
+    CheapRulerSingleton cr = CheapRulerSingleton.getInstance();
+
+    coslat = cr.cosIlat(lat2);
+    double coslat6 = coslat*cr.ILATLNG_TO_LATLNG;
```
Phyks: I kept the faster approximate distance computation below, and only compute the true distance at the end of the loop.

However, this code is very similar to the one found in  `WaypointMatcherImpl` and I think this could be factored into a common function call. What do you think?

brouter-core/src/main/java/btools/router/WaypointMatcherImpl.java
```diff
@@ -32,6 +33,7 @@ public WaypointMatcherImpl( List<MatchedWaypoint> waypoints, double maxDistance,
     this.islandPairs = islandPairs;
     for ( MatchedWaypoint mwp : waypoints )
     {
+      // TODO[Phyks]
       mwp.radius = maxDistance * 110984.; //  6378000. / 57.3;
```
Phyks: Same question about the constant value wrt Earth radius etc.

brouter-core/src/main/java/btools/router/RoutingContext.java
abrensch: Originally yes, but then I tuned to better distance results of other gpx tools

brouter-core/src/main/java/btools/router/RoutingContext.java
abrensch: seems buggy

brouter-core/src/main/java/btools/router/RoutingContext.java
abrensch: factor 10 is arbitrary, just to have sub-meter precision in the checksum

brouter-core/src/main/java/btools/router/WaypointMatcherImpl.java
abrensch: same answer...

brouter-core/src/main/java/btools/router/RoutingContext.java
abrensch: Isn't what CheapRuler.distance(..) uses internally as "kx", "ky" more or less the same then "coslat6", "0.00001" ? Except for that facrtor 111894.

So for me it seems you should'nt provide a coslat-lookup, but a kx,ky lookup, because that's basically what's needed all over the place. So by replacing coslat6 + 0.00001 -> kx+ky we could consistently redefine all these internal radius/distance measurements that are now in "cosinius normalized degrees" to be measures in meters, and everything woul be much more readable.

It would still be ugly code. RoutingContext.calcDistance does:
- distance calculation
- nogo matching
- start/endpoint matching
- cosinus calculation for later use

so it's result  is not just the return-value, but also rc.coslat, rc.shortestmatch, rc.wayfraction, nogo.radius - this is really spaghetti, but refactoring that maybe you should do some warmup first :-)

And youare right, WaypointMatcherImpl is doing similar things, because it's solving the same problem: finding the closest match of a point to a line connecting to other points.

brouter-core/src/main/java/btools/router/RoutingContext.java
Phyks: Ok, I'll write a fix for it then :)

brouter-core/src/main/java/btools/router/RoutingContext.java
Phyks: Ok, thanks for the infos!

brouter-core/src/main/java/btools/router/RoutingContext.java
Phyks: > Isn't what CheapRuler.distance(..) uses internally as "kx", "ky" more or less the same then "coslat6", "0.00001" ? Except for that facrtor 111894.

Indeed, that's exactly it. I was trying to avoid recomputing `kx` and `ky` since it involves computing powers of `cos` terms. But it could totally be cached and having a lookup for this would probably make sense. I'll go for it :)

> this is really spaghetti, but refactoring that maybe you should do some warmup first :-)

I'll try to write something to address this, try to refactor a bit the code and fix the remaining points. Feel free to merge this PR if you think it is ok. I can either push additional stuff in this one or open a new one, as it suits you best.

PR #130, opened 2018-12-07
Follow up on distance rework

Phyks:
Hi,

Thanks for the merge of the previous PR and the extra commit. Here is the last one on this subject (I think).

First commit is about storing the radius in meters everywhere in the code to have better precision (circle is then defined as point within a given radius in meter rather than having to deal with difference between latitude and longitude scales).

Second commit is just some doc commit to fix some errors I had when building JavaDoc.

Best,

### DIFF
--- a/brouter-core/src/main/java/btools/router/OsmNogoPolygon.java
+++ b/brouter-core/src/main/java/btools/router/OsmNogoPolygon.java
@@ -109,7 +109,6 @@ public class OsmNogoPolygon extends OsmNodeNamed
     double dlat2m = lonlat2m[1];
 
     double rad = 0;  // radius
-    double rad2 = 0; // radius squared;
 
     double dmax = 0; // length of vector from center to point
     int i_max = -1;
@@ -246,10 +245,9 @@ public class OsmNogoPolygon extends OsmNodeNamed
  /**
   * winding number test for a point in a polygon
   *
-  * @param p a point
-  * @param v list of vertex points forming a polygon. This polygon
-  *          is implicitly closed connecting the last and first point.
-  * @return the winding number (=0 only when P is outside)
+  * @param px longitude of the point to check
+  * @param py latitude of the point to check
+  * @return a boolean whether the point is within the polygon or not.
   */
   public boolean isWithin(final long px, final long py)
   {
--- a/brouter-core/src/main/java/btools/router/OsmNogoPolygon.java
+++ b/brouter-core/src/main/java/btools/router/OsmNogoPolygon.java
@@ -103,7 +103,7 @@ public class OsmNogoPolygon extends OsmNodeNamed
 
     int cx = (cxmax+cxmin) / 2; // center of circle
     int cy = (cymax+cymin) / 2;
-    
+
     double[] lonlat2m = CheapRulerSingleton.getLonLatToMeterScales( cy );
     double dlon2m = lonlat2m[0];
     double dlat2m = lonlat2m[1];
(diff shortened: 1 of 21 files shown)


brouter-core/src/main/java/btools/router/OsmNogoPolygon.java
```diff
@@ -103,54 +103,45 @@ else if (p.y > cymax)
...
     double dlon2m = lonlat2m[0];
     double dlat2m = lonlat2m[1];
 
     double rad = 0;  // radius
-    double rad2 = 0; // radius squared;
 
-    double dpx = 0; // x-xomponent of vector from center to point
-    double dpy = 0; // y-component
-    double dmax2 = 0; // squared lenght of vector from center to point
+    double dmax = 0; // length of vector from center to point
```
Phyks: I think this is a bit easier and more readable if using distance directly.

brouter-core/src/main/java/btools/router/OsmPath.java
```diff
@@ -346,7 +346,8 @@ protected void addAddionalPenalty(OsmTrack refTrack, boolean detailMode, OsmPath
         if ( rc.startDirectionValid )
         {
           double dir = rc.startDirection.intValue() / CheapRulerSingleton.DEG_TO_RAD;
-          lon0 = lon1 - (int) ( 1000. * Math.sin( dir ) / rc.getCosLat() );
+          double coslat = Math.cos(lat1);
```
Phyks: There was a bug in the last commit as `rc.coslat` is never updated anymore. I've just discarded it as it is not really needed except here. But this should probably be `kx` actually, I'll fix it.

brouter-core/src/main/java/btools/router/RoutingContext.java
```diff
@@ -404,28 +402,23 @@ else if ( s1 > s2 )
     return (int)(d + 1.0 );
   }
 
-  // assumes that calcDistance/calcCosAngle called in sequence, so coslat valid
   public double getCosAngle()
   {
     return cosangle;
   }
 
-  public double getCosLat()
```
Phyks: Coslat is no longer needed nor updated.

brouter-core/src/main/java/btools/router/RoutingContext.java
```diff
@@ -404,28 +402,23 @@ else if ( s1 > s2 )
...
-    return coslat;
-  }
-
   public double calcAngle( int lon0, int lat0,  int lon1, int lat1, int lon2, int lat2 )
   {
-    double dlat1 = (lat1 - lat0);
-    double dlon1 = (lon1 - lon0) * coslat;
-    double dlat2 = (lat2 - lat1);
-    double dlon2 = (lon2 - lon1) * coslat;
+    double[] lonlat2m = CheapRulerSingleton.getLonLatToMeterScales( lat1 );
```
Phyks: We can use kx and ky here to compute the direction without relying on coslat, I think.

brouter-routing-app/src/main/java/btools/routingapp/BRouterView.java
```diff
@@ -635,7 +635,9 @@ private void paintCircle( Canvas canvas, OsmNodeNamed n, int color, int minradiu
     int lat = n.ilat - centerLat;
     int x = imgw / 2 + (int) ( scaleLon * lon );
     int y = imgh / 2 - (int) ( scaleLat * lat );
-    int ir = (int) ( n.radius * 1000000. * scaleLat );
+
+    double[] lonlat2m = CheapRulerSingleton.getLonLatToMeterScales( centerLat );
+    int ir = (int) ( n.radius * scaleLat / lonlat2m[1]);
```
Phyks: I'm really not sure about the scaling here and I don't know much the Android app. I think this is fine, but definitely needs a second look.

brouter-util/src/main/java/btools/util/CheapRulerSingleton.java
```diff
@@ -33,7 +33,7 @@
   }
 
   private static double[] calcKxKyFromILat(int ilat) {
-    double lat = DEG_TO_RAD*(ilat-90000000)/1000000.;
+    double lat = DEG_TO_RAD*ilat*ILATLNG_TO_LATLNG - 90;
```
Phyks: Just a little rewrite to make it a bit more readable and remove magic constants, in my opinion.

brouter-core/src/main/java/btools/router/OsmPath.java
```diff
@@ -346,8 +346,9 @@ protected void addAddionalPenalty(OsmTrack refTrack, boolean detailMode, OsmPath
         if ( rc.startDirectionValid )
         {
           double dir = rc.startDirection.intValue() / CheapRulerSingleton.DEG_TO_RAD;
-          lon0 = lon1 - (int) ( 1000. * Math.sin( dir ) / rc.getCosLat() );
-          lat0 = lat1 - (int) ( 1000. * Math.cos( dir ) );
+          double[] lonlat2m = CheapRulerSingleton.getLonLatToMeterScales( lat1 );
+          lon0 = lon1 - (int) ( 1000. * Math.sin( dir ) / lonlat2m[0] );
+          lat0 = lat1 - (int) ( 1000. * Math.cos( dir ) / lonlat2m[1] );
```
Phyks: `rc.coslat` is never updated anymore, so I simply discarded it. The only remaining use case was here, where I think we can use `lonlat2m` this way, but should be checked.

PR #138, opened 2019-01-05
Add support for nogo polygons / polylines weight in brouter server

Phyks:
Hi,

After #117, nogo polygons / polylines can have a custom weight. The BRouter-server part of the code was not updated to be able to consume polygons / polylines with a custom weight. This is done in this PR.

The polylines/polygons parameter is now read two by two (for latitude and longitude pairs). If a trailing single value is available, it is considered to be the weight.

This PR also includes a fix for a bug with GeoJSON generation. When no voice hints were emitted, the comma at the end of the `cost:` line was deleted, resulting in an invalid GeoJSON file.

Best,

### DIFF
--- a/brouter-server/src/main/java/btools/server/request/ServerHandler.java
+++ b/brouter-server/src/main/java/btools/server/request/ServerHandler.java
@@ -271,13 +271,17 @@ public class ServerHandler extends RequestHandler {
         if ( lonLatList.length > 1 )
         {
           OsmNogoPolygon polygon = new OsmNogoPolygon(closed);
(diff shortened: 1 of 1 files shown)


Phyks:
Hi @abrensch,

Could you have a look at this PR? I dropped the first commit which was fixing an issue in output when no voice hint was generated, as it seems this has been fixed in a different commit (https://github.com/abrensch/brouter/blob/master/brouter-core/src/main/java/btools/router/OsmTrack.java#L594).

Thanks!
Best,

Phyks:
Thanks

PR #158, opened 2019-05-11
Content-Disposition filename must be quoted

bagage:
In case filename contains spaces or anything non ascii, we have to quote the content disposition header to have it work properly.

Similarly, avoid strange characters in filename.


bagage:
@abrensch as far as I could test, I think this is the last piece missing before enabling that feature in brouter and brouter-web :tada:

PR #162, opened 2019-06-01
Make it easier to directly run BRouter from source checkout

rkflx:
- Refactor/merge server.sh and local.sh
- Allow to run server.sh and server.cmd from source directory
- Update README

Test Plan:
- clone repo
- run maven
- drop rd5 into ``misc/segments4``
- launch ``./misc/scripts/standalone/server.sh``

(Note how no further configuration is necessary. In addition, compatibility with the directory structure as found in BRouter's zip archives is kept.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+misc/segments4
+misc/customprofiles
 target/
 *.BAK
 .classpath
--- a/README.md
+++ b/README.md
@@ -10,9 +10,22 @@ For more infos see http://brouter.de/brouter
 <a href="https://play.google.com/store/apps/details?id=btools.routingapp" target="_blank">
 <img src="https://play.google.com/intl/en_us/badges/images/generic/en-play-badge.png" alt="Get it on Google Play" height="90"/></a>
 
+
+## Build and Install
+
 Compile with (Java 6!):
 
 > mvn clean install -Dandroid.sdk.path=<your-sdk-path>
+
+To skip building for Android, add ``-pl '!brouter-routing-app'``.
+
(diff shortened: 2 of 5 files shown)


Phyks:
Not sure why I cannot comment directly on the diff, but on the new `local.sh`,

I think this should be `./server.sh`, isn't it? (Could also simply be a gitted symlink, but I'm not sure how this would work on Windows).

Best,

rkflx:
> (Could also simply be a gitted symlink, but I'm not sure how this would work on Windows).

To quote the commit message (c27fa57): "By using a symlink and comparing the running script name we can avoid duplicating most of the code."

Regarding how this might work on Windows, see https://github.com/git-for-windows/git/wiki/Symbolic-Links. However, by default "Enable symbolic links" is unchecked in the installer, so indeed the link becomes a textfile, which obviously does not work. Thanks for noticing.

> I think this should be `./server.sh`, isn't it? 

This won't allow to check for `basename`, and will break calling the script from arbitrary directories again.

I'll replace the link with a script setting `BINDADDRESS` instead.

nrenner:
I was actually thinking about moving those scripts to brouter-web, see #163.

Phyks:
> This won't allow to check for basename, and will break calling the script from arbitrary directories again.

Indeed, but the current "server.sh" would only work if the file is in `$PATH` :) The new fix with `dirname` looks fine to me, thanks.

Phyks:
@abrensch Would it be possible to merge this one? It would be super useful to get started hacking on BRouter I think.

Additionnally, note that the current `server.sh` file has Windows line endings, which behaves badly on Unix.

Issue #174, opened 2019-07-24
Support living_street tag


Phyks:
Hi,

It seems BRouter supports the `highway=living_street` tag (of course), but I could not find anything related to the `living_street=yes` tag. This tag is quite old, and rather undocumented (https://wiki.openstreetmap.org/wiki/Key:living_street), it is even considered a trolltag.

However, it is used 250k times around the world (https://taginfo.openstreetmap.org/keys/living_street#overview) which seems more than enough to be considered as a de facto tag and taken into account in BRouter.

So, what about adding `living_street=yes` support in `lookups.dat` and update the profiles so that `living_street=yes` is considered as a `highway=living_street` (or with a slight penalty)?

I can handle the changes if there is a consensus on this.

Best,

abrensch:
sure, go ahead. But we should scan teh open issues for all other pending lookup.dat requests, there's also issue #166 by you, and in the google group there's something about stop signs:

Issue #166, opened 2019-06-28
Sidewalk:bicycle tag


Phyks:
Hi,

I've just discovered the `sidewalk:bicycle=yes` tag which seems not very much used (450 times worldwide), mostly in Germany: https://overpass-turbo.eu/s/Kke.

I found a description for its use case only in the French wiki (case N0): https://wiki.openstreetmap.org/wiki/FR:Bicycle#Am.C3.A9nagements_cyclables_sur_trottoirs. This seems to differ from the usual S3/S4 case from the English wiki page (https://wiki.openstreetmap.org/wiki/Bicycle) by the fact that this is not segregated (so this is, to my understanding, an alias for the S3/S4 cases with `segregated=no`).

So far, it is not taken into account by the `lookups.dat` file and therefore by BRouter for routing. I don't really have a use case for having this tag, but it might be useful in Germany. What do you think?

Best,

PR #176, opened 2019-08-03
Add new lookups

Phyks:
Hi,

This should address #166 and #174 (one commit for each):

1. I updated the profiles so that `living_street=yes` is equivalent to `highway=living_street`.
One might argue that `living_street=yes` should have a slight penalty with respect to `highway=living_street` to reflect that this might not be fully a living street (`highway=secondary` + `living_street=yes` probably means there is much traffic and one can expect this to be a poor quality living street). I, however, don't have enough feedback so far on the use of `living_street` across the world (seems to be mostly used in Russia so far), so I think considering it as an alias is fine for a first version.

2. I added support for `sidewalk:bicycle` tag (and its `left`/`right` variants). This one is mainly used in Germany to describe an implicit sidewalk which can be cycled on (https://wiki.openstreetmap.org/wiki/DE:Key:sidewalk). I did not edit the `trekking` profile accordingly, it might be worth extending `isbike` to account for `sidewalk:bicycle=yes`, but I did not find any documentation (or missed it, I don't speak german :/) about whether this should be considered oneway or not by default.

3. @abrensch you mentionned
> in the google group there's something about stop signs

Do you have more infos on this? I can't find it back :/

Thanks!
Best

### DIFF
--- a/misc/profiles2/car-eco-suspect_scan.brf
+++ b/misc/profiles2/car-eco-suspect_scan.brf
@@ -32,6 +32,8 @@ assign initialcost switch route=ferry 20000 0
 #
 # calculate logical car access
 #
+assign isresidentialorliving = or highway=residential|living_street living_street=yes
+
 assign caraccess
        switch motorcar=
               switch motor_vehicle=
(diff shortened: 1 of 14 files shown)


Phyks:
I added a last commit to add support for `bicycle_road=yes` (https://taginfo.openstreetmap.org/keys/bicycle_road#overview), as requested by some profile developpers.

Phyks:
I added a commit to handle #183.

PR #179, opened 2019-08-11
Add MTB profile

bagage:
This is @Poutnik profile for MTB - which is not yet available in default profiles in brouter but I think it makes sense to have one?

I am using it for several years now and am pretty happy with it. Basically it is a small variant of trekking profile with prefer track/path over roads.

### DIFF
--- a/misc/profiles2/mtb.brf
+++ b/misc/profiles2/mtb.brf
@@ -1,0 +1,562 @@
+  # Poutnik's Trekking profile template for Trekking profile variants
+  # Version 2.6.* new max cost system
+  #
+  # *)  = See also https://github.com/poutnikl/Brouter-profiles/wiki 
+  # **) = See also https://github.com/poutnikl/Trekking-Poutnik/wiki
+  #
+  # MTB profile, based on MTBiker feedback
+  #
+  # Legend above is the placeholder for generated comments of final profiles
+  # See the HistoryLog.txt file for file history, the wikis above for verbose comments.
+
+  # bstart /global
+  ---context:global
+  assign   iswet                  0 # 0 as default, *) flag for weather conditions
+  assign   turnInstructionMode  = 1 # 0=none, 1=auto-choose, 2=locus-style, 3=osmand-style
+  assign   cycleroutes_pref       0.2 # also CRP *)  costfactor penalty for not being cycleroute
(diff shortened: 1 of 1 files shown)


poutnikl:
I think @zossebart  profile for MTB [MTB-zossebart.brf](https://github.com/zossebart/brouter-mtb) may make more sense, as it was created primarily for MTB. 

My MTB profile is rather Trekking bike profile "bended" toward MTB. Additionally, I do not ride MTB to test it to realize if I would like it.

zossebart:
I think both profiles might be useful.
I did only get feedback from other mtbikers very sparsely (they were all positive so far). As I developed the profile for areas I know (mid mountain ranges), I don't know if it is useful for more alpine areas for example.
As you have positive feedback from @bagage and I guess some others, your profile seems also valid for mtb use. I must admit that I did not test your mtb profile so far...

bagage:
> My MTB profile is rather Trekking bike profile "bended" toward MTB. Additionally, I do not ride MTB to test it to realize if I would like it.
For my own MTB use, I find it useful as is.

I didn't test @zossebart one, I'll add it to https://brouter.damsy.net to compare with it.

bagage:
I *do* think that indeed both are valuable - even if in my area (plain and small hills) I tend to prefer @poutnikl profile which favores tracks/paths a bit more.

@abrensch do you think both could be added? I think we are quite a few to use BRouter for MTB too, https://github.com/nrenner/brouter-web/issues/295 amongst others :).

abrensch:
I was always hesitant adding MTB profiles, because single-trail rides are not always legal in Germany. However, meanwhile there's an alternative instance that does contain MTB profiles: https://www.marcusjaschen.de/blog/2020/brouter-web/
I will think about it, but I'm also thinking about the general hosting subject, cause the current growth suggests to put that on more or stronger feet.

poutnikl:
I think majority of MTB users do not use them for their original downhill riding purpose.

They use them rather as "trekking plus" bikes, avoiding paved roads even more and using even more rough ways, that may be too much for regular travellers on trekking bikes.

Or, they are often used by hard core travellers.

afischerdev:
@abrensch 

> I'm also thinking about the general hosting subject

Do you still see a hosting problem? The lookups.dat already contains all tag mtb.brf is using.

Issue #152, opened 2019-05-06
GPX Export -- tag for "generation url"


bosdla:
Hi,
one useful feature in order to adapt an old track to a new would be to store the "generation url" into the gpx file header: like: 

```
<trk>
  <name>brouterxxxx</name>
  <url>http://brouter.de/brouter-web/#map=10/53.5506/10.0007/standard&lonlats=9.967346,53.450334|9.742126,53.275735</url>
```
Then it also would be possible to load a gpx track file by using this url.

Would it be hard to implement? What do you think?
Thanks,
Michael

PR #188, opened 2019-09-03
Add generation URL in GPX export

Phyks:
Hi,

Here is a first draft for adding generation URL in GPX export, as requested in #152. I did not add the full URL as this required passing too much information down to the core, in my opinion, and rather opted for a URL query string of parameters.

This is straightforward compatible with the hash part of the URL in BRouter-web.

I can extend this to KML and GeoJSON exports as well if you think this is worth it.

Happy to have any feedback on this! Thanks

### DIFF
--- a/brouter-core/src/main/java/btools/router/OsmTrack.java
+++ b/brouter-core/src/main/java/btools/router/OsmTrack.java
@@ -374,20 +374,20 @@ public final class OsmTrack
    * @param filename
    *          the filename to write to
    */
-  public void writeGpx( String filename ) throws Exception
+  public void writeGpx( String filename, RoutingContext rc ) throws Exception
   {
     BufferedWriter bw = new BufferedWriter( new FileWriter( filename ) );
-    formatAsGpx( bw );
+    formatAsGpx( bw, rc );
     bw.close();
   }
 
-  public String formatAsGpx()
+  public String formatAsGpx( RoutingContext rc )
   {
     try
     {
       StringWriter sw = new StringWriter( 8192 );
       BufferedWriter bw = new BufferedWriter( sw );
-      formatAsGpx( bw );
+      formatAsGpx( bw, rc );
       bw.close();
       return sw.toString();
     }
@@ -397,7 +397,7 @@ public final class OsmTrack
     }
   }
 
-  public String formatAsGpx( BufferedWriter sb ) throws IOException
+  public String formatAsGpx( BufferedWriter sb, RoutingContext rc ) throws IOException
   {
     int turnInstructionMode = voiceHints != null ? voiceHints.turnInstructionMode : 0;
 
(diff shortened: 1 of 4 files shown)


nrenner:
I'm a bit unsure if this should be in OsmTrack in the core. The URL format is specific for BRouter-Web. I guess the question is, whether a link for BRouter-Web should be generated for all callers or only when actually called from BRouter-Web. For the latter it could be generated in the ServerHandler and passed as a string property in the RoutingContext, or by a callback hook?

Ideally, the link would be a full URL, but without the map parameter. Hash parsing currently can't handle a missing map parameter, I created an issue for that: nrenner/brouter-web#265. Host and port could be read from the `Origin` or `Referer` header.

Don't know if it's worth also adding this to KML and GeoJSON.

As the code is a bit lengthy, I personally would prefer to factor out the URL string generation itself into a separate private method.

0709wiwiwi:
Very useful when applied to the exported gpx file.

I use this service.
https://t.ly/1EzE

Apparently a completely free service for individual users.  ?  
You still need to insert this link in the exported gpx file.  
Preferably you (personally) could already insert this in the BRouter website. 

An example gpx file is attached.  
[Gräbendorf -_ Dolgenbrodt_plus_link (7.9km).gpx.txt](https://github.com/abrensch/brouter/files/8946667/Grabendorf.-_.Dolgenbrodt_plus_link.7.9km.gpx.txt)

zod:
As already stated by @nrenner the URL format is specific to BRouter-Web. BRouter-Web no provides client-side track formatting which would allow BRouter-Web to add the URL.

I'm not sure how using a URL shortener relates to this issue. What's the benefit of using it? The URL should be part of the file anyway, so it doesn't matter if it's shortened. IMHO this would just add a dependency to a service without providing any benefit. If the services goes away any time in the future the embedded link in older files would be worthless.

0709wiwiwi:
@ zod.

- The BRouter link in the gpx is a luxury, but so very easy for later quick reuse.
- Short url preference example ? Yes because compactnes in the gpx file is just nicer.
- I am a gps user, not a dev, so this https://yourls.org/ is beyond my capabilities.
- Later you can add the "primal link" to the gpx export, so why not directly.
- A (short) url in the gpx file like this https://t.ly/cxec is imo so more valuable than next toy.
Notice a complete manual traject desgin by the url !  To compare with next QR download.

Manual design QR

nrenner:
I think this PR can be closed here, now that we do formatting in the client, the URL can be added in brouter-web.

As for using short link services, I'm generally against those for privacy and security reasons and won't click on those:
- privacy: the provider can/will make a profile of the sites you visit
- security: you don't see what the target URL is and if it's trustworthy

For formatting links in issues with a text instead of showing the URL, use the [Markdown](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) notation `[text](url)`.

> [Gräbendorf -_ Dolgenbrodt_plus_link (7.9km).gpx.txt](https://github.com/abrensch/brouter/files/8946667/Grabendorf.-_.Dolgenbrodt_plus_link.7.9km.gpx.txt)

Instead of attaching complete GPX files, where the whole `<trkseg>...` part is completely irrelevant for the issue at hand, and where everyone has to download, open and guess what part of it is your point, please include only the relevant part in the comment directly, formatted as code (`<>` button):

```xml
<trk>
   <link href="https://t.ly/7094"></link>
```

0709wiwiwi:
Norbert, you may probably wish this.

<link href="https://t.ly/7094">
<text>BRouter Gräbendorf-Dolgenbrodt (7.9km)</text>
</link>

Then this presents In Locus with a good nice link text
Preferably not by short url (service) , but the long original web link does not validate :-((
Has an error message...but I do not know why,  I am not a dev.

nrenner:
> Norbert, you may probably wish this.


Yes, but not as image. You can copy&paste the text, select it, and format it as code with `<>` button in the formatting toolbar above the comment editor. Or manually enclose it in triple backticks, see [Quoting code](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code):
````
```xml
<code>
```
````

> Ampersand trouble ?
> I, as a standard gps user, do not find the use of that original link user-friendly.

Yes, the XML encoding is a problem when copying manually, but the main purpose is that other programs can show the link. I guess the "standard user" does not open files in an editor.

**Comment by nrenner** [comment]
Now implemented in brouter-web in [dcec6d4](https://github.com/nrenner/brouter-web/commit/dcec6d4b7776211fc718dda097d8de2de7eb3f1f) (not released yet).

**Comment by 0709wiwiwi** [comment]
As a shortlink, it could be done without any problem.
As a long url, I would not dare to do because is critical.
Some programs do not (re)export as `&amp;`

**Comment by nrenner** [comment]
> Some programs do not (re)export as `&amp;`

Then that's a bug and should be reported.

**Comment by 0709wiwiwi** [comment]
That is correct. It concerns the gpx editor program. This has two other smaller bugs known to me. These are less important and less annoying than the export of correctly imported web links. And I would definitely like to report this to the developer pixel_k.  A French developer probably from Alsace. But unfortunately I cannot reach pixel_k since the website gps passions suddenly disappeared. Before, there was always a very good friendly support. 
So if someone has contact with pixel_k, please let me know. Tnks.
https://sourceforge.net/projects/gpxeditor/

Translated with www.DeepL.com/Translator (free version)

**Comment by 0709wiwiwi** [comment]
> Now implemented in brouter-web in [dcec6d4](https://github.com/nrenner/brouter-web/commit/dcec6d4b7776211fc718dda097d8de2de7eb3f1f) (not released yet).

> As the code is a bit lengthy, I personally would prefer to factor out the URL string generation itself into a separate private method.

Could you possibly add an example of how it will look in a short demo gpx attachment here Norbert?

**Comment by nrenner** [comment]
>> As the code is a bit lengthy, I personally would prefer to factor out the URL string generation itself into a separate private method.

Why that quote? This is now obsolete and was referring to this PR's proposed code, not the URL itself.

> Could you possibly add an example of how it will look in a short demo gpx attachment here Norbert?

Generated GPX link example (extract, localhost replaced manually):
```xml
 <trk>
  <name>Konstanz (0,2 km)</name>
  <link href="https://brouter.de/brouter-web/#map=16/47.6676/9.1704/standard&amp;lonlats=9.174475,47.669043;9.173611,47.667508&amp;profile=fastbike">
   <text>BRouter-Web</text>
  </link>
  <trkseg>
```

Full example GPX file: [view](https://gist.github.com/nrenner/2b9fab42cb5b3ebfa2e867716ff4b3ca) &middot; [raw for saving](https://gist.githubusercontent.com/nrenner/2b9fab42cb5b3ebfa2e867716ff4b3ca/raw/b2165ef3e42fcaf0c0faa9fde9ac24e4bac2824b/link-example.gpx)

0709wiwiwi:
Quote: As I did not know what that meant exactly, I mentioned it and asked.
Thanks for the example. It confirms that the gpx editor export is indeed failing.
I know how to solve , but I'm probably not the only one who uses the gpx editor. 
Pretty annoying though, especially since I (no one) still have contact with the developer.

zod:
As mentioned by @nrenner this PR is obsolete because brouter-web as it's own output formatting now.

Issue #79, opened 2017-05-04
Applying way tag bicycle=use_sidepath


poutnikl:
Th way tagvalue bicycle=use_sidepath has very minor usage, near 0.04%  in Germany, according to lookup.dat.  But its wide world usage is cca 1.2% according to [OSM bicycle=* taginfo](https://taginfo.openstreetmap.org/keys/bicycle#values)

IMHO, standard bicycle profiles should give a penalty for roads with bicycle=use_sidepath to push the routing on the sidepath.

AllroadsNL:
This is also for mofa=use_sidepath and moped=use_sidepath and foot=use_sidepath

In the Netherlands, we now have a preset part for use_sidepath, more and more is tagged.

To control routing, moped routing must be set right.

![josm use_sidepath](https://user-images.githubusercontent.com/8817345/30511161-6eead114-9ad3-11e7-9643-fe41f01685f3.png)

And there is also the tag *:forward *:backward, which is very important, for all.
Here on one side is a oneway a cycleway used by moped:designated mofa:yes  foot must use cycleway in both directions,
On the other side a cycle:left=lane
Tags
bicycle:forward=use_sidepath
foot=use_sidepath
mofa:forward=use_sidepath
moped:forward=use_sidepath

![use_sidepath](https://user-images.githubusercontent.com/8817345/30511244-bbdee748-9ad4-11e7-8c86-8bf695ae3db2.png)

[routing situation](http://brouter.de/brouter-web/#zoom=19&lat=50.800827&lon=5.928272&layer=OpenStreetMap&lonlats=5.928363,50.800593|5.921271,50.807914&nogos=&profile=moped&alternativeidx=0&format=geojson)

Maybe also a other problem, in the Netherlands we have many cycleway traffic_sign NL:G12a with moped=designated compulsory. Does Brouter take designated with moped.



Many traffic sign access tags also use forward backward.

poutnikl:
As mentioned in #85, moped[:*]=* nor mofa[:*]=* tags are not yet condidered in BRouter data files, so for now, profiles cannot use them.

AllroadsNL:
Hope, they will be soon!

Phyks:
It seems that if `bicycle=use_sidepath` is used, then `isbike` is false (https://github.com/abrensch/brouter/blob/master/misc/profiles2/trekking.brf#L55) and then the way has a higher penalty. So it should actually already be pushing the route to the side path, isn't it?

[...]

PR #191, opened 2019-09-15
Handle bicycle=use_sidepath the same as bicycle=no

polyscias:
See https://github.com/abrensch/brouter/issues/79, this patch has bicycle=use_sidepath threatened the same as bicycle=no.

In NL ways that have a seperate cycle path along it do not have access=no but access=use_sidepath as some special types of cycles (with more than two wheels and wider than 75 cm) still may still use the main road. Because of this, this patch does not change vm-forum-liegerad-schnell.brf and vm-forum-velomobil-schnell.brf

### DIFF
--- a/misc/profiles2/fastbike-asia-pacific.brf
+++ b/misc/profiles2/fastbike-asia-pacific.brf
@@ -70,7 +70,7 @@ assign bikeaccess
                            switch or vehicle=private vehicle=no
                                   0
                                   1
-                 not or bicycle=private or bicycle=no bicycle=dismount
+                 not or bicycle=private|no|dismount|use_sidepath
 
 #
 # calculate logical foot access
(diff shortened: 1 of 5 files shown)


Phyks:
Hi,

You missed some `bicycle=no` handling I think. See https://github.com/abrensch/brouter/blob/master/misc/profiles2/trekking.brf#L109 for instance (and there are a couple of others in other profile files).

I don't see any reason not to handle `bicycle=use_sidepath` as `bicycle=no` (at least, they are the same from a legal point of view). This should however already be handled by the fact that a `bicycle=use_sidepath` already has a penalty with respect to the matching cycle way.

This however, should not solve #79 in my opinion (see my comment there).

abrensch:
Sorry I cannot merge this.

"use_side_path" *IS* weaker then "no". "no" means absolutly forbidden, and if it's only a 1m section that was "forgotton" at  a compilcated  crossing. Applying the same for use_sidepath would cut many connections.

I admit there are missing appropriate penalties for use_sideptah in most of the profiles, but it should be a penalty, not a total block.

PR #196, opened 2019-09-29
Handle user points of interest

bagage:


brouter-core/src/main/java/btools/router/OsmTrack.java
```diff
@@ -534,13 +537,22 @@ public String formatAsGpx( BufferedWriter sb ) throws IOException
       }
     }
 
+  for( int i=0; i<=pois.size() - 1; i++ ) {
```
Phyks: Missing indent?

brouter-core/src/main/java/btools/router/OsmTrack.java
```diff
@@ -534,13 +537,22 @@ public String formatAsGpx( BufferedWriter sb ) throws IOException
...
+
     if ( exportWaypoints )
     {
       for( int i=0; i<=matchedWaypoints.size() - 1; i++ )
       {
-        sb.append( " <wpt lon=\"" ).append( formatILon( matchedWaypoints.get(i).waypoint.ilon ) ).append( "\" lat=\"" )
-          .append( formatILat( matchedWaypoints.get(i).waypoint.ilat ) ).append( "\">\n" )
-          .append( "  <name>" ).append( matchedWaypoints.get(i).name ).append( "</name>\n" );
+          MatchedWaypoint wt = matchedWaypoints.get(i);
+        sb.append( " <wpt lon=\"" ).append( formatILon( wt.waypoint.ilon ) ).append( "\" lat=\"" )
```
Phyks: Missing indent?

brouter-core/src/main/java/btools/router/RoutingContext.java
```diff
@@ -21,7 +21,7 @@
 
 public final class RoutingContext
 {
-  public void setAlternativeIdx( int idx )
+    public void setAlternativeIdx(int idx )
```
Phyks: Extra indent?

abrensch:
I want to avoid linking apache commons, so I will, after merging, create a dummy class btools.utils.StringUtils with dummy methods for escapeJson and escapeXml10 and hope that someone puts the correct logic in.

nrenner:
> I want to avoid linking apache commons

I'm fine with this, just to clarify:
- is it about `brouter-core` (and Android) specifically or the whole project?
  e.g. maybe we could include it in the `brouter-server` project only and set/pass a wrapper to the core when run from there, the core could just have an interface and an empty dummy impl.
- is it about Apache Commons specifically or external dependencies in general?
  I would have suggested [OWASP Java Encoder Project](https://www.owasp.org/index.php/OWASP_Java_Encoder_Project) ([Encode.forXmlContent](https://owasp.github.io/owasp-java-encoder/encoder/apidocs/org/owasp/encoder/Encode.html#forXmlContent-java.lang.String-)) and [OWASP JSON Sanitizer](https://www.owasp.org/index.php/OWASP_JSON_Sanitizer), which seem more specialized and thus probably with smaller footprint, but haven't compared them yet.

abrensch:
>     * is it about Apache Commons specifically or external dependencies in general?

Dependencies in general. Once you open Pandora's box you end up anyhow with apache+google+log4j+slf4j+spring+trove, so it doesn't matter where you start.

But right, it's about the core package. I'm advertising BRouter as an embeddable library, and putting a dependency nightmare on top would kill these efforts.

However, the interface decoupling you sketched just describes the problem, not the solution. This PR here is about echoing client-parameters into a GPX or GeoJSON, and that shouldn't touch the routing core. I'm aware of that problem since long, and I already started to implement a new routing "core" that does not use the core package (and that does not know about GPX and GeoJSON). But I didn't finish that...

Don't know. Having dependencies in the server-package should be no problem, but maybe that could also be the trigger to separate the HTTP server as a separate project. I'm just not the right person to manage an external classpath. I'm doing that in my professional life and I don't want that as a hobby project.

PR #220, opened 2020-01-28
Add Github Action to publish Github Package

muthenberg:
This is a simple Github Action workflow, which tries to publish Maven artifacts to the Github Packages Repository on each commit to master. The workflow will only be successful if a new version has been released. We could improve this if each version would be tagged explicitly.


abrensch:
I cannot fully understand that, so I just merge and see what happens... I think I alsohave to enable github actions for this repo?

Tagging releases sure would be a good thing.

PR #224, opened 2020-02-02
Improve http server

wladich:
* return specific http codes for errors
* do not terminate old thread when threads limit reached
    Original implementation of the thread pool was limiting max number of
    threads by terminating oldest thread if the limit was already reached and
    a new request has arrived. This was causing broken responses under load.
    Instead we now wait until one of the threads completes before starting a
    new one.

### DIFF
--- a/brouter-server/src/main/java/btools/server/RouteServer.java
+++ b/brouter-server/src/main/java/btools/server/RouteServer.java
@@ -36,6 +36,11 @@ import btools.util.StackSampler;
 public class RouteServer extends Thread
 {
   public static final String PROFILE_UPLOAD_URL = "/brouter/profile";
+  static final String HTTP_STATUS_OK = "200 OK";
+  static final String HTTP_STATUS_BAD_REQUEST = "400 Bad Request";
+  static final String HTTP_STATUS_FORBIDDEN = "403 Forbidden";
+  static final String HTTP_STATUS_NOT_FOUND = "404 Not Found";
+  static final String HTTP_STATUS_INTERNAL_SERVER_ERROR = "500 Internal Server Error";
 
 	public ServiceContext serviceContext;
 
@@ -80,6 +85,8 @@ public class RouteServer extends Thread
               String line = br.readLine();
               if ( line == null )
               {
+                writeHttpHeader(bw, HTTP_STATUS_BAD_REQUEST);
+                bw.flush();
                 return;
               }
               if ( line.length() == 0 )
@@ -108,7 +115,7 @@ public class RouteServer extends Thread
               {
                 if ( agent.indexOf( tk.nextToken() ) >= 0 )
                 {
-                  writeHttpHeader( bw );
+                  writeHttpHeader( bw, HTTP_STATUS_FORBIDDEN );
                   bw.write( "Bad agent: " + agent );
                   bw.flush();
                   return;
@@ -118,11 +125,13 @@ public class RouteServer extends Thread
 
             if ( getline.startsWith("GET /favicon.ico") )
             {
-            	return;
+              writeHttpHeader( bw, HTTP_STATUS_NOT_FOUND );
+              bw.flush();
+              return;
             }
             if ( getline.startsWith("GET /robots.txt") )
             {
-              writeHttpHeader( bw );
+              writeHttpHeader( bw, HTTP_STATUS_OK );
               bw.write( "User-agent: *\n" );
               bw.write( "Disallow: /\n" );
               bw.flush();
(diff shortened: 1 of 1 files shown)


abrensch:
Hi Sergej,

thanks for that, and I'm fine with the status codes. But I don't think the thread-pooling will work that way.

The server-instance at brouter.de is running on a high timeout (9 minutes) and there are still ghost threads that do not come to an end before the timeout is reached. So your blocking semaphore would block "cheap" requests while waiting for the ghosts.

Maybe a "mixed approach" with 2 timeouts could be a solution, guaranteeing each thread a minimum running time (e.g.  1 second)?

I expect a further load increase in summer above 100k requests/day, so I take load issue seriously. Are you encoutering problems at the brouter.de instance, or are you using your own?

brouter.de is running on a 6 core (Xeon E5-2680 v3) system, so I think I could increase from 5 to 8 threads. Preferable on 2 or more Java-VMs, with nginx as a load balancer.

wladich:
Hi.
I want to explain motivation of this PR and ask some questions.

I am planning to host my own instance of brouter, I expect to have short distance requests (few kilometers max) at a rate up to 100-200 rps. So I am very interested in server stability.
I have made simple stress test (on a very slow server), like so: 
* started brouter with maxthreads=5
* `ab -c 10 -t 3 "localhost:17777/brouter?lonlats=-85.64315,40.13512|-85.64064,40.14186&profile=trekking&alternativeidx=0&format=geojson"`

In the output there is: 
```
Complete requests:      280
Failed requests:        60
   (Connect: 0, Receive: 0, Length: 60, Exceptions: 0)
```

With "-v 4" I can see that some responses are `operation killed by thread-priority-watchdog after 0 seconds`.

On brouter.de I see similar behaviour. 
With the proposed patch all request succeed.

(Following is my opinion, please do not take it as if I insist) In my opinion, a well behaving web service should stop accepting incoming requests if it does not have capacity to process them. Request that have not yet exceeded configured timeout should be considered valid and should not be terminated simply because we have new requests. Lack of capacity can be monitored by increase in response time or presence of 504 errors in nginx in front of brouter. In such case, maybe it is better to reduce maxRunningTime.

But you say there are also "ghost" threads which exceed maxRunningTime. 
Does that mean that they are getting stuck in the middle of `for(;;)` loop in _findTrack, not getting to `if ( maxRunningTime > 0 )`? I guess they are not being terminated by terminate() as well. Do you have any idea how it can happen? 

I could not reproduce this situation with quick an simple test: with original version I set maxRunningTime=60 and made 10 "heavy" requests with some delay between each of them - first five were aborted with "operation killed by thread-priority-watchdog", remaining ones stopped with "pass0 timeout after 60 seconds"  and no "ghosts" were seen in `pstree`. 
Maybe you know some requests which are known to create the ghosts or there are some other conditions? BTW, I guess using `pstree` is not the best way to monitor number of threads, how do you check them?

abrensch:
Sure, the simulation using "ab" with 10 threads and no thinktime gives the expected result. It's just designed like that.

Could by I was not clear enough for the "ghosts". They do not exceed maxRunningTime, but terminate themselve at maxRunningTime reliable. These are threads where e.g. the target-island detection failed, so they are trying to route to a location that is not reachable. Or a user accidentally issued a routing request Madrid->Moscow. Even if the client disconnects the http-connection, the server thread keeps running. These threads should not consume resources that are needed for real, short-running routing requests. 


So you expect, for 10 client-threads, 5 server-threads and 100ms original response time, to have a total response time of 200ms and no errors?

I think we can achieve that by having the new thread wait at max 1 second before killing anything.

If you are hunting for short request throughput there's another place where you might have a look: The class "ProfileCache" can keep only one "parsed profile", so it does not work for higher concurrency. Re-Using a parsed profile does not only save the actual parsing, but it also re-uses the Expression-Cache. So extending the Profile-Cache to store more profiles (at least the same number as maxThreads) could speed things up even if you always use the same profile.


regards, Arndt

wladich:
Seems I am late with my reply, but anyway here it is:

I think I got your point: you want to try serve long running requests and at the same time not do get DoS-ed when several such requests completely block new requests. 
Original version solved it by killing the oldest thread - not very elegant by was doing the job.
My variant with semaphore can easily stop receiving new requests if someone issues decent amount of "Madrid to Moscow" requests and maxRunningTime is big (you said it is 9 minutes now).

The proper solution is possibly to create a separate pool for long running threads:

- if a thread does not complete in few seconds, move it to "slow pool"
- if there are already more than N threads  running in slow pool, suspend the newly moved thread
- when one of slow threads terminates, resume next thread.

But that is out of scope of this PR.

BTW, I noticed that nginx at brouter.de returns 504 after 350 seconds, so maybe 9 minutes maxRunningTime you mentioned in your first comment is excessive.

abrensch:
So I finally came up with the mixed solution that guarantees 2 second to each request before it may be force-terminated.

I also extended the profile-cache to twice the size of the thread-pool.

And I changed the http-code for the "expected" errors (timeout, target-island, ...) from 500 to 400 ("bad request")

All my tests are o.k. now, including your "ab" test with 10 threads against a 5 thread server-instamce: it gave no errors and a short-request throughput of about 270 requests/second on my 6 core (Xeon E5-2680 v3) system.

Thanx for pointing out the timeout misconfiguration. Intention was to have the nginx timeout somewhat higher tgen the routing tmeout, so I changed it to 600/540 seconds now.

PR #242, opened 2020-05-17
export Oruxmaps Navigation instructions

vodie:


poutnikl:
Note that current hiking profiles from poutnikl/Brouter-profiles  based on the profile template poutnikl/Hiking-poutnik already contain priorityclassifier.  hiking-beta is a very obsolete version and should not be updated.

The best would be probably  replacing hiking-beta by hiking-SAC2, naming it just hiking.

PR #243, opened 2020-05-24
Switch build system to gradle

Erlkoenig90:
In order to be able to use modern Android tools including all the nice features of Android Studio I'd like to suggest switching the build system to gradle. This PR introduces the necessary changes. I kept the basic module structure and created gradle build files accordingly. A few files needed to be moved. The server and map-creator modules are compiled to Java applications as before, the routingapp is compiled into an Android app and the other modules are built as Java libraries. I removed the `pom.xml` files as I think gradle covers everything those did. The project can be imported in Android Studio (File -> New -> Import project) and compiled & run as usual. This PR is the base for another one ... :)


abrensch:
Hi,

thanx for that, I'm sure going with gradle is a good thing and I want to go for it.

However, I need more time to understand and get it running. After solving some Android SDK issues quickly (e.g. non accepted licences) I'm now stuck with a junit-dependency. It's probably my fault and I did not yet dig deep enough to ask clever questions.

I will come back here, but first I need success for my release builds (with both api-10 + api-28 versions) and for an automatic server build (that as far as I now is missing to get keep the F-Droid listing alive)

Erlkoenig90:
Hm, normally Android Studio (the IntelliJ one) / Gradle should automatically install and use the junit dependency as it is configured in the `build.gradle` files. To build from command line, run `ANDROID_SDK_ROOT=/path/to/Android/Sdk ./gradlew assembleRelease` from the project's root directory. To make signed apk's, see e.g. [here](https://stackoverflow.com/a/21020469).

afischerdev:
I'd like to merge this.
But I had to learn that after merging the project is not running out of the box via command line.
So I will start a second part of migration - will take a little time.

Do you have some blocking issues not to merge now?

abrensch:
>> Do you have some blocking issues not to merge now?

Go ahead, no avtivity currently from my side.

regards, Arndt

PR #244, opened 2020-05-24
Cleanup file / base directory access

Erlkoenig90:
I wanted to clean up the definition / access of the brouter base directory to be more aligned with Android's storage model - particularly, directly using `Context.getExternalFilesDirs` to find writable directories without manually constructing any paths. Effectively, the paths presented in the path selection dialog should always work. For the manual path entry, I added the capability to request the `WRITE_EXTERNAL_STORAGE` permission, so that more paths can be selected. Unfortunately, this still doesn't allow us to write to the actual SD card outside of the app-specific `Android` directory, but _does_ allow directly writing to `/storage/emulated/0` if the user wishes to. I also did some general clean-up, using `java.io.File` instead of just `String` in some places, and also try-with-resources. This breaks pre-KitKat compatibility, but I suspect apps like Locus won't run on devices this old anyways :)

With Android 11 "scoped storage", apps won't be able to access arbitrary shared locations at all any more. Therefore, it is probably necessary to refactor the various file-based data accesses in brouter and e.g. store everything in app-private directories and exchange all data via the Service interface. E.g. to avoid having to manually deal with profile files, brouter could offer a GUI to install/manage profiles, and client apps like Locus could query the installed profiles via the service API to present a selection dialog to choose one for routing. This would obviate the need for the user to manually deal with profile files. The file-based router interface probably can't be made to work on Android 11, so this probably needs to be completely converted to a Service-based one. A prettier, more user-friendly and localized GUI would probably also be nice. To avoid a lot of complexity, it would probably be a good idea to drop some backward-compatibility, perhaps doing this as a 2.0 version, which then would only work with accordingly-updated client apps.

Would there be any interest to refactor brouter like that?

This is based on the switch to gradle in #243.

### DIFF
--- a/brouter-core/src/main/java/btools/router/RoutingEngine.java
+++ b/brouter-core/src/main/java/btools/router/RoutingEngine.java
@@ -41,7 +41,7 @@ public class RoutingEngine extends Thread
 
   private volatile boolean terminated;
 
-  protected String segmentDir;
+  protected File segmentDir;
   private String outfileBase;
   private String logfileBase;
   private boolean infoLogEnabled;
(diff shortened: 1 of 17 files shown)


afischerdev:
@Erlkoenig90 
I'd like to merge this. 
Do you want to update your branch first or should I go the way to resolve the conficts?

Erlkoenig90:
Hi, I have no further updates, so you can do the merge :)

Issue #249, opened 2020-06-10
Process additional tags for analysis


nrenner:
We have a new feature in BRouter-Web that shows length distribution for the highway, surface, and smoothness tags ([brouter-web#304](https://github.com/nrenner/brouter-web/pull/304)).

With `processUnusedTags` set to `false` by default, this analysis doesn't get all the tags and values it would need though.

I see following options:

1. Add `processUnusedTags=true` to all profiles (this is what mjaschen does at his instance)
2. Pass a profile parameter from the client that overrides the default `processUnusedTags` setting
3. Pass a list of additional tags to include in the `RoutingContext`  
   (probably hard-coded in the `ServerHandler`)
4. Re-requesting the route when opening the analysis tab with `processUnusedTags=true` as profile parameter or in a modified custom profile
5. Re-requesting the route like in 4., but only when the user presses a button in the analysis tab

I don't know how much always using all tags will degrade performance and memory usage vs. how much issuing additional route requests for all tags on demand will increase overall load. Option 3 might be a compromise.

poutnikl:
I remember Arndt explicitly disabled  processing all tags by default in some BRouter version due performance reasons. I suppose enabling it by default should be limited for   debugging or profile development/testing only.

nrenner:
> But if it's just the smoothness tag missing, why not adding a dummy-usage?
> assign dummyUsage = smoothness=

@mjaschen What do you think about this suggestion? Would you want to make a PR adding such a dummy usage for missing tags to the profiles?

mjaschen:
> What do you think about this suggestion? Would you want to make a PR adding such a dummy usage for missing tags to the profiles?

That's totally fine with me!

For future-proofing the solution: Would it be possible to provide multiple tags/arguments to the `dummyUsage` parameter?

Currently there's only need for the `smoothness=` argument but who knows what surprises the future holds for us … ;-)

nrenner:
> For future-proofing the solution: Would it be possible to provide multiple tags/arguments to the dummyUsage parameter?

Yes, as I understand, `dummyUsage` is just an arbitrary variable name that isn't used anywhere. The idea is to add any valid syntax with no semantic meaning, just to make use of those extra tag(s) in the profile.

[...]

PR #250, opened 2020-06-13
add `dummyUsage` to certain profiles

mjaschen:
If BRouter calculates a route with a cycling or hiking profile,
the `smoothness` tag is included in the response. That enables
a statistical analysis for the calculated route (distribution of
smoothness tags by distance).

See the analysis sidebar at https://brouter.m11n.de/ for an example.

References: 

- #249 
- nrenner/brouter-web#45


vodie:
The vm-forum profiles already use the smoothness tag. They don't need a dummy usage.

mjaschen:
Thanks for hint, I‘ll update the PR.

PR #251, opened 2020-06-13
fix/unify the file extensions used in map creator

mjaschen:
Creating routing data files (`*.rd5`) isn't possible at the moment:

The "node tiles" are created in `OsmCutter` with file extensions
`.ntl` ([source](https://github.com/abrensch/brouter/blob/master/brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java#L341))

In later steps the node tiles are loaded. That fails in the current
version, because different file extensions are used:

- `NodeCutter` tries to load node tiles with `.tlf` extension ([source](https://github.com/abrensch/brouter/blob/master/brouter-map-creator/src/main/java/btools/mapcreator/NodeCutter.java#L37))
- `NodeFilter` tries to load node tiles with `.tls` extension ([source](https://github.com/abrensch/brouter/blob/master/brouter-map-creator/src/main/java/btools/mapcreator/NodeFilter.java#L51))
- `WayCutter` tries to load node tiles with `.tlf` extension ([source](https://github.com/abrensch/brouter/blob/master/brouter-map-creator/src/main/java/btools/mapcreator/WayCutter.java#L36))

This commit changes the file extensions used when loading the
node tiles to `.ntl` in the mentioned files.

### DIFF
--- a/brouter-map-creator/src/main/java/btools/mapcreator/NodeCutter.java
+++ b/brouter-map-creator/src/main/java/btools/mapcreator/NodeCutter.java
@@ -34,7 +34,7 @@ public class NodeCutter extends MapCreatorBase
   {
     init( nodeTilesOut );
 
-    new NodeIterator( this, true ).processDir( nodeTilesIn, ".tlf" );
+    new NodeIterator( this, true ).processDir( nodeTilesIn, ".ntl" );
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


abrensch:
To be true I don't understand the "old" pipeline anymore.

This is because it was replaced by the faster pipeline that uses less steps and less intermediate files (but more memory).

This faster pipeline is now used in the unit test ( https://github.com/abrensch/brouter/blob/master/brouter-map-creator/src/test/java/btools/mapcreator/MapcreatorTest.java ) and in the production script ( https://github.com/abrensch/brouter/blob/master/misc/scripts/mapcreation/process_pbf_planet_production.sh ) so the old pipeline should be considered deprecated.

The faster pipeline uses only 3 process calls:

- OsmFastCutter
- PosUnifier
- WayLinker

OsmFastCutter combines former steps ( OsmCutter, NodeFilter, WayCutter, NodeCutter, WayCutter5, NodeCutter5 ) by connecting them via internal pipes.

So I think I should remove the main() and process() methods of these steps alltogether.

But can acce

PR #257, opened 2020-06-29
Consider sett as paved in trekking.brf & fastbike.brf

Totorrr:
add "sett" as a paved surface (often nicer for bikes than paving_stones)
otherwise, sett surfaces tend to be avoided in routing, while it is usually smooth
even for thin wheels


afischerdev:
this have differences to the old compared scripts. But I controlled it and found ok.

PR #245, opened 2020-05-24
Manage cycleway left right opposite*

Florimondable:
Should fix #203 from #123.
I was not be able to manage cycleway=opposite_share_busway since it's not in the database.
I don't know how to add it.
https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dshare_busway

PR #268, opened 2020-09-25
Consider cycleway:left&right in onewaypenalty

Totorrr:
Avoids discarding some fine ways during routing

### DIFF
--- a/misc/profiles2/fastbike-verylowtraffic.brf
+++ b/misc/profiles2/fastbike-verylowtraffic.brf
@@ -124,7 +124,9 @@ assign badoneway =
 assign onewaypenalty =
        if ( badoneway ) then
        (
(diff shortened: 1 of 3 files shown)


Phyks:
Hi! This should be done as well in https://github.com/abrensch/brouter/pull/123 for the trekking profile, along with many other improvements. Happy to have any feedback on this one (positive or negative) and especially on whether it could supersede this one or fails on some cases. Thanks!

Totorrr:
Would you be willing to pull @abrensch @afischerdev @zod ?

afischerdev:
You are right, there are some outstanding profile changes.

I'm willing to add this - nice, has no conflicts on the first view.
But I see an other pending request #245 that like to deal with the same ideas. 
Please have a look there if this changes fit your needs.
As you will see the #245 has conflicts with other profiles, so I  would like to clear that first and merge it then.

Issue #293, opened 2021-03-20
Exclude highway=razed from routing


bus-mt:
BRouter routes over ways tagged with highway=razed. Example: https://brouter.de/brouter-web/#map=14/52.0361/9.4109/OpenTopoMap,HikeBike.HillShading&lonlats=9.391594,52.047476;9.419489,52.018909 includes https://www.openstreetmap.org/way/358348988/history which goes straight through a power plant.
BRouter should ignore ways tagged as highway=razed.

polyscias:
Have a look at https://wiki.openstreetmap.org/wiki/Tag:highway%3Drazed and you see why this tag is ignored.

Have a look at [taginfo](https://taginfo.openstreetmap.org/tags/?key=highway&value=razed) and you see the numbers are that small it is not surprise it is ignored.

mjaschen:
see PR #294

mjaschen:
It seems that this is a tagging issue. Thanks @polyscias for pointing this out.

bus-mt:
Makes sense. I looked it up in wiki, but only in the german translation where the additional clear hint is missing.

abrensch:
I added alias-values to the lookup.dat:

  proposed planned virtual
  abandoned disused razed demolished dismantled

that's the simplest solution, because it does not require to change profiles

PR #294, opened 2021-03-21
add `highway=razed` to lookup table

mjaschen:
#293 revealed that `highway=razed` is missing in `lookups.dat`.

This pull-request adds the missing entry. Taginfo counts [679 occurrences](https://taginfo.openstreetmap.org/search?q=highway%3Drazed) of `highway=razed` in OSM as of today.


polyscias:
Why add a tag that is marked as deprecated, see https://wiki.openstreetmap.org/wiki/Tag:highway%3Drazed

mjaschen:
You're absolutely right. I didn't noticed the deprecated flag.

Issue #314, opened 2021-06-30
App: change of the interface


afischerdev:
I'd like to add some params to the interface 'IBRouterService.aidl'
so that the interface becomes nearly equal to server.
Old params are still usable. 

```
. . .
//  "trackFormat"    = [kml|gpx|json] default = gpx

// "lonlats"         = lon,lat|... (unlimited list of lon,lat waypoints separated by |)
// "nogos"           = lon,lat,radius|... (optional, radius in meters)
// "polylines"       = lon,lat,weight|... (unlimited list of lon,lat and weight (optional) separated by |)
// "polygons"        = lon,lat,weight|... (unlimited list of lon,lat and weight (optional) separated by |)
// "profile"         = profile file name without .brf 
// "alternativeidx"  = [0|1|2|3] (optional, default 0)
// "exportWaypoints" = 1 to export them (optional, default is no export)
// "pois"            = lon,lat,name|... (optional)
// "extraParams"     = Bundle key=value list for a profile setup (like "profile:")
// "timode"          = turnInstructionMode [0=none, 1=auto-choose, 2=locus-style, 3=osmand-style, 4=comment-style, 5=gpsies-style, 6=orux-style] default 0

```

trackname - pathToFileResult  are equal
heading - direction  are equal
timode - turnInstructionFormat do same thing in a different way

nrenner:
> // "pois"            = lon,lat,name|... (optional)

This is one of the parameters that brouter-web only sends to get them back formatted, but have nothing to do with routing. Now that we have clint-side formatting implemented ([brouter-web#399](https://github.com/nrenner/brouter-web/pull/399), but not released yet), I was thinking of deprecating those.

EssBee59:
Hello nrenner,
here is possibly a confusion:
The IPC interface (aidl) is on Android between a navigation app (Osmand, locusmaps, oruxmaps) and the Brouter-app.
I think, the brouter-web is using an other interface (the http-interface to the brouter-java server?)
Regards
Ess Bee

afischerdev:
I think I missed some answers, sorry:

@nrenner 
I thought it could be an easy way to get out the way point elevation.
But you are right, it only formats the output.
At my point of view it could be dropped
OR it could be advanced to a new function to get elevation without sending a set of lonlats.
Are there any other user of this feature?

@EssBee59 
The idea was to keep the old params but use the new one first when set.

devemux86:
> heading - direction are equal

[...]

Issue #306, opened 2021-05-25
update all downloaded tiles button


lucamp:
Hi,
a button to update all downloaded tiles will be very appreciated

Thanks
Luca

zod:
Implemented in 1.6.3

Issue #301, opened 2021-05-04
foot maxspeed hardcoded


pkundrat:
Issue #125 implemented ETA computation for bike and hike with nice variables for bike, but for hiking, the max speed is 
hardcoded at 6 (brouter-core/src/main/java/btools/router/StdPath.java line 255)

[https://en.wikipedia.org/wiki/Tobler%27s_hiking_function](Wiki on Tobler hiking function)
The velocity on the flat terrain is 5 km / h, the maximum speed of 6 km / h is achieved roughly at -2.86.
On flat terrain this formula works out to 5 km/h. For off-path travel, this value should be multiplied by 3/5, for horseback by 5/4

So i suggest we reuse maxspeed variable from bike profiles, so that various hiking profiles could redefine this to get proper ETA.

Maybe the same can be achieved by uphillcost/downhillcost, but if you want to create a profile that matches your performance, the fiddling with those values to get the expected result is rather finicky.

afischerdev:
It affects the StdPath class which enables speed handling only for foot or bike. I'd like to enable others as well.
I'd like to fix this with a call of rc.maxSpeed which is init at readGlobalConfig() with 6 when footMode. Profile can overwrite this.
For all others (none foot or bike) speed is forewarded  with speed = wayMaxspeed.

afischerdev:
This brings me to other speed related things on the json format:

- there is a 'showspeed' parameter for json export to show speed instead of elevation..
  What about a new array 'speeds' like the 'times' array from #298 ?
  
- there is a 'speedprofile' array that is shown when 'vmax' is set in profile. But this removes the 'messages' array.
  What about a switch 'sendSpeedProfile' in the profile and then it comes additional to messages?

abrensch:
>> This brings me to other speed related things on the json format:
>>
>> there is a 'showspeed' parameter for json export to show speed instead of elevation..

that's a debug hack and it's intended to see the calculated speed in the elevation diagram

>> there is a 'speedprofile' array that is shown when 'vmax' is set in profile.
>> But this removes the 'messages' array. What about a switch 'sendSpeedProfile'
>> in the profile and then it comes additional to messages?

ja, that's another hack for a (yet) unpublished client for electric car routing.
The change you propose is o.k., however I wasn't sure what I can add to a geo-json
while keeping it valid geo-json

afischerdev:
Ok, I also use the showspeed for debug only. I leave it as before.

[...]

Issue #286, opened 2021-01-17
Black screen when swiping up or down


MakiWolf:
Black screen when swiping up or down
![issue](https://user-images.githubusercontent.com/73300633/104852079-f0cc8100-58f8-11eb-875b-643384db550d.gif)

PR #318, opened 2021-07-24
Update for Android 11 part 1

afischerdev:
This update has some parts for the move to Android 11

- change the inferface #314
- change main folder .../Android/media/btools.routingapp/
- has an own import folder for files like favorites or nogo* gpx
- export json #246
- update all tiles #306
- hardcoded speed #301
- black screen #286

still pending
- new icon brouter-web [286](https://github.com/nrenner/brouter-web/issues/286)
- refreshed profiles
- Android watchdog #277 - no error results during testing
- download manager #271 - no error results during testing
- download rd5 in background #227
- update for lookups.dat and profiles over the air

others
- git workflow: is there someone how can do the move to gradle? I've never done it and doesn't have a test environment.

This version is based on the idea other apps use the interface with a 'pathToFileResult' set to null - so the result comes back as string.
It is still possible to use the interface file driven - in Android 10+. But only if the result path is inside the .../Android/media/btools.routingapp/.. folder. This means to other apps using BRouter file export they have to deal with the Android storage access framework via a document url.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,4 @@ misc/customprofiles
 *.BAK
 .classpath
 .project
-*.iml
 !/.idea/runConfigurations
--- a/brouter-core/src/main/java/btools/router/RoutingContext.java
+++ b/brouter-core/src/main/java/btools/router/RoutingContext.java
@@ -123,6 +123,15 @@ public final class RoutingContext
   {
     BExpressionContext expctxGlobal = expctxWay; // just one of them...
 
+    if (keyValues != null) {
+      // add parameter to context
+      for (Map.Entry<String, String> e : keyValues.entrySet()) {
+        float f = Float.parseFloat(e.getValue());
+        expctxWay.setVariableValue(e.getKey(), f, false );
+        expctxNode.setVariableValue(e.getKey(), f, false );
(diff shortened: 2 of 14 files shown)


EssBee59:
> * git workflow: is there someone how can do the move to gradle? I've never done it and doesn't have a test environment.

Hello!
I am not really expert for maven / gradle, but I try to help ...
For my understanding:
Is your test version "brouter-test-and11" compiled with maven or gradle?

I dowloaded the zip file on my windows and could build the app using gradlew (manually) or android-studio.
So, I think, without maven?

The only real problem I had was  with the java (jar) version for using with the brouter-web:
To get it running, I had to change the build.gradle of brouter-server and the start commands.

afischerdev:
Yes, all was build with gradlew.
I add a task 'fatJar' to build the server/mapbuilder jar with depending classes.

EssBee59:
Thank for the info!
I could not find the task "fatjar" but I assume, you have currently no problem /question to gradle or fatjar.  (my brouter fatjar, built with gradle, is running well) 
regards

afischerdev:
I'll add this to the readme next update.

PR #344, opened 2021-10-11
GitHub Workflow: Provide Artifacts (APK & ZIP)

zod:
To simplify testing the result of each build is attached as an artifact to the workflow run.

It seems like the GitHub runners have the android SDK installed so it's possible to provide the APKs in addition to the brouter.zip. The GitHub UI is a bit [quirky](https://github.com/actions/upload-artifact/issues/3#issuecomment-743985727) and the brouter.zip download is zipped again before the download.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -16,11 +16,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 8
+    - name: Set up JDK 11
       uses: actions/setup-java@v2
       with:
-        java-version: '8'
+        java-version: '11'
         distribution: 'zulu'
         cache: gradle
+    - name: Create local.properties
+      run: touch local.properties
     - name: Build with Gradle
       run: ./gradlew build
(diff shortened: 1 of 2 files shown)


afischerdev:
- The definition of the depend in gradle file is fine. But I still would expect a fat jar file in the artefacts - I could use it for testing the server or the map generation. It's good to know that the process is running well, but do we need two files with unsigned apks as artifacts?
- The publishing process is a thing for the owner - he carries the key for signing. But wouldn't it be useful the have a feature to run the workflow manually as well? E.g. I create a new release, zip is generated with unsigned apks. The keys owner has to redo to complete.

zod:
I've updated the branch to only upload the `brouter-*.zip` as artifact.

EssBee59:
Hello,
As encountered in #364 , an error occurs on my Android 11 phone by installing the unsigned apk:
adb install BRouterApp.....apk
3015 KB/s (1880673 bytes in 0.609s)
Failure [INSTALL_PARSE_FAILED_NO_CERTIFICATES: Failed collecting certificates for /data/app/vmdl470500224.tmp/base.apk: Failed to collect certificates from /data/app/vmdl470500224.tmp/base.apk: Attempt to get length of null array]

Is it possible to create a test version instalable on all phones?

zod:
This is a merged PR which indicates that it's finished. Please create a new issue and just reference this PR for context information.

PR #345, opened 2021-10-11
Add .editorconfig to harmonize indention

zod:
EditorConfig (https://editorconfig.org) helps maintain consistent coding
styles for multiple developers working on the same project across
various editors and IDEs.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+end_of_line = lf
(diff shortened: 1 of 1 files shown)


zod:
I know that it's frowned upon to propose style changes without some contributions but I think it helps the project because currently there are many files which have mixed indention settings. I think the code shoudn't be reformatted but when changing the file/block it could still help to have consistent indention of the code.

PR #348, opened 2021-10-15
Generate assets with gradle

zod:
The android app contains some assets which are manually generated from other files in the repository. This could also be automated with gradle to avoid human errors and always ship the latest files. I've implemented it as a first draft for the `profiles2.zip` to open up the discussion.

The downside of the current implementation is that building the android app now requires sed and bash which aren't normally available on Windows.


brouter-routing-app/build.gradle
```diff
@@ -94,3 +99,17 @@ dependencies {
     implementation project(':brouter-util')
 
 }
+
+task generateProfiles(type: Exec) {
+    commandLine = "../misc/scripts/generate_profile_variants.sh"
```
zod, line 104: This executes the `generate_profile_variants.sh` to generate all profiles. I think that's not the best way because the build now requires bash & sed.

brouter-routing-app/build.gradle
```diff
@@ -94,3 +99,17 @@ dependencies {
...
+}
+
+task generateProfilesZip(type: Zip) {
+    dependsOn generateProfiles
+    archiveFileName = "profiles2.zip"
+    from "../misc/profiles2"
+    exclude "all.brf"
+    exclude "car-vario.brf"
+    exclude "softaccess.brf"
```
zod, line 113: There are some profiles excluded from the zip.

afischerdev:
Great, thanks a lot

PR #349, opened 2021-10-17
Update rd5 sizes

zod:
Update rd5 sizes to current values using update `ReadSizes.java`. To avoid downloading all the segments from the server it just parses the index. This can be quite fragile but since `ReadSizes.java` isn't used frequently and only by the developers it can be adopted if needed.


misc/utils/ReadSizes.java
```diff
@@ -40,9 +41,48 @@ private static void scanExistingFiles(File dir) {
...
+    } catch (MalformedURLException e) {
+      System.out.println("Invalid URL");
+    } catch (ProtocolException e) {
+      System.out.println("Unable to download segment index");
+    } catch (IOException e) {
+      System.out.println("Unable to download segment index");
+    }
+  }
+
+  private static void parseAndUpdateTileSize(String indexLine) {
```
polyscias, line 65: I think it is good practice to document code that does parsing using a sample of what should be parsed, in this case:

`<a href="E0_N35.rd5">E0_N35.rd5</a>                                         17-Oct-2021 01:03            1290596`

Also mentioning that (typically) https://brouter.de/brouter/segments4/ is used would be good.

polyscias: Thanks, that's good to have!

zod:
@afischerdev Is there anything missing before this can be merged?

afischerdev:
Sorry, no, I missed this point

PR #352, opened 2021-10-21
Set preserveLegacyExternalStorage for easier upgrades

zod:
When updating from a previous app which has it's basedir on e.g.
/sdcard/emulated/0/brouter it's still possible to access the files on
android 11 with this flag.


zod:
@afischerdev Any objections to this change?

PR #354, opened 2021-10-22
Common minSdkVersion and shared implementation

zod:
The [`brouter_1_6_1.zip`](https://brouter.de/brouter_bin/brouter_1_6_1.zip) contains two apk: api10 and api28 which have different `targetSdk`/`minSdk` versions

| File | minSdkVersion | targetSdkVersion |
| - | - | - |
| `BRouter_api10.apk` | 10 | 10 |
| `BRouter_api28.apk` | 10 | 28 |

The build currently produces also two apk: api10 and api19 which have the followding `targetSdk`/`minSdk` versions

| File | minSdkVersion | targetSdkVersion |
| - | - | - |
| `BRouterApp.1.6.2-api10-release-unsigned.apk` | 10 | 19 |
| `BRouterApp.1.6.2-api19-release-unsigned.apk` | 19 | 30 |

So the `minSdkVersion` (which specifies the minimal android version required to run) increased for Play Store version and the `targetSdkVersion` changed for the legacy version. The filename suffix now specifies the `minSdkVersion` instead of the `targetSdkVersion` as it was previously.

This PR restores the old naming convention and increases the `minSdkVersion` for both variants. The higher `minSdkVersion` allows both apks to use the AndroidX library and therefore to merge the code of the different variants.

| File | minSdkVersion | targetSdkVersion |
| - | - | - |
| `BRouterApp.1.6.2-api19-release-unsigned.apk` | 14 | 19 |
| `BRouterApp.1.6.2-api30-release-unsigned.apk` | 14 | 30 |


afischerdev:
I'm not sure if this is true.
I compiled two versions (14-19, 14-30) as suggested.
Both versions have same size - contains same libs. 
And installed a 14-30 apk on a tablet with Android 4.4 (api 19) and a 14-19 apk on a phone with Android 10 (api 29). App runs well in both situations - except of a little warning like 'App designed for older systems'.

zod:
Merged as part of #356

PR #356, opened 2021-11-02
Improve permission handling

zod:
Instead of checking for targetSdkVersion and device version we can just check which permissions are granted by the android platform. This enables devices running <= API28 to use the "classic mode" of reading waypoints from map tool directories without installing the special `api10` APK.

This branch currently includes the commits from #352 and #354 to test if the SD card access is still possible on API30 devices and the implementation is easier with the merged implementation. If those other PRs are merged I'll rebase this PR. The most relevant commit is the [last one](https://github.com/abrensch/brouter/pull/356/commits/dc95984199bfad4b40e019bcc2ccb34d1a1ac455)


afischerdev:
I like the improvement.
But still not sure if using api 10 or moving forward to 14.
Still waiting for some other comments.

afischerdev:
@zod
What do you think, waiting until weekend?
I don't have the time for testing before.

zod:
Yes, I don't get any work done either. I think perhaps we can use the `CoordinateReaderInternal` always as a fallback and remove the `CoordinateReaderNone`

afischerdev:
Good idea. What do you think about doing one step more and add CoordinateReaderInternal  to the reader list?

Meanwhile the CoordinateReaderInternal is a fix point for my nogo*.gpx data. When I read about road working in my area I design a route there and export it to nogo folder. Naming: 
nogo_streetname_from_to_date.gpx

afischerdev:
@zod 
Do you have any additional code? I'm ready to start.
After merge I'll add some parts I was waiting for:

- better migration check
- update CoordinateReader
- ...

zod:
I've added the fallback to `CoordinateReaderInternal` and removed `CoordinateReaderNone`

PR #364, opened 2021-11-13
Add intent receiver for profiles

### LINKED ISSUE #362: Allow sending profiles to app
Hello

Till now, changing a standard profile or adding a custom-profile is a complex task (editing  and  saving the new .brf file in ./profiles2/. is only possible with a file manager)

I think, it is possible to add / change profiles in a simpler way, here an example:

-	New profiles (when not available on internet-pages) should be created and tested using the Brouter-web
(screen of PC is better, Brouter-web offers different ways to change a profile with options or full code)
-	A lot of possibilities exist to transfer the .brf file to the Smartphone (Mail, Messenger, etc.. by need a special function using QRcodes)
-	Now with a minor extention in Brouter-app, the .brf file could be saved in ./profiles2/. with a simple click on the attachement of the mail or messager

I tested successfully the necessary extention is the Brouter-app (add intent-filter in manifest.xml and a peace of java code to store the  ,brf in ./profiles2)  
Very easy to use… a further advantage is, no problem with access-rights (Android-11 !!)  as Brouter itself store the .brf file

By interest I can provide my manifest and the peace of java code

Regards
Ess Bee

_Originally posted by @EssBee59 in https://github.com/abrensch/brouter/issues/346#issuecomment-962500866_

Implementation: https://github.com/abrensch/brouter/issues/346#issuecomment-962642519

zod:
Fixes #362

Implementation from @EssBee59

### DIFF
--- a/brouter-routing-app/src/main/AndroidManifest.xml
+++ b/brouter-routing-app/src/main/AndroidManifest.xml
@@ -43,6 +43,8 @@
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
+                <data android:scheme="file" />
+                <data android:scheme="content" />
                 <data android:mimeType="text/plain" />
             </intent-filter>
 
--- a/brouter-routing-app/src/main/AndroidManifest.xml
+++ b/brouter-routing-app/src/main/AndroidManifest.xml
@@ -36,9 +36,9 @@
             android:theme="@android:style/Theme.NoTitleBar.Fullscreen"></activity>
         <activity
             android:name=".BImportActivity"
-            android:label="BRouter Profile Import"
+            android:label="Import Profile"
             android:exported="true"
-            android:theme="@android:style/Theme">
+            android:theme="@android:style/Theme.NoTitleBar">
             <!-- some apps (bluemail) do not recognize the .brf file extention, startactivity+intent is done for attachement with text/plain -->
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
--- a/brouter-routing-app/src/main/AndroidManifest.xml
+++ b/brouter-routing-app/src/main/AndroidManifest.xml
@@ -25,6 +25,51 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <!-- some apps (bluemail) do not recognize the .brf file extention, startactivity+intent is done for attachement with text/plain -->
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="text/plain"/>
+            </intent-filter>
+
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+                <category android:name="android.intent.category.LAUNCHER" />
+                <category android:name="android.intent.category.MULTIWINDOW_LAUNCHER" />
+            </intent-filter>
+            <!-- general intent to register .brf files -->
+            <intent-filter
+                android:label="@string/app_name"
+                android:priority="50">
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="file"/>
+                <data android:scheme="content"/>
+                <data android:host="*"/>
+                <data android:pathPattern=".*\\.brf" />
+                <data android:pathPattern=".*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\..*\\..*\\.brf" />
+            </intent-filter>
+
+            <intent-filter
+                android:label="@string/app_name"
+                android:priority="50">
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="file"/>
+                <data android:scheme="content"/>
+                <data android:host="*"/>
+                <data android:mimeType="*/*"/>
+                <data android:pathPattern=".*\\.brf" />
+                <data android:pathPattern=".*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\..*\\..*\\.brf" />
+            </intent-filter>
         </activity>
         <activity android:name=".BInstallerActivity"
             android:label="@string/app_name"
--- a/brouter-routing-app/src/main/AndroidManifest.xml
+++ b/brouter-routing-app/src/main/AndroidManifest.xml
@@ -16,37 +16,48 @@
         android:label="@string/app_name"
         android:allowBackup="false"
         android:preserveLegacyExternalStorage="true">
-        <activity android:name=".BRouterActivity"
+        <activity
+            android:name=".BRouterActivity"
             android:label="@string/app_name"
             android:exported="true"
-            android:screenOrientation="unspecified" 
+            android:screenOrientation="unspecified"
             android:theme="@android:style/Theme.NoTitleBar.Fullscreen">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+        </activity>
+        <activity
+            android:name=".BInstallerActivity"
+            android:label="@string/app_name"
+            android:screenOrientation="landscape"
+            android:launchMode="singleTask"
+            android:exported="true"
+            android:theme="@android:style/Theme.NoTitleBar.Fullscreen"></activity>
+        <activity
+            android:name=".BImportActivity"
+            android:label="BRouter Profile Import"
+            android:exported="true"
+            android:theme="@android:style/Theme">
             <!-- some apps (bluemail) do not recognize the .brf file extention, startactivity+intent is done for attachement with text/plain -->
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
-                <data android:mimeType="text/plain"/>
+                <data android:mimeType="text/plain" />
             </intent-filter>
 
-            <intent-filter>
-                <action android:name="android.intent.action.MAIN" />
-                <category android:name="android.intent.category.LAUNCHER" />
-                <category android:name="android.intent.category.MULTIWINDOW_LAUNCHER" />
-            </intent-filter>
             <!-- general intent to register .brf files -->
             <intent-filter
                 android:label="@string/app_name"
                 android:priority="50">
                 <action android:name="android.intent.action.VIEW" />
+
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
-                <data android:scheme="file"/>
-                <data android:scheme="content"/>
-                <data android:host="*"/>
+
+                <data android:scheme="file" />
+                <data android:scheme="content" />
+                <data android:host="*" />
                 <data android:pathPattern=".*\\.brf" />
                 <data android:pathPattern=".*\\..*\\.brf" />
                 <data android:pathPattern=".*\\..*\\..*\\.brf" />
@@ -58,12 +69,14 @@
                 android:label="@string/app_name"
                 android:priority="50">
                 <action android:name="android.intent.action.VIEW" />
+
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
-                <data android:scheme="file"/>
-                <data android:scheme="content"/>
-                <data android:host="*"/>
-                <data android:mimeType="*/*"/>
+
+                <data android:scheme="file" />
+                <data android:scheme="content" />
+                <data android:host="*" />
+                <data android:mimeType="*/*" />
                 <data android:pathPattern=".*\\.brf" />
                 <data android:pathPattern=".*\\..*\\.brf" />
                 <data android:pathPattern=".*\\..*\\..*\\.brf" />
@@ -71,23 +84,17 @@
                 <data android:pathPattern=".*\\..*\\..*\\..*\\..*\\.brf" />
             </intent-filter>
         </activity>
-        <activity android:name=".BInstallerActivity"
-            android:label="@string/app_name"
-            android:screenOrientation="landscape"
-            android:launchMode="singleTask"
-            android:exported="true"
-            android:theme="@android:style/Theme.NoTitleBar.Fullscreen">
-        </activity>
+
         <service
             android:exported="true"
             android:name=".BRouterService"
             android:enabled="true"
             android:process=":brouter_service" />
-        <service android:name="btools.routingapp.DownloadService"
+        <service
+            android:name="btools.routingapp.DownloadService"
             android:label="Download Service"
             android:icon="@mipmap/ic_launcher"
-            android:enabled="true"
-            />
+            android:enabled="true" />
 
     </application>
 </manifest>
--- a/brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
+++ b/brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
@@ -1,6 +1,11 @@
 package btools.routingapp;
 
+import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.lang.reflect.Method;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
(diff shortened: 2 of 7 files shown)


brouter-routing-app/src/main/AndroidManifest.xml
```diff
@@ -25,6 +25,51 @@
...
+                <data android:scheme="file"/>
+                <data android:scheme="content"/>
+                <data android:host="*"/>
+                <data android:mimeType="*/*"/>
+                <data android:pathPattern=".*\\.brf" />
+                <data android:pathPattern=".*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\..*\\..*\\.brf" />
+            </intent-filter>
```
zod, line 72: This seems like a duplicate of the code block before - is it really necessary?

brouter-routing-app/src/main/AndroidManifest.xml
```diff
@@ -25,6 +25,51 @@
...
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="text/plain"/>
+            </intent-filter>
+
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+                <category android:name="android.intent.category.LAUNCHER" />
+                <category android:name="android.intent.category.MULTIWINDOW_LAUNCHER" />
+            </intent-filter>
```
zod, line 39: Warning "Missing URL" - shoudn't it contain a `scheme`?

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
```diff
@@ -76,6 +87,24 @@ public void onCreate( Bundle savedInstanceState )
...
+      return;
+    }
+// end change
+
+    // Get an instance of the PowerManager
+    //mPowerManager = (PowerManager) getSystemService( POWER_SERVICE );
+
+    // Create a bright wake lock
+    //mWakeLock = mPowerManager.newWakeLock( PowerManager.SCREEN_BRIGHT_WAKE_LOCK, getClass().getName() );
+
```
zod, line 107: Unrelated changes

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
```diff
@@ -76,6 +87,24 @@ public void onCreate( Bundle savedInstanceState )
...
 
+    // CHANGE: consider start with Intent to import a profile
+    Intent intent = getIntent();
+    String action = intent.getAction();
+    String type = intent.getType();
+    System.out.println("Brouter START xxxxxx BrouterActivity");
+    if (Intent.ACTION_VIEW.equals(action) ) {
+      System.out.println("Brouter Intent.ACTION_VIEW detected");
+      import_profile(intent);
+      return;
```
zod, line 98: It doens't run any of the other code in the BRouterActivity. Perhaps it would make sense to move it to an own Activity?

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
```diff
@@ -730,4 +759,123 @@ public void onRequestPermissionsResult (int requestCode, String[] permissions, i
...
+          filesize = cursor.getLong(1);
+        }
+      } finally {
+        if (cursor != null)
+          cursor.close();
+      }
+      System.out.println("Brouter filename=" + filename + "\n file size=" + filesize);
+      scanResults.setText(scanResults.getText() + "File name=" + filename + "\nFile size=" + filesize + " bytes\n");
+      //  is the file extention ".brf" in the file name
+      if (filename.indexOf(".brf") != -1 && (filename.indexOf(".brf") == (filename.length() - 4))) {
```
zod, line 801: * filename could be null if the `query()` doesn't return any values
* Use the `endsWith()` method

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
```diff
@@ -730,4 +759,123 @@ public void onRequestPermissionsResult (int requestCode, String[] permissions, i
...
+        System.out.println("Brouter  OK, file extention is .brf!!!");
+      } else {
+        System.out.println("Brouter  ERROR, please provide a file with the valid extention \".brf\"...");
+        // report error in UI and stop
+        scanResults.setText(scanResults.getText() + "ERROR: File extention must be \".brf\" \n");
+        return;
+      }
+      // profile size is generally < 30 kb, so set max size to 100 kb
+      if (filesize > 100000) {
```
zod, line 810: The 100000 should be defined as static const in the class.

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
```diff
@@ -730,4 +759,123 @@ public void onRequestPermissionsResult (int requestCode, String[] permissions, i
...
+        InputStream input = cr.openInputStream(dataUri);
+
+        BufferedReader reader = new BufferedReader(
+          new InputStreamReader(input));
+        Profile_code = reader.lines().collect(Collectors.joining(
+          System.getProperty("line.separator"))).toString();
+        System.out.println("Brouter Profile_CODE=" + Profile_code);
+
+        // consistency check
+        if (Profile_code.indexOf("highway=") == -1 || (Profile_code.indexOf("costfactor") == -1) || (Profile_code.indexOf("---context:global") == -1)) {
```
zod, line 830: The check if a profile is valid shoud probably be done in brouter-core

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
```diff
@@ -730,4 +759,123 @@ public void onRequestPermissionsResult (int requestCode, String[] permissions, i
...
+
+      ContentResolver cr = getContentResolver();
+      String Profile_code = "";
+      try {
+        //    try to read the file
+        InputStream input = cr.openInputStream(dataUri);
+
+        BufferedReader reader = new BufferedReader(
+          new InputStreamReader(input));
+        Profile_code = reader.lines().collect(Collectors.joining(
```
zod, line 825: Requires at least API24 (see also [build log](https://github.com/abrensch/brouter/runs/4198052032?check_suite_focus=true))

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
```diff
@@ -730,4 +759,123 @@ public void onRequestPermissionsResult (int requestCode, String[] permissions, i
...
+      }
+      catch (Exception e)
+      {
+        System.out.println("Brouter  exception by read config15.dat  " + e );
+        scanResults.setText(scanResults.getText() + "ERROR: " + e + " /n");
+      }
+      finally
+      {
+        if ( configInput != null ) try { configInput.close(); } catch (Exception ee) {}
+      }
```
zod, line 859: Use `ConfigHelper.getBaseDir(this)`

brouter-routing-app/src/main/AndroidManifest.xml
```diff
@@ -25,6 +25,51 @@
...
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="file"/>
+                <data android:scheme="content"/>
+                <data android:host="*"/>
+                <data android:pathPattern=".*\\.brf" />
+                <data android:pathPattern=".*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\..*\\.brf" />
+                <data android:pathPattern=".*\\..*\\..*\\..*\\..*\\.brf" />
```
zod, line 54: Are those multiple pathPattern necessary?

zod (commented):
* Why does it use the `VIEW` intent instead of `SEND` and `SENDTO`. This would allow receiving multiple profiles (e.g. selected in a file manager)
* There are quite some print statements. If they should be kept, they should be migrated to Log statements.

EssBee59:
> I'm not reviewing my own PR since I'm acting as a PR-proxy for @EssBee59
> 
> * Why does it use the `VIEW` intent instead of `SEND` and `SENDTO`? This would allow receiving multiple profiles (e.g. selected in a file manager)
> * There are quite some print statements. If they should be kept, they should be migrated to Log statements.

Hello,
most of the apps are using "view" in the intent by clicking the attachement (we have no influence on this!).
So, to accept the intent in Brouter we have to use VIEW.

I did not test an import of "multiple profiles", possibly an error will occurs.

The log statements was for debug / test, they can be migrated (or cleared by need)

brouter-routing-app/src/main/AndroidManifest.xml
EssBee59: > This seems like a duplicate of the code block before - is it really necessary?

Yes, all apps (as example osmand registration of .gpx files) do so by registration!
(it is not exactly a duplicate!!!)

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
EssBee59: Clear this comments (I had to create the mWakeLock before calling the import_profile, else an error occured on the onresume,,,)

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
EssBee59: Sure, create an own activity is possible...
I found this entry point, and it was running.
But in any case, you have to test in the same way (if (Intent...) to start the import_profile...

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
EssBee59: Feal free to enhance the code and the stability of the method

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
EssBee59: Any enhancement is good

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
EssBee59: Sorry for this problem...
Can you use another compatible method?
(Think about the different types of intent, as example with BlueMail--- not all the read methods will run)

brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java
EssBee59: I am not really familiar with brouter...
Sorry!

brouter-routing-app/src/main/AndroidManifest.xml
EssBee59: I can not help here..

brouter-routing-app/src/main/AndroidManifest.xml
EssBee59: I think so, as it is used by quite every registration ...

EssBee59: Hello,
most of the apps are using "view" in the intent by clicking the attachement (we have no influence on this!).
So, to accept the intent in Brouter we have to use VIEW.

I did not test an import of "multiple profiles", possibly an error will occurs.

The log statements was for debug / test, they can be migrated (or cleared by need)

EssBee59: Not sure, but all the apps I could look at do so ....

EssBee59:
> I'm not reviewing my own PR since I'm acting as a PR-proxy for @EssBee59
> 
> * Why does it use the `VIEW` intent instead of `SEND` and `SENDTO`? This would allow receiving multiple profiles (e.g. selected in a file manager)
> * There are quite some print statements. If they should be kept, they should be migrated to Log statements.

Action "VIEW" in Android is used for "open with", action "SEND" is used for "share / send to".
(for Brouter "open with" is I think the right function)
SEND should be used as example to forwoard the file(s) with a mail or messenger app. Yes, SEND supports multiple files, but "Open with"/VIEW is possible with only 1 file. I think this is OK / the best behaviour?

Print statements were used for test/debug. They can be changed or cleared

EssBee59:
About the question ACTION_VIEW or ACTION_SEND in the intent I checked again and am quite sure that ACTION_VIEW is the right mode for us.
ACTION_SEND should be used when the file is to be attached to a mail or messaging app. In that case, multiple send is of course possible and suported.
When you intent a local action on the file (open with) then multiple files is not a very good idea!

Note:
Depending on their function, Android app´s should choose the right intent mode...but it is possible to register on both modes at the same time.
As example the GpxViewer registers GPX files on both modes... but if you send multiple files, you get only 1 opened...

BTW, i prepared last year a new app (brouterscan) that easilly transfer GPX files from the Brouter-web to Android navigation-apps.
To test it, create a track using this url:
http://brouter.de/essbee/#map=11/49.5771/8.5371/osm-mapnik-german_style
Then --> export --> generate QRCode   and follow the instructions to install and use the app

Now I hope that the brouter-web team will implement the QRCode feature
https://github.com/nrenner/brouter-web/issues/329
https://github.com/nrenner/brouter-web/pull/370

After implementaion in Brouter-web I could set the app brouterscann in production state on google play OR we move the fonctionality into the brouter-app!!!
(in this way, we had a promotion of brouter-app in the brouter-web AND in the reverse way too)

zod:
> About the question ACTION_VIEW or ACTION_SEND in the intent I checked again and am quite sure that ACTION_VIEW is the right mode for us. ACTION_SEND should be used when the file is to be attached to a mail or messaging app. In that case, multiple send is of course possible and suported. When you intent a local action on the file (open with) then multiple files is not a very good idea!
Okay, it also seems the default action is often ACTION_VIEW. To initiate a ACTION_SEND using e.g. Solid Explorer I have to do a long press to select the file and Share later. I'm not really sure what you mean by "When you intent a local action on the file (open with) then multiple files is not a very good idea!"

> Note: Depending on their function, Android app´s should choose the right intent mode...but it is possible to register on both modes at the same time. As example the GpxViewer registers GPX files on both modes... but if you send multiple files, you get only 1 opened...
I think that's a specific issue of GpxViewers implementation and not a general issue of sending multiple files.

> BTW, i prepared last year a new app (brouterscan) that easilly transfer GPX files from the Brouter-web to Android navigation-apps. To test it, create a track using this url: http://brouter.de/essbee/#map=11/49.5771/8.5371/osm-mapnik-german_style Then --> export --> generate QRCode and follow the instructions to install and use the app
> 
> Now I hope that the brouter-web team will implement the QRCode feature [nrenner/brouter-web#329](https://github.com/nrenner/brouter-web/issues/329) [nrenner/brouter-web#370](https://github.com/nrenner/brouter-web/pull/370)
> 
> After implementaion in Brouter-web I could set the app brouterscann in production state on google play OR we move the fonctionality into the brouter-app!!! (in this way, we had a promotion of brouter-app in the brouter-web AND in the reverse way too)
This is totally unrelated to this PR or the issue this PR addresses so I think you should create a new issue if you want to discuss this idea.

EssBee59:
> This is totally unrelated to this PR or the issue this PR addresses so I think you should create a new issue if you want to discuss this idea.

Thank for the fead back, sorry, I have to learn working with appropriates PR´s and be more patient
Regards
Ess Bee

brouter-routing-app/src/main/AndroidManifest.xml
zod: I've added file & content scheme to fix the warning.

zod:
I've fixed most of the issues and changed the UI a bit
* It is now a new Activity `BImportActivity`
* The profile isn't automatically imported when opening the file with the BRouter app but the user has to press the _Import Profile_ button to avoid accidentally importing a profile

You can download an APK [here](https://github.com/abrensch/brouter/actions/runs/1467549888) by clicking on the `ZIP`.

EssBee59:
> You can download an APK [here](https://github.com/abrensch/brouter/actions/runs/1467549888) by clicking on the `ZIP`.

Hello,
It looks very good (a very good idea to ask the user to confirm the import)
I liked to retest the api30 version on my Android 11 phone, but I get an error by installating the apk using a file manager.
I retried using "adb" and got the following error:

adb install BRouterApp.....apk
3015 KB/s (1880673 bytes in 0.609s)
Failure [INSTALL_PARSE_FAILED_NO_CERTIFICATES: Failed collecting certificates for /data/app/vmdl470500224.tmp/base.apk: Failed to collect certificates from /data/app/vmdl470500224.tmp/base.apk: Attempt to get length of null array]

Possibly a problem under Android 11.
Could you sign the app or send a debug version?
(Not a real problem if not possible, I could dowload the sources and generate locally) 
Regards
...

zod:
Those builds are created automatically. During implementation (#344) it was decided that the debug APKs are not required but perhaps this decision has to be revised.

I'd like to avoid creating an APK manually so it's best if you build it yourself.

EssBee59:
> I'd like to avoid creating an APK manually so it's best if you build it yourself.

I could build and sign the app:
All tests passed successfully !
(For me, only the luminosity of the display field "editTextProfile" is very low, but posiibly only on my test phone)

Regards

afischerdev:
Two remarks on this

```
if (Profile_code.indexOf("highway=") == -1 || 
    (Profile_code.indexOf("costfactor") == -1) || 
	(Profile_code.indexOf("---context:global") == -1)) {
```
Please don't test for 'highway' otherwise a variante of 'river.brf' will fail. Section test should be enough.

- save
Please make sure that no pre delivered profile is overwritten. The new rule is to check delivered profiles and lookup each download session. Otherwise it is changed back to the public profile on download data.

A list is found in 'serverconfig.txt' (check_profiles=...).

EssBee59:
Thank for your remarks that should be implented!
@ zod:
I could prepare some code for that if you want, but your code is much better ?
Ess Bee

> Two remarks on this
> 
> ```
> if (Profile_code.indexOf("highway=") == -1 || 
>     (Profile_code.indexOf("costfactor") == -1) || 
> 	(Profile_code.indexOf("---context:global") == -1)) {
> ```
> 
> Please don't test for 'highway' otherwise a variante of 'river.brf' will fail. Section test should be enough.
> 
> * save
>   Please make sure that no pre delivered profile is overwritten. The new rule is to check delivered profiles and lookup each download session. Otherwise it is changed back to the public profile on download data.
> 
> A list is found in 'serverconfig.txt' (check_profiles=...).

Thank for your remarks, I think they should be implented!
@ zod:
I could prepare some code for that if you want, but your code is much better... ?
Ess Bee

zod:
Thanks @afischerdev I'll implement them. I think the code for handling the information from `serverconfig.txt` should be moved from `DownloadService` to a separate class to avoid duplication. Should I do a new PR or just include it in this one?

afischerdev:
@zod 
Good idea to modularize it. 
From my point of view it's part of the 'intent receiver problem'. So I think no extra PR. 
On the other hand, we have some minor changes in the issues. Maybe we will get better if we use milestones in the future.

PR #385, opened 2021-12-31
Fix GitHub workflows

zod:
GitHub action secrets default to empty values if they aren't defined in a repository. Any fork of the repo doesn't have access to the secrets and the jobs therefore fail.

`gradle-publish.yml` has a step which doesn't contain any `run` or `uses` block. This file also contains signing information but we only build the packages not the apk. Do we really need the signing stuff?

There's also `gradle-publish-test.yml`. Can this be merged to `gradle-publish.yml`?

We currently use the [environments](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#about-environments) which according to documentation are for different deployments. We have only one deployment. What's the reason behind it?


abrensch:
>> There's also gradle-publish-test.yml. Can this be merged to gradle-publish.yml?

 gradle-publish-test.yml was just my struggle to get the signing via workflow running.
now that the regular gradle.yml also does this, there's no difference between them, so that gradle-publish-test.yml can be deleted.

gradle-publish.yml adds a publishing step, but I did not yet try to understand how that works

zod:
I've marked the PR as draft because there were some open questions. Shall we discuss theme here or shall I create a new issue for them?

afischerdev:
Ups, I haven't noticed the draft. I only looked at the 'delete' and the code.
But when you mean the gradle-publish.yml was never running as expected on release building, then it could be a new issue.

PR #390, opened 2022-01-07
Use layouts in DownloadManager

zod:
Currently DownloadManager has most of the logic in `BInstallerView`. This includes showing segment selection and download progress. Views should only be used for displaying information, not general logic like deleting files etc.

I've moved most logic `BInstallerActivity` and used layouts to show information/button and download progress.
- Replaced custom touch event handling with `GestureDetector` and `ScaleGestureDetector`
- Avoid moving map out of bounds
- Simplify drawing code

### DIFF
--- a/brouter-routing-app/build.gradle
+++ b/brouter-routing-app/build.gradle
@@ -93,6 +93,7 @@ android {
 dependencies {
 
     implementation 'androidx.appcompat:appcompat:1.3.1'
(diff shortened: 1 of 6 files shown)


afischerdev:
I'm sorry for the delay.
Thanks for the first step on new verson. I had a short test on new layout version.

When scaling down the map is flickering on my phone and moving from corner to corner. 
And the map can be scaled down until it's like a one pixel pic. What about a stop on e.g. 50%?

PR #392, opened 2022-01-15
no turnrestrictions for footmode

vodie:
With turnrestrictions on in footmode we get a wrong route:
https://github.com/nrenner/brouter-web/issues/469


abrensch:
i addition, I added a patch to use TR bike exceptions also for foot-mode

afischerdev:
@abrensch 
> i addition, I added a patch to use TR bike exceptions also for foot-mode

It sounds so easy, but I had a short look on the changes. Great, that's why I like this project so much.

Issue #402, opened 2022-02-23
highway=no is used for routing


hgcvm:
I noticed that the fast profile (perhaps others as well) use a highway=no for routing. 

An example of highway=no can be found here: https://www.openstreetmap.org/way/835341884

The tag highway=no is used to indicate there is no road or path here. https://wiki.openstreetmap.org/wiki/Tag:highway=no

(in this specific case, it's used for cycling highways that are supposed to be there one day, but planning and construction has not yet started)

polyscias:
Yes, support for highway=no is missing, the usage is quite low so no surprise it is missing.

Easy to add support for it, see #403

polyscias:
I thought it was easy to add support but I see that requires an update of lookups.dat so I dropped the pull request.

polyscias:
Did have a look at some Brouter suspects today and found a problem with a way that was tagged using access="no @ (2022 Feb 12-2022 March 14)". That is not correct, conditional access should be tagged with access:conditional=. The question is however why brouter is evaluating the access="no @ (2022 Feb 12-2022 March 14)" at all.

This issue is in that regards the same as highway=no (in way context) for this issue, both are not part of looksup.dat.

In the [profile_developers_guide](https://zod.github.io/brouter/developers/profile_developers_guide.html) you can find:

> if the data contains a value that is not contained in the lookup tables (but its key is known) that value is treated as unknown

and unknown is encoded as literally unknown in the .rd5 files, you can test in the profile on it using <key>=unknown.

With the trekking bicycle profile the value of the highway key is not for really checked for bicycle access, only highway=motorway|motorway_link|proposed|abandoned are forbidden, highway=no gets the [default cost of 2](https://github.com/abrensch/brouter/blob/96ab3bf5c2b2b21f7068c5e91387c43761f87269/misc/profiles2/trekking.brf#L256).

For the car-vario profile the "problem" is that for the access key [only the values yes|permissive|designated|destination give access](https://github.com/abrensch/brouter/blob/96ab3bf5c2b2b21f7068c5e91387c43761f87269/misc/profiles2/car-vario.brf#L66), everything else including unknown means no caraccess

polyscias:
Good idea to add highway=no to highway=abandoned|disused|razed|demolished|dismantled

If the hiking/walking profile are already routing over "abandoned" I see no big problem that is also done for highway=no.

mfaure:
[...]

PR #403, opened 2022-02-23
Add support for highway=no, see #402

polyscias:
See #402.

Not used that often, highway=no, 1912 time, but so is highway=abandoned, 2893 times.


polyscias (commented):
For some reason the first version of this patch (adding "|no") does not work. As far as I can see the syntax is correct and I also see highway=no is in lookups.dat:

`highway;0000000436 no`

Looking better I see that this highway=no definition is under ---context:node, so that explains why this patch is not working

PR #410, opened 2022-03-30
Use more railway lines for routing

### LINKED ISSUE #409: Rail profile should also follow metro lines
Metro rail is also rail 😀

The example below is supposed to follow the metro from Shin-Osaka metro station to Namba. There absolutely is a metro line between the two, without a line change. So it doesn't make sense to me why it wouldn't be able to follow this railway.

I have even tried with the ÖPNVKarte layer, to place the start/finish pins *exactly* on the metro line. It still managed not to find a route.

Example:
image

Also, the message is not correct. It should say "closer to a railway". (well it should in *this* particular scenario, but I mean in other scenarios where it would be true, you know what I mean, surely)

Here is the ÖPNVKarte layer as proof, notice the blue line:
image

And tramlines/streetcars:

http://brouter.de/brouter-web/#map=16/34.3936/132.4666/osm-mapnik-german_style&lonlats=132.474942,34.396623;132.455286,34.38838&profile=rail

image

zod:
Closes #409

### DIFF
--- a/misc/profiles2/rail.brf
+++ b/misc/profiles2/rail.brf
@@ -13,8 +13,8 @@ assign turncost 0
 assign initialcost 0
 
 assign costfactor
-  switch railway=rail 1
-  switch railway=narrow_gauge 1
+  switch railway=rail|light_rail|narrow_gauge 1
+  switch railway=tram|subway 2
   100000
 
 ---context:node  # following code refers to node tags


thany:
Good to see it having been fixed two week ago. Solution works as expected.

How long until the website picks up this change?...

zod:
I think there is no periodic deployment, but that's @abrensch or @nrenner territory

nrenner:
I updated the rail.brf on the server, so this now works:
http://brouter.de/brouter-web/#map=16/34.3936/132.4666/osm-mapnik-german_style&lonlats=132.474942,34.396623;132.455286,34.38838&profile=rail

PR #413, opened 2022-04-03
Use WorkManager for Downloads

zod:
I've looked at using DownloadManager to use android to provide background downloads but it didn't work out because we need special handling because of the diff updates. The modern android way is to use WorkManager which provides all the background handling on the different android versions.

Using WorkManager requires AppCompatActivity so I switched all activities to keep a consistent (and more modern) look. The progress for downloads is now also reported with a progress bar per file in the notification area (where you can also cancel it) and on top of the download manager view to give an indication.

This is still a bit work-in-progess because I didn't test the diff updates and some smaller issues but I'm looking for feedback if you like the rework.


afischerdev:
@zod 
Great work, I'm on the way to test your changes. 
So I like to report a problem with incomplete downloads: 
My test scenario: 
clear segements4 folder (*.rd5 only), order 2 tiles, start download, download note starts in status line, app looks ready with work, so I leave the app.
The current download stopps, after a few seconds the second download starts. At the end I have 2 files (second is correct, first is renamed *.tmp to *.rd5, but contains only parts).
A new start of download 'repairs' the situation.

zod:
I'm not sure why the download stopped when leaving the app, for me it worked fine and I couldn't reproduce it. But the check if the downloaded file is valid was incorrect and I've fixed it. This should avoid moving an incomplete download file to *.rd5.

afischerdev:
@zod 
Thanks for the update

>But the check if the downloaded file is valid was incorrect and I've fixed it. 

Works, no fragments any more.

But leaving app still closes the download process (test on A10 and A12).
When I remove the `System.exit(0);` in BInstallerActivity, [OnDestroy](https://github.com/zod/brouter/blob/02eddeff818e60d5156e6349e4f76aea7d931fe3/brouter-routing-app/src/main/java/btools/routingapp/BInstallerActivity.java#L208), the process runs as expected.
But I had to activate `finish();` in [BRouterActivity](https://github.com/zod/brouter/blob/02eddeff818e60d5156e6349e4f76aea7d931fe3/brouter-routing-app/src/main/java/btools/routingapp/BRouterActivity.java#L127) to close the app.

poutnikl:
I wonder, why not to set Brouter manager, after download is started, automatically going to background? Currently I do it manually, being some time ago pleasantly surprised it is possible now. ( downloading it for some time via Android termux with installed have, running bash script calling brouter.jar to do incremental download.)

zod:
Ah, I just opened the launcher but didn't use the back button to terminate the activity (which caused also a termination of the app). When using finish it would stop whole app, but now it shows the normal main dialog.

> I wonder, why not to set Brouter manager, after download is started, automatically going to background? Or, putting there a button for such an option.
You can't immediately calculate a route and not everyone notices the progress in the notification so it's better to keep the screen open. But I think the whole dialog based concept should be overhauled but that's a different task.

> Currently I do it manually, being some time ago pleasantly surprised it is possible now.
It was implemented by @afischerdev for the last release and I've extended it to use WorkManager

afischerdev:
> You can't immediately calculate a route and not everyone notices the progress in the notification so it's better to keep the screen open. 

In the current production version you can't leave the app on back button. But task switcher will work. And it is possible to route on a tile which is not updated/downloaded at the moment.

On the new dev version: when I don't leave the app I can work as on start: activate new tiles, delete one. Seems to work all with the new worker. Great.

I would prefer the finish version after press back in download manager. But this is personal opinion and belongs to my workflow. I update my tiles from time to time, but changes on profiles I do in the client app.

> But I think the whole dialog based concept should be overhauled but that's a different task.

May be, but it has nothing to do with the download process. But this is personal opinion.

By the way, outfit: 
The download box should inform about the current tile and kind of process: download or update. But additional remarks like checksum should be removed from user info to debug info.

Screenshot_20220406-101304

poutnikl:
The task switcher is what I do, watching the progress on notification area. But, BR being landscaped, it is  inconvenient for me to do so, mostly due the phone case edge. IMHO the landscape is not really needed, but that is just my nitpicking.

afischerdev:
@zod 
I'm missing the [checkScripts() functions in DownloadService ](https://github.com/abrensch/brouter/blob/7fef58a7b6c0ae1bab6f71f395330e771134cd06/brouter-routing-app/src/main/java/btools/routingapp/DownloadService.java#L342).
This enables the change of lookups.dat without publishing a new app. See #414 as a sample.
It downloads the a new lookups.dat and then start download new tiles.

afischerdev:
> I've renamed the function to [downloadLookupAndProfiles](https://github.com/zod/brouter/blob/download-worker/brouter-routing-app/src/main/java/btools/routingapp/DownloadWorker.java#L149) but the content is still similar.

Great, I haven't tested now, but am I right, the function downloads the scripts every time? 
There was a size check before download. 

That brings me to another point. On download a new lookup table it could be a good idea to remove all *.rd5. Wasn't in the source before, I think.

zod:
> Great, I haven't tested now, but am I right, the function downloads the scripts every time? There was a size check before download.

I think the size check hurts more than it helps. If you change a value in a profile it won't be downloaded because the filesize didn't change. The profiles are also rather small (<500 KB) and segments updates are usually several MB so it shoudn't increase total transfer that much. We could check the [`Last-Modified`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified) header if we want to download only if the file is changed.

> That brings me to another point. On download a new lookup table it could be a good idea to remove all *.rd5. Wasn't in the source before, I think.

I'm not sure with the compatibility of `lookups.dat` to segments. I thought the files are still compatible if it's only a minor update. Just deleting the files might cause some confusion for the user. If they aren't compatible the user should be prompted to update those segments.

afischerdev:
You are right, size check is not very engaging.
But on Last-Modified what will you check against? E.g. we update lookups.dat next week. In two weeks you install BRouter on your new phone. Means lookups.dat on phone is newer then net version.
May be download all every time is best choise but not very elegant as well.

> minor updates

I think they will still work but may be in an other way as expected. E.g. the change is a new alias for access=yes. In old version the result was access=unknown.
We could inform the user when lookups.dat is modified that update for all is needed (when we know that).

zod:
> You are right, size check is not very engaging.
> But on Last-Modified what will you check against? E.g. we update lookups.dat next week. In two weeks you install BRouter on your new phone. Means lookups.dat on phone is newer then net version.

We can preserve the modification date when unpacking the profiles. This way the lookup has the modification date from creation and not from installation.

afischerdev:
> We can preserve the modification date when unpacking the profiles

Good idea. I had a short look into the last ZIP (18.4.) api30.apk. The profiles in the containing profile.zip also carry this date. But I'm sure the gzip will have an option to keep teh original file date.

zod:
> Good idea. I had a short look into the last ZIP (18.4.) api30.apk. The profiles in the containing profile.zip also carry this date. But I'm sure the gzip will have an option to keep teh original file date.

We should preserve the timestamp when unpacking the zip (at installation time). During build time it doesn't really matter, because the profiles/lookup in the repository should be the latest version anyway. I think it's an optimization to slightly reduce the download size which can be implemented later.

I think the implementation of this feature is now complete and can be merged.

afischerdev:
> I think it's an optimization to slightly reduce the download size which can be implemented later.

I agree. 
Main problem in this case is the git checkout. All files get the checkout date. Ok, we are not alone with the problem. But so far I haven't found a viable solution for git action as well.

PR #422, opened 2022-04-17
Update locus xml generation to have only one extension section

polyscias:
I saw https://groups.google.com/g/osm-android-bikerouting/c/htEAGAOsBkc coming by and had a look at the details.

I could reproduce the XML validation problem using Renner_original.gpx but by adding http://www.garmin.com/xmlschemas/GpxExtensions/v3 as schema that problem is gone

### DIFF
--- a/brouter-core/src/main/java/btools/router/OsmTrack.java
+++ b/brouter-core/src/main/java/btools/router/OsmTrack.java
@@ -591,8 +591,10 @@ public final class OsmTrack
 
     if ( turnInstructionMode == 2 )
     {
-      sb.append( "  <extensions><locus:rteComputeType>" ).append( "" + voiceHints.getLocusRouteType() ).append( "</locus:rteComputeType></extensions>\n" );
-      sb.append( "  <extensions><locus:rteSimpleRoundabouts>1</locus:rteSimpleRoundabouts></extensions>\n" );
+      sb.append( "  <extensions>\n" );
+      sb.append( "   <locus:rteComputeType>" ).append( "" + voiceHints.getLocusRouteType() ).append( "</locus:rteComputeType>\n" );
+      sb.append( "   <locus:rteSimpleRoundabouts>1</locus:rteSimpleRoundabouts>\n" );
+      sb.append( "  </extensions>\n" );
     }
 
     sb.append( "  <trkseg>\n" );


afischerdev:
I don't think this will fix the problem. 
On my test generation Notepad++ and Truugo brings an error (Truugo: 'must contain namespace URI & URL' ). Means to me: this should be added to xsi:schemaLocation

`
http://www.garmin.com/xmlschemas/GpxExtensions/v3 https://www8.garmin.com/xmlschemas/GpxExtensionsv3.xsd
`

The main reported problem was before 'trkseg' tag are two 'extensions' tags where only one is expected. Please [see ](http://www.topografix.com/GPX/1/1/#type_trkType)
`

polyscias:
Yes, correct, the XML Validator says "Syntax Welformed" but also "xsi:schemaLocation must contain namespace URI & URL in pairs. Did you use a space as a delimiter?"

Let me updated the patch

afischerdev:
> Did you use a space as a delimiter?

Yes, but but update has a problem: 'http://www.garmin.com/xmlschemas/GpxExtensions/v3' is defined twice. 

And again:
The main problem is that we have two 'extensions' tags before 'trkseg' tag. 
There has to be only one. [Line](https://github.com/polyscias/brouter/blob/cc644bc10447e1b95349dfb7c50f3ed79285c11b/brouter-core/src/main/java/btools/router/OsmTrack.java#L595): the locus tags can be both inside one 'extensions' tag.

polyscias:
Yes, reviewing the .xsd's the missing garmin xsd in not the problem and yes, the original post already indicated the problem is two extension tags.

I was unsure as in http://www.topografix.com/GPX/1/1/gpx.xsd I see:

```
<xsd:element name="extensions" type="extensionsType" minOccurs="0">
  <xsd:annotation>
    <xsd:documentation>
     </xsd:documentation>
  </xsd:annotation>
</xsd:element>
```

No maxOccurs= but checking I read the default value of maxOccurs is 1,

I did updated the patch, but the title of this PR is not matching what is done anymore so let me close it and generate a new one.

zod:
You can also just use the edit button near the PR title to change it

polyscias:
Thanks for the hint @zod, I did so.

The patch now updates the Locus XML generation so there is only one extension section.

With this patch the XML is valid where is was not before and, see [the post](https://groups.google.com/g/osm-android-bikerouting/c/htEAGAOsBkc), with it no longer Garmin programs report the generated .gpx as being corrupt

Issue #416, opened 2022-04-09
Update for lookups.dat


afischerdev:
Initial with #414, we like to update the lookups.dat and start the collection

Add 'cycleway:both' (#414)

```
cycleway:both;0000000001 lane
cycleway:both;0000000001 shared_lane
cycleway:both;0000000001 no none
cycleway:both;0000000001 track
cycleway:both;0000000001 separate
cycleway:both;0000000001 share_busway
cycleway:both;0000000001 shared
cycleway:both;0000000001 yes
```

Add 'cycleway:lane' (#398)
see wiki https://wiki.openstreetmap.org/wiki/Key%3Acycleway%3Alane

```
cycleway:lane;0000000001 advisory
cycleway:lane;0000000001 exclusive
cycleway:lane;0000000001 pictogram

cycleway:both:lane;0000000001 advisory
cycleway:both:lane;0000000001 exclusive
cycleway:both:lane;0000000001 pictogram

cycleway:right:lane;0000000001 advisory
cycleway:right:lane;0000000001 exclusive
cycleway:right:lane;0000000001 pictogram

cycleway:left:lane;0000000001 advisory
cycleway:left:lane;0000000001 exclusive
cycleway:left:lane;0000000001 pictogram
```
Change 'highway=no' (#402)

```
highway;0000002942 abandoned disused razed demolished dismantled no
```

Change 'traffic_calming' removed * (1x)

```
traffic_calming;0000006808 yes *
```

afischerdev:
Add 'unhewn_cobblestone' (#270)

```
surface;0000113800 cobblestone cobblestone:flattened unhewn_cobblestone
cycleway:surface;0000000011 cobblestone cobblestone:flattened unhewn_cobblestone
```

EssBee59:
Great work!

As explained in #414, I think this update has for me first priority (number of occurences) and should be implemented.

On the other side, it is probably not possible to support/implement all the requests related to lookup/ new OSM tags:

I think many impacts should be considered: as example the workload, the impact on the size of the rd5 files, the procesing time of a script where the number of checked tags grows and grows
 
By tags with  low/middle occurences (my old requests #241 and #271) I am today not sure, it was a good idea to open them
(of course they are intersting for my own use)

polyscias:
Nice start indeed!

```
traffic_sign:direction;0000000001 forward
traffic_sign:direction;0000000001 backward
traffic_sign:direction;0000000001 both
```
[...]

PR #424, opened 2022-04-24
Update hiking profile

zod:
The hiking profile is quite outdated. Android users can manually get there preferred version from https://github.com/poutnikl/Brouter-profiles but it would be nice for users of brouter-web if they can use the current profile without manually uploading it.

I've used the SAC3 version because it seemed that hiking-beta was an older version of the SAC3 profile. To inform users that they should expect some smaller paths I've named it `hiking-mountain` but removed the SAC information as it might be confusing.

@poutnikl I didn't find any license information on your repository, but I guess since the profile was already used in the past you are also ok with an updated version.

@nrenner I've just had a look how brouter-web uses the profiles and noticed that renaming the profile would require updating brouter-web and all translations to remove the beta term. Should we keep the old filename?


poutnikl:
The hiking-beta profile was an early beta version of what later become the profile template for all my hiking profiles that are just the generated parametrized template from the respective GitHub hiking template  project.

nrenner:
> renaming the profile would require updating brouter-web and all translations to remove the beta term. Should we keep the old filename?

I don't mind the rename and would want to remove the ` (beta)` in any case.

zod:
@abrensch or @nrenner
I'm not sure who has access to https://brouter.de/brouter/profiles2/ but it would be great to provide `hiking-mountain.brf` because the update mechanism of the android app fails otherwise since #426 was merged.

afischerdev:
And I would add, may be there is a general way for publishing. eg. may be #416 comes to an end and would like see the world.

nrenner:
@zod Done.

Just curious, do you need this for own installation from master?

> the update mechanism of the android app fails otherwise 

Shouldn't that be made more stable then?

PR #426, opened 2022-04-24
Change config for hiking-mountain

afischerdev:
Add new profile to server config file and remove old hiking script.


zod:
Sorry, I forget that we use the filename also in configuration. Just in case you didn't know: You can also add commits to PRs of other people if they allow it.

I think perhaps we can also generate this zip automatically like the `profiles2.zip`. Do we need the `serverconfig.txt` as asset or just the `segments4.zip`?

afischerdev:
> Just in case you didn't know: You can also add commits to PRs of other people if they allow it.

Yes, I did not know that.

> I think perhaps we can also generate this zip automatically like the profiles2.zip. Do we need the serverconfig.txt as asset or just the segments4.zip?

Only the zip files are required for the app installation.
But I think generating those zip files is the better/clearer way than keeping the zips in git folder.

PR #427, opened 2022-04-24
Update README.md

arturhooks5:
added the command for downloading segments

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,11 @@ Segments files from the whole planet are generated weekly at
 You can download one or more segments files, covering the area of the planet
 you want to route, into the `misc/segments4` directory.
 
+There is the command to download all segments to current folder (Linux):
+```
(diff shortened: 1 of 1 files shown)


polyscias:
I am wondering what is the use case for downloading all segments.

brouter is not good for long routes and normally only one segment is enough unless you live on the edge of segments.

arturhooks5:
I'm using brouter on my website - so I want to update all earth by cronjob. That's the case.

arturhooks5:
@poutnikl In my case, I'm updating segments every 6 months... Ofc if somebody wants to kill the source server, I can't stop them and I don't think that adding or deleting such script can avoid it.
Maybe I should add downloading in the one stream by default.

polyscias:
Understood, for running a brouter-web server having all segments makes sense.

Still I think it is not wise to add this patch, whoever is using brouter on a website should know how to do this and adding it to README.md will enable people without a need to execute this. How about dropping this PR?

arturhooks5:
@polyscias, if my case is so rare maybe you are right and you should drop it

polyscias:
That are not that many people that run their own brouter server on their website I am sure.

Like written, I find the risk that incorporating the command in the README triggers people who not really need it to execute a full download while people that need it should be smart enough to find it themselves.

PR #433, opened 2022-05-22
Repair BRouterWorker

afischerdev:
To get same results for nogo lines and areas both (server and Android version) should use same routine. So I removed some test code.


zod:
It would be great to add an automated test to ensure not breaking the feature in the future.

afischerdev:
The request is understandable. But is there a sensible idea to protect against forgetting test code?

Example:
we have a test routine for
result = getData(a param);
Now when I want to try something new with it, my new call is getDataNew(a param) or something like that. getData still sends back beautiful test data. But the app still works with getDataNew data when the change back is forgotten.

Issue #429, opened 2022-05-03
Use intents to replace app specific coordinate readers


zod:
We currently have app specific implementations to read points of interest from app specific files and write calculated tracks to their directories. Those implementations exist for LocusMap, OsmAnd and OruxMaps. Those apps also support using BRouter using the service interface, which is the only way to use BRouter on newer android versions and our api30 build. The service interface offers some smaller drawbacks, because No-Go areas aren't possible in all those apps.

As different builds cause an increased maintenance and also confuse users I suggest to remove those app specific coordinate readers from BRouter. To still support the classic workflow we should switch to intent based communication with those apps. The app should share the waypoints with BRouter which can import them into its storage and after route calculation provide the track to the app via intents.

I've implemented a small proof of concept and recorded a video to show how users would interact with the app (e.g. OsmAnd) and BRouter to calculate a route.

https://user-images.githubusercontent.com/49181/166556496-918e74be-90f2-4c1c-96b8-0311fabb6c4e.mp4

If we remove those readers we can also drop our api19 build because we no longer need to write to other app's storage.

polyscias:
I think that is a pretty good idea, I never liked the App specific way of doing things and I think this one is more intuitive.

Maybe the Brouter App can be streamlined so that if it receives such a share, it immediately pops in to the selection of the profile and the waypoints, skipping the selection of the Main Action (Download Manager/BRouter App) and the Select Action (Serve-mode/Select From).

Are the shared WayPoints saved in Brouter context?

EssBee59:
Yes good idea and solution...
But same opinion as Polyscias, in this version too long / not easy to use.

I had the same requirements by using Osmand + Brouter app and choosed 2 years ago an other aproach:
- I firstly enhanced the interface Osmand-Brouter app with the new parameter "profile name" 
(This parameter standard in the last version of Brouter-app)
- In a second step I added as further parameter the content of the Osmand "favourites.gpx" file

The code added in Osmand and Brouter is very limited, of course I have to support it currently by my own.

But I can in Osmand address every profile stored in Brouter-app, and define whether nogo´s are to be considered for routing. (api30 is well running, no access problem at all).

zod:
> Maybe the Brouter App can be streamlined so that if it receives such a share, it immediately pops in to the selection of the profile and the waypoints, skipping the selection of the Main Action (Download Manager/BRouter App) and the Select Action (Serve-mode/Select From).

[...]

PR #439, opened 2022-06-02
Remove coordinate readers

zod:
As discussed in #429 this removes the app specific coordinate readers and allows sharing the GPX track via intents.

Because access to other apps' storage is no longer required I've also removed the api19 build variant & WpDatabaseScanner.


zod:
Thanks for the feedback. I've updated the branch and BRouter now finishes (like it did before when pressing after routing calculation) which should avoid the blank screen.

PR #449, opened 2022-07-13
Reformat codebase

zod:
In #446 checkstyle was added to the build but most results suppressed because the code didn't match the expectations.

This reformats all code using the android studio default style and disables the suppressed.


afischerdev:
I had a view on the file changes, that looks allright.
But I'm not sure with the checkstyle workflow . Does it mean all incoming Java file are checked and reported/changed?
Or does it only work with Android Studio as editor?

afischerdev (approved):
@zod 
Is there a chance to change the rules for switch/case command?
Sample: The VoiceHint.java has several cases in one line. Useful for comparing value and result.

zod:
> But I'm not sure with the checkstyle workflow . Does it mean all incoming Java file are checked and reported/changed?
> Or does it only work with Android Studio as editor?

It's integrated into gradle. Running `./gradlew check` shows all errors reported by checkstyle. There are also more tasks available ([checkstyle docs](https://docs.gradle.org/current/userguide/checkstyle_plugin.html)). Because `check` is also part of `./gradlew build` our github action will also report those errors.

> Is there a chance to change the rules for switch/case command?

Unfortunately [IntelliJ does not support it](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360006505539-Do-not-break-switch-case) without manually specifying `//@formatter:off` for this specific region.

afischerdev:
@zod 
Thanks for the clearing.
Did you note that there are now some file conflicts?

zod:
I've rebased the branch onto master and reformatted again. It should now apply cleanly.

zod:
I've just noticed the failing build but it's too late for today. Will fixup tomorrow.

PR #456, opened 2022-08-03
Test voiceHints

zod:
This is just an initial version to start a discussion if/how we can test the voicehint code.

It has several drawbacks
- Requires data from another testcase (`MapcreatorTest`) which is really bad. Tests should not depend on each other, but it was the quickest way to get some segment files without adding large data to the repository
- Requires route calculation. Looking at the code it should be possible to store a `OsmTrack` object as binary file but then we shift the issue to test data creation which would require segment files.

Let me know if we should extend those tests somehow to make them sensible or if you think it's a waste of time.


EssBee59:
Hello zod,

I agree, tests automation should be discussed, it is not waste of time!

I am not expert in Junit or component tests, so I just like to explain what I would pragmatically do: 

-	Create a set of test-cases (I delivered a first version)
-	Try to execute these tests with the minimum effort

The manual execution of the 15 test-cases will take… 15 minutes ??
But an automated execution will take 2 seconds??

As automated execution I would create a unix script doing the following :
For each test-case
1-Fire the URL direct to the brouter JAVA server, as example for test-case 1 :
http://testsystem:17777/brouter?lonlats=16.254963,47.790453|16.261663,47.799066&profile=car-fast_Osmand&alternativeidx=0&format=gpx&trackname=testcase1

2-check the downloaded gpx file (look for  <turn>KL</turn> )

As explained, better solutions with junit or other tools are perhaps possible, this is just what I can suggest

zod:
It seems you didn't look at the changes proposed in this PR.

JUnit tests are written in Java and are usually superior to shell scripts because IDEs offer integration which allows selecting tests and analyzing the results. They are also integrated into our gradle build which makes them easily executable during our GitHub action workflows. Tests should normally only run those parts which they try to test (unit test) and not other stuff like the HTTP server (system test).

The downside is that we can't just assume that segments file exists. Therefore I've added some (really small) segments which are generated from the `dreieich.osm.gz` which is already part of the repository. If we can produce test cases using those segments we can easily run them without any requirements to run a server over HTTP.

afischerdev:
Why not using the standard junit test routines? Open a class, add some parameter, call a function, test result?

Sample:
```
 @Test
  public void testVoiceHint() {
    
    // #375
    // http://brouter.de/brouter-web/#map=18/49.86998/8.57579/osm-mapnik-german_style&lonlats=8.575778,49.870421;8.574706,49.870072
    
    VoiceHint vh = new VoiceHint();
    vh.angle = 101;
    vh.goodWay = new MessageData();
    vh.goodWay.turnangle = 101;

    MessageData bad = new MessageData();
    bad.turnangle = -91;
    vh.addBadWay(bad);

    bad = new MessageData();
    bad.turnangle = -115;
    vh.addBadWay(bad);

    bad = new MessageData();
    bad.turnangle = 5;
    vh.addBadWay(bad);

    bad = new MessageData();
    bad.turnangle = 61;
    vh.addBadWay(bad);

    vh.calcCommand();
    assertEquals(
        "Voice hint result 1",
        VoiceHint.TSHR,
        vh.cmd
    );
  }

```

zod:
It still uses JUnit, but with the hamcrest matchers which I think are more readable and provide better output when tests fail.

In general I prefer unit tests like your suggestion but since there are large parts of untested code it was easier to write an integration test which tests more parts of the code, e.g. MessageData, VoiceHint and app specific formating (OsmAnd, Locus, ...).

afischerdev:
Ok, I had a look at the dreieich osm file. It contains more that I thought - a roundabout as well.
We have to learn if this all fits the problems we found outside this area.
Lets add some more tests.

EssBee59:
> Limit <bounds minlat="49.9934800" minlon="8.7012100" maxlat="50.0098900" maxlon="8.7255500"/>

The limits define a square of 1.7 km, right?
As it is very very small, it think it will be hard to find the selected test-cases

afischerdev:
@EssBee59 
When you ever had run a gradle process:
It generates the rd5 file for testing (E5_N45.rd5,E5_N50.rd5).
You will find it in this folder
`.\brouter-map-creator\build\resources\test\tmp\segments`

EssBee59:
> You will find it in this folder

Thank you: I could install the mini rd5 files, my first tests confirm that the covered area is very very small.
... I think, only a very limited among of the test cases can be moved within this area. 
The size of the 2 mini rd5 files is 7 and 8 kb: really small!
could we afford to create greater test-rd5? 

else, other possible solutions?

afischerdev:
@EssBee59 
As you see on my request #467 the new data size is growing. I don't know if there is a limit. But it makes sense for me to keep the data file small to become shorter test times.

But I also see the 'six ends juction' tests are not possible in this area. 
A second data set could be an idea. 
Or as I mentioned before use the standard junit test routines special for angle tests.
The other rules like prio check on crossing bigger roads, roundabouts exits, concatenation of voice hints or not and more are possible with the current data set.

zod:
I thought about your comment again and agree that we should focus more on unit tests. It's just easier to create reproducible tests this way.

We could still add some integration tests but not try to check all the possibilities there.

PR #458, opened 2022-08-04
Update lookups

afischerdev:
This is the call for a bigger update on lookups.dat

The Android apk doesn't need an update for that. Lookups.dat and profiles are updated automaticly.
Next apk update will then contain the new versions.

So only web services are affected.

And an update for the profiles is needed when there is an interest on man_made=pier or other changes.

Please see #416 for more.


zod:
If we generate new segments using this update `lookups.dat` all users must update their segment files as soon as they get the new `lookup.dat`. Correct?

* I'm not to sure about compatibility. When do we need to increase the major version?
* I don't think we need `hiking-beta.brf` in the repository, but just as downloadable profile.

afischerdev:
> If we generate new segments using this update lookups.dat all users must update their segment files as soon as they get the new lookup.dat. Correct?

Yes. When the server update is done and an apk user updates the data lookups.dat and profiles are downloaded first then rd5 download starts. But when user has called e.g. only one rd5 but has 4 on board he has a problem when next enter an outdated rd5 - he will get a notice.
We already discussed if we drop the others or not.

>    I'm not to sure about compatibility. When do we need to increase the major version?

When index changes a new number is needed. Smaller changes like aliases don't change it.

>    I don't think we need hiking-beta.brf in the repository, but just as downloadable profile.

Yes, until new apk version is out. Current looks for hiking-beta
But we could exclude it from profile zip

polyscias:
I think breaking things for (many) users is bad and like I wrote in [#416](https://github.com/abrensch/brouter/issues/416#issuecomment-1100707613), I see quite some pain for removing oneway=true.

I did post PR to update the profiles of @poutnik, see poutnikl/Trekking-Poutnik/pull/30 but that has not been merged as of now.

Apart from the profiles of @poutnik it also impacts most other profiles so I think it is better to keep it in the profile for now, but let's remove it for the brouter profiles.

> But when user has called e.g. only one rd5 but has 4 on board he has a problem when next enter an outdated rd5 - he will get a notice.

How does that notice look like? 
I run my tests with the server but there the error reporting is not good, the error is only logged at the server side.

Also, what about [return meta.lookupVersion == 10;](https://github.com/abrensch/brouter/blob/70c0e7ca3ebcd125b9c94ed2fa842745580525a9/brouter-routing-app/src/main/java/btools/routingapp/BRouterView.java#L330)

For the remainder the patch looks good to me, I can run later this weekend a map generation run with it and give some stats.

nrenner:
Quoting @polyscias from [#416 (comment)](https://github.com/abrensch/brouter/issues/416#issuecomment-1100011014):

> For tracktype=4, can we added "grade3-5" as alias, that is mapped 917 times

I wondered where the `tracktype=grade3-5` value comes from.

The [OSM tag history](http://taghistory.raifer.tech/#***/tracktype/grade3-5) graph (see also [Taginfo chronology](https://taginfo.openstreetmap.org/tags/tracktype=grade3-5#chronology)) looks suspicious:

OSM tag history graph for tracktype=grade3-5 showing two big steps

And indeed, there are two unique uses:
- 945 ways added by an import in 2014 with [changeset 21660337](https://www.openstreetmap.org/changeset/21660337) for some region in Norway [1]
- 538 ways added since July by a single paid mapper from Kaart [2], almost all to `highway=residential` in South Africa

The latter and other sporadic edits are all using the iD editor, which actually has that value in the list of suggestions when using the properties table, which is most probably because of the value count of that single import.

Therefore I think the `grade3-5` alias should not be included, such an endorsement would only further propagate the use of this undocumented and unsupported value.

----

[1] 
```sh
curl -o changeset-21660337.osc https://www.openstreetmap.org/api/0.6/changeset/21660337/download
grep "grade3-5" changeset-21660337.osc | wc -l
```
[2] https://overpass-turbo.eu/s/1kNX

polyscias:
Nice analysis @nrenner!

I am perfectly fine with removing the grade3-5 alias but my opinion is, also with knowing this story, such an alias is fine. 

There are more tags in lookups.dat that are undocumented, for instance noexit=yes in way context and junction=approach

poutnikl:
> I did post PR to update the profiles of @poutnik, see poutnikl/Trekking-Poutnik/pull/30 but that has not been merged as of now.

I will try to address it and generate new profiles, but after Aug 15, due vacation.

Another pain will be for LocusMap users, with profiles embedded to application or online LocusMap client. They use profiles from several sources.

polyscias:
> I have somehow missed the PR and related discussion. I will try to address it and generate new profiles, but after Aug 15, due vacation.

That would be nice.
 
> Another delay will be for LocusMap users, with profiles embedded to application or online LocusMap client. They use profiles from several sources.

Do you have some more information on this, in particular the exact profiles that are used? I like to add them to my collection of profile that I check for impact of this update.

>  I can run later this weekend a map generation run with it and give some stats.

With quite some delay I did run it and looking at all the .rd5 files generated versus those on http://brouter.de/brouter/segments4/
I see a mix of segments that did increase in size and decrease in size but net effect is that the overall size is -0.04%

The extremes:

```
W90_N30.rd5   4998766   34854799 +14.34%
E25_S10.rd5    477592    3252900 +14.68%
E45_S30.rd5    113831     768699 +14.81%
..
E10_N65.rd5   -505069    3792746 -13.32%
E15_N65.rd5   -952119    7259065 -13.12%
E10_N60.rd5  -3225188   23634055 -13.65%
```

I am planning to get the updated lookup.dat, profiles and one .rd5 on my mobile and see what it does.

polyscias:
See also [brouter-web#669](https://github.com/nrenner/brouter-web/issues/669), can we add also:

- railway=platform
- public_transport=platform
- man_made=pier

The first two are accessible on foot and a pier is often a way to access a ferry both on foot and by bicycle.

afischerdev:
@polyscias 
All your wishes are already done:

> * railway=platform

[update](https://github.com/afischerdev/brouter/blob/4cae6202ff84569602a58e6b1bb9c4a28eee07e4/misc/profiles2/lookups.dat#L404)
> * public_transport=platform

[update](https://github.com/afischerdev/brouter/blob/4cae6202ff84569602a58e6b1bb9c4a28eee07e4/misc/profiles2/lookups.dat#L323)

> * man_made=pier

[update](https://github.com/afischerdev/brouter/blob/4cae6202ff84569602a58e6b1bb9c4a28eee07e4/misc/profiles2/lookups.dat#L321)

afischerdev:
@abrensch @nrenner 
Are there any timelines for publishing this update?

abrensch (commented):
Do you really think it's neccessary to change the major lookup version?

That would intruduce a major hassle after rollout, see the lookup version check in https://github.com/abrensch/brouter/blob/master/brouter-mapaccess/src/main/java/btools/mapaccess/PhysicalFile.java and there's no pre-check on the major version in the download-manager

So it would mean to create a new download-location, update that in the download manager and keep the old major version rd5's for download by not-yet updated android apps

abrensch:
> My understanding until now was the minor version is for updates without index
> changes e.g. new alias values. Major version is for changes on the index. And that
> is what we have with this update.

adding values and adding tags are also changes compatible with minor updates

I see that lookups.dat from this PR is not minor compatible, e.g. due to the deletion of lines who's values as been aliased

But instead of deleting these lines we could just place dummy-values as placeholders

Opening Pandoras Box by sending incompatible RD5's into the world is something that needs more careful attention and engeneering, compared to making a lookups.dat patch minor-compatible

I will have a closer look but that wil take sone days

polyscias:
> But instead of deleting these lines we could just place dummy-values as placeholders

I did some analysis of all profiles I could find, see https://github.com/abrensch/brouter/issues/416#issuecomment-1100707613, I think non-used values that are not used by these profiles can be safely removed without triggering compatibility issues. 

I made a patch for the profiles of @poutnikl but that has not been merged yet (no blame), it shows that deleting certain unused values will indeed give compatibility issues and I think it is not worth that pain so I agree we better place dummy-values as placeholders for these particular values.

It would mean to retain (as dummy-values):

- oneway=true
- sac_scale=T1-hiking
- highway=toll_bridge
- highway=city_entry

I did leave out the keys for https://github.com/utack/utack_brouter_profile but created https://github.com/utack/utack_brouter_profile/pull/35

afischerdev:
I add a lookup without major change
All deleted lines have a dummy for the index
All new values are at the end of section.

polyscias:
I had a look at lookups_no_major.dat, it looks nice to me. Let me run in the weekend a rd5 creation test with it.

Based on the brouter_route_placeholder_dummy's I understand the order of key-values in the lookups.dat is how things are encoded.

What I am wondering is if it is not possible to replace "deleted" key-values that are now converted to brouter_route_placeholder_dummy's to new key-values. That would work for:

- highway, bicycle, motorcar and motorcycle and 1 key-value for vehicle in way context.
- highway, barrier and entrance in node context

Also brouter_route_placeholder_dummy 59..80 are not needed to my opinion.

I think, to prevent breaking profiles, oneway=true, sac_scale=T1-hiking, highway=toll_bridge and highway=city_entry should be retained (no brouter_route_placeholder_dummy) although their usage in OSM is zero or almost zero.

On the Android App: I think that the download manager should be updated so it detects that there is a new lookups.dat and if so gives the (as only) option to update lookups.dat and all tiles. That will prevent problems.

I am not worried about handmade profiles, like I wrote I did check all public available profiles I could find and no problems apart from those 4 mentioned key-values. I do think there are almost no or zero persons that have not published their handmade profile and are using keys that are deleted in this lookups_no_major.dat

afischerdev:
@polyscias 
Ok I add the breaking entries again.

polyscias:
I dd run mapcreation on a looksup.dat based on lookups_no_major.dat but with my proposed changes above.

It looks good, the total size of all .rd5's did go from 7878.0 MB to 7877.9 MB so a very small decrease in size.

On compatibility:

- Is there any way to test if a new lookups.dat is compatible with the older?
- Also if there is no update of the major version number I think it is bad if a user is running with a mix of old and new .rd5's.

afischerdev:
> Is there any way to test if a new lookups.dat is compatible with the older?

I would think we could test a new lookup against the standard profiles.

> Also if there is no update of the major version number I think it is bad if a user is running with a mix of old and new .rd5's.

I would also vote for the major change.

polyscias:
What I think would be ideal is if the App was updated that once people open the App to download/update segments, it detects there is a new lookup version, and pops up a screen explaining there is a of change the segment format. 

Then it asks if the user wants to update and if so, all existing segments will be deleted and have to downloaded again by the user. Optionally the App can offer to re-download all existing segments itself.

I think that is relatively easy to implement but as of now the App is not working this way and also if the App is changed say in the next month, it will not be that if this lookups update in 2 months everybody's App is updated.

In that sense, if we do not want to wait long with this lookup update, it is probably better to go with a compatible update.

I think that in the same time the App should be updated so in the future an update as described above can be done.

I use also QMapShack with brouter integration, I am wondering how that handles lookup updates, let me have a look at that.

abrensch:
>     * Also if there is no update of the major version number I think it is bad if a user is running with a mix of old and new .rd5's.

It's also bad without a lookup change to mix RD5s from different preprocessor runs

It was never supported, and in the past there were actual exceptions from that constellation.

Chances are that I mitigated them with my patch for the bee-line hack. Look at the change to OsmPath.java from this commit:

  https://github.com/abrensch/brouter/commit/eb716506fd934989177e8c3933b22c58b5a41574

The condition that I changed here from an exception to a "do nothing" is when someone changes a way direction of a way crossing a RD5-tile-boundary and using old/new files at that boundary. So there's no exception anymore but it's still wrong..

Technical background is that the tagging ("description") is encoded at "forward"-links only. But in that constellation a link is considered  "backward" at both terminating  nodes, so the tagging is not encoded anywhere.

But also for the more likely case that for a crossing link a terminating node is shifted to a slightly different position that link will just disappear, which is also wrong.

But in my opionion that status quo of just accepting that sort of "glitches" for the (unsupported) case of mixing RD5s from different preprocessor runs is o.k. It's better then detecting these conditions and delaing with the support cases.

afischerdev:
> I am not worried about handmade profiles, like I wrote I did check all public available profiles I could find and no problems apart from those 4 mentioned key-values. I do think there are almost no or zero persons that have not published their handmade profile and are using keys that are deleted in this lookups_no_major.dat

It is possible to check the lookups.dat to the used profiles like this:
`java -cp ../lib/brouter-1.6.3-all.jar btools.expressions.IntegrityCheckProfile profiles2/lookups.dat profiles2`

You will need the last master or the lib from [git action](https://github.com/abrensch/brouter/actions/runs/3465143219)

misc/profiles2/lookups.dat
```diff
@@ -30,34 +30,30 @@ highway;0000039003 platform
 highway;0000037192 proposed planned virtual
 highway;0000010307 raceway
 highway;0000003152 rest_area
-highway;0000002942 abandoned disused razed demolished dismantled
+highway;0000002942 abandoned disused razed demolished dismantled no
 highway;0000002631 services
 highway;0000002133 corridor
-highway;0000002093 crossing
```
quaelnix, line 36: It would be very unfortunate if this was removed. `highway=crossing` is used [more than 7 million times](https://taginfo.openstreetmap.org/tags/highway=crossing) and is useful in car, bike, and pedestrian profiles.

misc/profiles2/lookups.dat
```diff
@@ -642,53 +679,33 @@ sidewalk:right:bicycle;0000002667 yes designated
...
+construction;0000360177 road highway
+
 
  ---context:node
 
 highway;0001314954 bus_stop
 highway;0001130090 crossing
 highway;0001031274 turning_circle
-highway;0000609262 traffic_signals
+highway;0000609262 traffic_signals traffic_signals;crossing
```
quaelnix, line 708: [`traffic_signals;crossing`](https://taginfo.openstreetmap.org/tags/highway=traffic_signals%3Bcrossing) basically does not exist.

quaelnix:
I don't think that we should support false tagging practices like:
- [highway=traffic_signals;crossing](https://taginfo.openstreetmap.org/tags/highway=traffic_signals%3Bcrossing)
- [traffic_sign:direction=forward;backward](https://taginfo.openstreetmap.org/tags/traffic_sign%3Adirection=backward%3Bforward)
- [traffic_signals:direction=forward;backward](https://taginfo.openstreetmap.org/tags/traffic_signals%3Adirection=backward%3Bforward)

misc/profiles2/lookups.dat
```diff
@@ -30,34 +30,30 @@ highway;0000039003 platform
...
+highway;0000002942 abandoned disused razed demolished dismantled no
 highway;0000002631 services
 highway;0000002133 corridor
-highway;0000002093 crossing
 highway;0000001440 bus_stop
-highway;0000001274 yes
-highway;0000000679 unsurfaced
-highway;0000000108 byway
-highway;0000000037 driveway
-highway;0000000021 mini_roundabout
```
quaelnix, line 42: The [`highway=mini_roundabout`](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dmini_roundabout#Mini-roundabouts) tag is used in cases where the island of a traffic circle is traversable. If it is removed, such traffic circles can no longer be detected, because there is no road with [`junction=roundabout`](https://wiki.openstreetmap.org/wiki/Tag:junction%3Droundabout#Roundabouts) there either.

misc/profiles2/lookups.dat
polyscias: Thanks for the review but please note that this is `highway=crossing` in way context, so, see the taginfo link that is only tagged 360 times right now and it looks to me like a tagging error, according to the [https://wiki.openstreetmap.org/wiki/Tag%3Ahighway%3Dcrossing](url) is it not allowed.

misc/profiles2/lookups.dat
polyscias: Same comment as above, this is in way context.

polyscias:
> I also think that we should not support false tagging practices like this:
> 
>     * [tracktype=grade3-5](https://taginfo.openstreetmap.org/tags/tracktype=grade3-5)
>     * [highway=traffic_signals;crossing](https://taginfo.openstreetmap.org/tags/highway=traffic_signals%3Bcrossing)
>     * [traffic_sign:direction=forward;backward](https://taginfo.openstreetmap.org/tags/traffic_sign%3Adirection=backward%3Bforward)
>     * [traffic_signals:direction=forward;backward](https://taginfo.openstreetmap.org/tags/traffic_signals%3Adirection=backward%3Bforward)

For tracktype=grade3-5 see also and [earlier comment](https://github.com/abrensch/brouter/pull/458#issuecomment-1207211472l), meanwhile the usage is higher.

Yes, these values are not documented within OSM but that is an OSM problem and that problem should be solved in OSM.

I added those value aliases based the usage, for brouter the goal is to support as much the data that is present.

misc/profiles2/lookups.dat
quaelnix: You are right, sorry!

quaelnix:
> for brouter the goal is to support as much the data that is present.

I think that's a legitimate point, and I agree that the last three of my examples do little to no harm, but I'm still not convinced that making `grade3-5` an alias of `"grade4` would be a wise idea.

misc/profiles2/lookups.dat
quaelnix: My mistake! Sorry for the unnecessary objection.

quaelnix:
Adding [kerb=*](https://wiki.openstreetmap.org/wiki/Key:kerb) to the lookup table would allow us to determine an appropriate penalty for `barrier=kerb`.

afischerdev:
To get this PR a little bit forward I made a test with an new app to handle the described problems on a version change.

It notified a version change and offer a select box

version_problem

It also allow only fully rd5 on version change.
What do you think?

polyscias:
Yes, it would be good to get this PR forward and such a Version Problem message is exactly what is needed. 
One suggestion: Instead of the word "data" I would use the word "tiles"

Apart from this I think we need a way to force a version update of the App. What now can happen is that the tile format is updated but the App is still old so this Version Problem screen will not be displayed and bad things will happen.

I did a bit of thinking on this and I think it would be good to add a "minappversion" to the lookups.dat file like:
```
---lookupversion:10
---minorversion:14
---minappversion:1.6.3
 
---context:way
```

With an update initiated through the App, the code should first download the new lookups.dat and check if it's the version is larger than the minappversion, if not, it should indicate the App should be updated and the update process should be exited.

polyscias:
> Adding [kerb=*](https://wiki.openstreetmap.org/wiki/Key:kerb) to the lookup table would allow us to determine an appropriate penalty for `barrier=kerb`.

The current lookups.dat has already wheelchair=* and in that light adding kerb=* would make sense I think.
```
kerb;0000000001 lowered flush;lowered lowered_and_sloped
kerb;0000000001 flush no none
kerb;0000000001 raised yes regular normal
kerb;0000000001 raised
```

afischerdev:
@polyscias 
Thanks for the reply. The wording for data is changed in my preview.

> With an update initiated through the App, the code should first download the new lookups.dat and check if it's the version is larger than the minappversion, if not, it should indicate the App should be updated and the update process should be exited.

The next app could have the lookup version check before download. So I'm not sure if we need an app check as well.
The problem is that we can only change the lookups.dat when a new app is online. 
Otherwise it is possible to publish a new lookup table now but this runs into the described trouble

quaelnix:
@polyscias, I would omit `kerb=yes`, which is equal to `kerb=` by definition:
> [Some sort of kerb is present, but it can't or hasn't yet been determined whether it is raised, lowered, flush, etc.]((https://wiki.openstreetmap.org/wiki/Key:kerb#Tagging))

and do it like this:
```
kerb;0000000001 lowered
kerb;0000000001 flush no
kerb;0000000001 raised
kerb;0000000001 rolled
```
`lowered` and `rolled` should be separated, as only the latter is an obstacle for a wheelchair.

polyscias:
> > With an update initiated through the App, the code should first download the new lookups.dat and check if it's the version is larger than the minappversion, if not, it should indicate the App should be updated and the update process should be exited.
> 
> The next app could have the lookup version check before download. So I'm not sure if we need an app check as well. The problem is that we can only change the lookups.dat when a new app is online. Otherwise it is possible to publish a new lookup table now but this runs into the described trouble.

Yes, the new App should be online before the lookup version update but this is not enough.

Suppose the App update is released 1 April and the updated lookups.dat and .rd5 tiles on 1 October.
Now an user tries on 15 October to upgrade it's tiles but his/her last App update was before 1 April. His/her old App will not pop up this new "Version Problem" dialog.

This is something that can be fixed using a `minappversion`. If this functionality was already available it would be possible to force that user to upgrade it's App version first.

It will not help likely for this update as it also needs "soaking time" but I think we should add it "now", so we can benefit from it after the "soacking time".

polyscias:
Based on the kerb request of @quaelnix and the chipseal request of @mcliquid I did update further my new version of lookups.dat, see [lookups update as of 20230122](https://github.com/polyscias/brouter/blob/bf812189a1c82a81db0b85f0125f42723c27d3d1/misc/profiles2/lookups.dat)

quaelnix:
We should probably also support [crossing:markings](https://taginfo.openstreetmap.org/keys/crossing:markings#values).

misc/profiles2_db/racebike-verylowtraffic.brf
```diff
@@ -146,11 +141,8 @@ assign badoneway =
 assign onewaypenalty =
        if ( badoneway ) then
        (
-         if ( cycleway=opposite|opposite_lane|opposite_track       ) then 0
-         else if ( cycleway:left=opposite|opposite_lane|opposite_track  ) then 0
-         else if ( cycleway:right=opposite|opposite_lane|opposite_track ) then 0
+         if ( cycleway=opposite|opposite_lane|opposite_track ) then 0
          else if ( oneway:bicycle=no                         ) then 0
-         else if ( junction=roundabout|circular              ) then 60
```
quaelnix: I think you were using an outdated version of this profile as base?

See: https://github.com/afischerdev/brouter/commit/5e217ff67bfe4612e94299ea301806e08c77d144

afischerdev: @EssBee59
@quaelnix is right. The racebike profile creates wrong ways [please see reference](http://148.251.191.149/brouter-web/#map=18/50.00054/8.70236/osm-mapnik-german_style&lonlats=8.700464,50.000788;8.702331,50.000761&profile=racebike-verylowtraffic)

quaelnix:
I think we should also add [path=crossing](https://taginfo.openstreetmap.org/tags/path=crossing) while we're at it.

afischerdev:
@quaelnix 
Ok, what about [path=sidewalk](https://taginfo.openstreetmap.org/tags/path=sidewalk#overview)?

quaelnix:
Good question. I probably wouldn't add it.

quaelnix:
There were multiple requrests to support [`railway=ferry`](https://taginfo.openstreetmap.org/tags/railway=ferry#overview) over the years. However this tag is only used 17 times world wide - at least so far.

quaelnix:
Would it be ok to remove `sac_scale=t1-hiking` or do we have to keep it for backwards compatibility?

afischerdev:
@quaelnix 
There are no more entries.

quaelnix:
I guess we should not only support cycleway:surface but also `footway:surface`.

laem:
Hi, if I understand well, to use these new tags in Brouter, the RD5 tiles must be regenerated with this new lookup file. 

The additions are very interesting for my use case. Thanks for your work ! 

 Is there a "next" branch where RD5 files are already available ?

quaelnix:
> Is there a "next" branch where RD5 files are already available ?

Afaik not.

xuiqzy:
Hi, what is needed for these lookup updates to be merged and to be available in brouter web and other brouter based use cases?
(Mainly interested in the cycleway:both, cycleway:lane and related updates)

afischerdev:
To all friends of profiles:

Please keep in mind for testing your profiles against a new/changed `lookups.dat` call
```
java -cp  brouter-1.7.3-all.jar  btools.expressions.IntegrityCheckProfile profiles2/lookups.dat profiles2_my
```

devemux86:
@afischerdev 

Any news on this pull request?

afischerdev:
Unfortunately not.
At the beginning of this request there were some concerns about the hard change of the lookup version number.
I can't say if they are all gone. 

When we are ready for update:
All the actions of using the new lookups fro generation of rd5 files is part of the maintaining. The Server is running since installation in last summer with the same configuration. There is always high traffic and hard to detect a date for a bigger update session. As you see all this needs an update - not only a new BRouter library and a new lookups file.

And it would be a good idea to update the BRouter web for this issue #710. So we prepare a nightly docker build for  the server build but it could not be used with the BRouter web docker file. And it would be easier to install direct from github on an server update - I guess.

afischerdev:
I've changed the weighting of the entries. And did a test against the old lookup for Germany export.
Therefore I used the new method for pseudo-tags #771.
And then we have smaller file sizes as before.

```
old: 11 Datei(en),    319.061.394 Bytes
new: 11 Datei(en),    312.655.222 Bytes
```

misc/profiles2/lookups.dat
ProBackup-nl, line 1: Nice, to include 
moped;0000000001 use_sidepath
and
mofa;0000000001 use_sidepath

afischerdev:
Added `speed_pedelec` as discussed in #790

misc/profiles2/lookups.dat
```diff
@@ -69,123 +59,245 @@ surface;0000005778 mud
...
+
+estimated_traffic_class;0000000001 1
+estimated_traffic_class;0000000001 2
+estimated_traffic_class;0000000001 3
+estimated_traffic_class;0000000001 4
+estimated_traffic_class;0000000001 5
+estimated_traffic_class;0000000001 6
+estimated_traffic_class;0000000001 7
 
 lit;0000809223 yes
```
mjaschen, line 150: I'd like to add `lit=no` which would allow to avoid ways that are known to unlit.

Background: I'm developing a pedestrian profile currently and would like to provide an option "Avoid unlit ways".

ProBackup-nl: Great that `highway;0000000001 elevator` in the way context is finally here. That will improve routing through the Kennedyfietstunnel in Antwerp.

ProBackup-nl (approved):
Great to have highway=elevator here.

afischerdev:
Added T1-hiking again to the lookups.dat. It is not used anymore but the hiking-mountain profile contains it. 
That is not the problem - we will change the inbuilt profile -, but I see childs of this profile driven by user ideas will break.

So I followed @polyscias 
> I think breaking things for (many) users is bad

And a check of the recently added tags/values ​​was also performed.
With the last small resort we have now a total for the same Germany export with 250K bytes less.

afischerdev:
Please do not forget this test for your profiles

```
java -cp brouter.jar  btools.expressions.IntegrityCheckProfile profiles2/lookups_my.dat profiles2
```

afischerdev:
The values `yes` and `designated` already have seperate entries, so it is not necessary to merge them.
Size check old and new shows equal sizes.

ProBackup-nl:
@afischerdev I don't like the alias for the `traffic_calming` `choked_table` in lookups.dat. A choked table is not a synonym for a `table`. In real world the choked table is the extra annoying combination of a `choker` **and** a `table`.  With a normal choker without traffic, you have the luck that you can pass at full speed. With the `choked_table` without other traffic, you still need to reduce speed because of the vertical deviation. In a good profile the `choked_table` needs to have a penalty of approximately the `table` plus the `choker`.

afischerdev:
@ProBackup-nl 
Thanks for the additional info.
At the moment we have this nodes (new lookup with counts for Germany, alias included)
```
traffic_calming;0000011529 island
traffic_calming;0000008972 choker
traffic_calming;0000007656 bump
traffic_calming;0000007408 chicane
traffic_calming;0000006360 table
traffic_calming;0000004850 hump
traffic_calming;0000001841 cushion
traffic_calming;0000000543 rumble_strip
traffic_calming;0000000368 dip
traffic_calming;0000000273 yes
```

Looks a little bit like the traffic_calming icons.
But we are interested in routing data not in mapping data. We need to know that there comes a speed breaker and may be it is a hard or more a soft format or only a narrowed road. 
I think we don't need all this types.

You are very interested and committed to this topic. My be you like to do more investigation how to reduce this types and classify the types by penalty.
And I noticed there are lots of way with this tag. We have to collect them as well.

xuiqzy:
@afischerdev In case some of the `traffic_calming` values are to be combined, I wanted to note that some differences like cushion vs hump might not be important for cars but are important for bicycles and motorcycles (can pass on the side) as well as buses/emergency vehicles (pass over it without a slowdown). So it would be great to keep those differences available for routing profiles if possible.
If these considerations are already clear, sorry for the noise!

ProBackup-nl:
> We need to know that there comes a speed breaker and may be it is a hard or more a soft format or only a narrowed road. I think we don't need all this types.
> 
> May be you like to do more investigation how to reduce this types and classify the types by penalty. And I noticed there are lots of way with this tag. We have to collect them as well.

What we need to know, depends on the type of transportation: by foot, horse, boat, train or MTB the traffic calming relevance is absent, on a slow bicycle a little near absent. More relevant is traffic calming when using a velomobile or moped, and those are different compared to a car or wide truck.

The iD editor has 12 tags translated into Dutch, of which 8 tags contain an icon. Amazing is that from the brouter profiles2 only gravel and quaelnix-gravel use traffic_calming=* (one penalty for any value). The frequency of travel_calming in NL and riding a velomobile without suspension lead me to improving the vm-touristic profile.

For me personally the traffic_calming order of importance:
1. is it a `bump` or `double_dip`? No Go -> possible velomobile damage
2. because `double_dip`s are sometimes mistagged as 2 seperate `dip`s, I will avoid the `dip` too.
3. is it full street width or can I pass by on the main road, without a change of surface? F.e. `cushion`s can often be passed by when riding on a motor bike or velomobile.
4. is the vertical deviation partly made out of preformed, often concrete, element(s), especially the ramps? Yes, then trapezoid type = horrible; otherwise its more likely a sinusoidal type, which is better acceptable.
5. lateral deviation is important too, `chicane` != `choker`
6. both lateral and vertical deviation `choked_table` is a different type.

afischerdev:
@xuiqzy @ProBackup-nl 
Agree with all, but we are talking not about profiles. 
Here the context is lookups.dat And the entries must give the possibility for the profile to use it in every case that is needed.
So what is your suggestion for an entry list?

traffic_calming;0000000000 main_value list of alias

ProBackup-nl:
For `node:traffic_calming` (without much thought about aliases):

1. yes
2. bump
3. chicane
4. choked_table
5. choker
6. dip
7. double_dip
8. table
9. rumble_strip

no|island|rumble_strip|painted_table

For `node:surface`:
1. `asphalt`
2. `paving_stones`|`paved`| `sett`| `cobblestone`|anything with a textured surface (not `concrete`)
3. all others

For bicycle profiles it's handy to know if there is a `bicycle_bypass=yes`, that would reduce the penalty. Tagged 324 times in a full NL bbox, 645 times in Europe.

ProBackup-nl:
@afischerdev I updated the comment above, with traffic_calming number 9 `rumble_strip`, and surface number 2 `paving_stones`. The reason is why I don't prefer to cycle to some villages cognitively without active memory why. After looking back the images I saw the `good` `asphalt` was cut open at 12 different spots within 500 meter and filled with `paving_stones`. Each strip individually not enough to mark, 5-10 meters, some with a slight vertical deviation, not enough for a hump. All 12 tiny road surface differences combined in 45 seconds become one big annoying new traffic calming thing.

dadebue:
Hi, noob question here: I want to have the via_ferrata tags information. I can't just go ahead and use the new lookups with the current rd5 files right? From the developer guide:

> The lookup-table is allowed to grow over time, to include more tags
and values as needed. To support that evolution, it carries a major
and a minor version number. These numbers are also encoded into
the routing data files, taken from the lookups.dat that is used
to pre-process the routing data files.

RuHitt:
Hello everyone,
When can we expect the new 'lookups.dat' file to be included in a release? This would be really helpful for using 'new' tags in routing files. This would be especially useful for evaluating 'cycleway:both=lane', as this tag is now used exclusively for 'cycle paths'. You could argue that these are not actually cycle paths, but nevertheless, my current profile cannot route through these roads appropriately.
Is there any way I can help speed up this process?

Kind regards, Ruben

afischerdev:
Was sdone with #829

PR #464, opened 2022-09-29
Enable use of direction to app

afischerdev:
Enable use of direction to app #314


devemux86:
So with this fix, we will be able to use the "direction" parameter with the Android service interface,
like the "heading" parameter in the website: changing the 1st turn instruction instead of the route?

Android and website will have different parameters: "direction" and "heading"?
As mentioned in the other issue, the "direction" does not work on the website:
https://github.com/abrensch/brouter/issues/314#issuecomment-1260981720

afischerdev:
> Android and website will have different parameters: "direction" and "heading"?
As mentioned in the other issue, the "direction" does not work on the website:

Yes, that is still the way it goes.
Because I have no different result over the brouter web client. Not sure if there is an extra handling for heading.
Your sample wasn't working for me. So I made an own [sample](http://brouter.de/brouter-web/#map=16/49.9507/8.7022/standard&lonlats=8.703382,49.951496;8.692053,49.947948&heading=0) that doesn't work on the website but offline with lib 1.6.3 and 1.6.4 it worked like expected.  

It can also be tested via direct call:
[heading=0](https://brouter.de/brouter?lonlats=8.703382,49.951496|8.692053,49.947948&profile=trekking&alternativeidx=0&format=gpx&timode=3&heading=0)
[heading=180](https://brouter.de/brouter?lonlats=8.703382,49.951496|8.692053,49.947948&profile=trekking&alternativeidx=0&format=gpx&timode=3&heading=180)

brouter-routing-app/src/main/aidl/btools/routingapp/IBRouterService.aidl
```diff
@@ -1,38 +1,39 @@
...
+    //  "nogos"           = lon,lat,radius|... (optional, radius in meters)
+    //  "polylines"       = lon,lat,lon,lat,...,weight|... (unlimited list of lon,lat and weight (optional), lists separated by |)
+    //  "polygons"        = lon,lat,lon,lat,...,weight|... (unlimited list of lon,lat and weight (optional), lists separated by |)
+    //  "profile"         = profile file name without .brf 
+    //  "alternativeidx"  = [0|1|2|3] (optional, default 0)
+    //  "exportWaypoints" = 1 to export them (optional, default is no export)
+    //  "pois"            = lon,lat,name|... (optional)
+    //  "extraParams"     = Bundle key=value list for a profile setup (like "profile:")
+    //  "timode"          = turnInstructionMode [0=none, 1=auto-choose, 2=locus-style, 3=osmand-style, 4=comment-style, 5=gpsies-style, 6=orux-style] default 0
+    //  "direction"       = angle (optional to give a route a start direction)
```
devemux86, line 32: Isn't the 1st section about the Android parameters, so we should include the "direction"?
And the 2nd section about the website parameters,  so we should include the "heading"?

devemux86:
> Your sample wasn't working for me.

If we remove the "format=gpx", my last samples can be seen in the browser (same route, different 1st instruction):
[heading=0](https://brouter.de/brouter?lonlats=13.388700485229478,52.517017364501946%7C13.390865999999981,52.516046&heading=0&profile=car-fast&alternativeidx=0&timode=3) (u-turn)
[heading=180](https://brouter.de/brouter?lonlats=13.388700485229478,52.517017364501946%7C13.390865999999981,52.516046&heading=180&profile=car-fast&alternativeidx=0&timode=3) (straight)

afischerdev:
> Isn't the 1st section about the Android parameters, so we should include the "direction"?
And the 2nd section about the website parameters, so we should include the "heading"?

No, the aidl is the discription of the whole interface to the app. It has two section to make clear old / new interface.
But to make it equal I will replace direchtion with heading - cause this is already working and there is no app that uses direction - it is not working without this PR. 

> If we remove the "format=gpx", my last samples can be seen in the browser (same route, different 1st instruction):

Yes, I know. but it isn't a funny use case go the same way with different instructions. Anyway it shows 'heading' is recognized by the server.
My use case for this feature is: design a route, start route in app, on a right turn  go left (or what ever), app should notice it and reroute with the new heading. And then I don't want to hear a u-turn.

afischerdev:
@zod 
What do you think should we add this first or should I add this to the first peace of 1.6.4 updates. I plan to start with the new interface variables. That contains also aidl and BRouterWorker.

zod:
According to `git blame` the `direction` parameter is used by locus (fdf06cbdd00bcb1d7f423dc40d218f06f6660970). I think we should deprecate `direction` but still support it.

zod (changes requested):
According to `git blame` the `direction` parameter is used by locus (fdf06cbdd00bcb1d7f423dc40d218f06f6660970). I think we should deprecate `direction` but still support it.

afischerdev:
@zod
Yes, you are right.
But when I changed this I found that it needs a
`rc.forceUseStartDirection = true;`
otherwise it is only used on recalc and not on start
[please see ](https://github.com/abrensch/brouter/blob/51dc7fb1fbc310796fba8b76e3c92b1b3b935891/brouter-core/src/main/java/btools/router/RoutingEngine.java#L615)
I don't know what was the intention in the old days to add this and if it is still used. But in my eyes I would like to get a route  that starts on North when I define a direction/heading North.
But when there is a deeper sense we could hold both in the app.

abrensch:
>> I don't know what was the intention in the old days
>> to add this and if it is still used. But in my eyes I would
>> like to get a route that starts on North when I define a
>> direction/heading North. But when there is a deeper
>> sense we could hold both in the app.

there's only "direction" in the api, not the speed,  but direction is undefined at low speed.

So it's a hack to restrict the usage of "direction" for recalculations, where this is most needed (to put a bias on the forward solution). Correct solution would be a cutoff on speed.

afischerdev:
@abrensch 
Ok, good to know. 
With the new change this should work again - Locus still sends direction.
And when using 'heading' it works from the beginning - only tested on lib 1.6.4.

PR #471, opened 2022-10-21
Remove cycleroute always having access granted

Totorrr:
Some bicycle routes can have portions which are not accessible.

Potential reasons for this:
 - temporary construction work on the ways (sometimes for months),
 - local access change without simultaneous update of bicycle route relation
   (lack of mapper's knowledge or lack of time to redefine an updated route),  
 - ... 

This branch changes fastbike & trekking profiles so that being on a cycle 
route does not automatically mean "access granted" for bicycles. The normal 
access checkings are used instead.

However, the accesspenalty is set to 15 if the no-access way belongs to such a 
route, so that the way is not completely forbidden if there is no other 
alternative unavailable.

### DIFF
--- a/misc/profiles2/fastbike.brf
+++ b/misc/profiles2/fastbike.brf
@@ -87,14 +87,13 @@ assign defaultaccess
 # calculate logical bike access
 #
 assign bikeaccess =
-       or any_cycleroute
-          switch bicycle=
-                 switch bicycle_road=yes
-                    true
-                    switch vehicle=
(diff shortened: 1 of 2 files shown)


zod:
Do you have an example where this behavior would help?

afischerdev:
I haven't heard of @zod for a while. If you don't mind I'll merge the PR, I like the idea to block this ways not totally.
But what do you think of a change also on the other profiles?

PR #472, opened 2022-10-21
Correc indent typo

Totorrr:
Minor 1 char change to improve profile code readability.


zod:
I think the indention is correct and other profiles have the same indention.

Totorrr:
@zod The `10.0` belongs to the switch line 207, not to the series of switch in between). This is why I suggest aligning it to line 207.

PR #473, opened 2022-10-23
Move testcases from brouter-server

zod:
The tests within brouter-server module don't actually test brouter-server but some other modules.


afischerdev:
Great idea

PR #474, opened 2022-10-28
Tests for RouteServer

zod:
After moving the tests in #473 `RouteServer` was left without any tests. Unfortunately it's not possible to write good unit tests because it allocates the `RoutingEngine` and all other dependencies by itself.


zod:
While implementing the tests I've noticed that the referrer handling (added in cdac1cc5aaa97d3bda16c9f2682f74560ab9b244) doesn't work because it always checks for an uppercase `Referer` but the input is converted to lowercase. @abrensch is this code required? Should it be fixed or removed?

abrensch:
>>  is this [Referer] code required? Should it be fixed or removed?

you're right, disfunctional, can be removed

It was some troubleshouting with robot activity

afischerdev:
@zod 
I got the tests run when enter this
`httpConnection.setConnectTimeout(1000);`

zod:
Increasing the timeout didn't work for the action runner. If the server isn't running yet it doesn't run into timeout but the connection is refused immediately.

afischerdev:
> Increasing the timeout didn't work for the action runner. If the server isn't running yet it doesn't run into timeout but the connection is refused immediately.

May be, but still: with 'httpConnection.setConnectTimeout(100);' tests do not run on my equipment. I need 1000 as parameter.

zod:
Then you should upgrade your hardware ;)

Feel free to fix it or wait for me to do it next week.

afischerdev:
> Then you should upgrade your hardware ;)

Yes, I think so. Do I have a sponsor?

> Feel free to fix it or wait for me to do it next week.

I'll wait.

PR #481, opened 2022-11-08
Add test profiles with lookups.dat

afischerdev:
As discussed in #458 I add a test for profiles against the current lookup.dat

And a command line test version as well

### DIFF
--- a/brouter-expressions/src/main/java/btools/expressions/IntegrityCheckProfile.java
+++ b/brouter-expressions/src/main/java/btools/expressions/IntegrityCheckProfile.java
@@ -11,11 +11,11 @@ public class IntegrityCheckProfile {
       return;
     }
 
-	  IntegrityCheckProfile test = new IntegrityCheckProfile();
+    IntegrityCheckProfile test = new IntegrityCheckProfile();
     try {
       File lookupFile = new File( args[0] );
       File profileDir = new File( args[1] );
-	    test.integrityTestProfiles(lookupFile, profileDir);
+      test.integrityTestProfiles(lookupFile, profileDir);
     } catch (Exception e) {
       System.err.println(e.getMessage());
     }
--- a/brouter-expressions/src/main/java/btools/expressions/IntegrityCheckProfile.java
+++ b/brouter-expressions/src/main/java/btools/expressions/IntegrityCheckProfile.java
@@ -6,15 +6,15 @@ import java.io.IOException;
 public class IntegrityCheckProfile {
 
   public static void main( final java.lang.String[] args ) {
-    if ( args.length != 2 ) {
-      System.out.println( "usage: java IntegrityCheckProfile <lookup-file> <profile-folder>" );
+    if (args.length != 2) {
+      System.out.println("usage: java IntegrityCheckProfile <lookup-file> <profile-folder>");
       return;
     }
 
     IntegrityCheckProfile test = new IntegrityCheckProfile();
     try {
-      File lookupFile = new File( args[0] );
-      File profileDir = new File( args[1] );
+      File lookupFile = new File(args[0]);
+      File profileDir = new File(args[1]);
       test.integrityTestProfiles(lookupFile, profileDir);
     } catch (Exception e) {
       System.err.println(e.getMessage());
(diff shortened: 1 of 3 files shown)


brouter-expressions/src/main/java/btools/expressions/IntegrityCheckProfile.java
```diff
@@ -0,0 +1,50 @@
+package btools.expressions;
```
zod, line 1: Why is it necessary to add a commandline util? Which tests can't be handled using normal tests?

brouter-expressions/src/test/java/btools/expressions/IntegrityCheckProfileTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+    assertNotNull("Missing profiles", files);
+
+    for (File f : files) {
+      if (f.getName().endsWith(".brf")) {
+        BExpressionMetaData meta = new BExpressionMetaData();
+        BExpressionContext expctxWay = new BExpressionContextWay( meta );
+        BExpressionContext expctxNode = new BExpressionContextNode( meta );
+        meta.readMetaData( new File( profileDir, "lookups.dat") );
+        expctxNode.setForeignContext(expctxWay);
```
zod, line 27: Mixed style: Sometimes spaces after parenthesis and sometimes not. After the large reformatting we currently have a clean style within the repository and I'd like to keep it that way.

brouter-expressions/src/test/java/btools/expressions/IntegrityCheckProfileTest.java
```diff
@@ -0,0 +1,34 @@
+package btools.expressions;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
```
zod, line 4: Unused import

zod (changes requested):

brouter-expressions/src/main/java/btools/expressions/IntegrityCheckProfile.java
afischerdev: The command line tool is for real life.
JUnit works at the end of the development process.

brouter-expressions/src/test/java/btools/expressions/IntegrityCheckProfileTest.java
devemux86: > Is there any automatic for that?

If mean "Reformat code" and use IDEA, there are several shortcuts:
https://www.jetbrains.com/help/idea/reformat-and-rearrange-code.html

afischerdev: @devemux86 
Yes, that's the way it goes. But not very handy when you work with Notepad++.  So with automatic I mean 'reformat on save' or 'reformat on copy to git' or 'reformat an build with gradle' or whatever.

zod: AndroidStudio (which ist based on IntelliJ), VSCode and I guess some other IDEs offer formatting on save. If once setup you don't have to Care about formatting.

There ist no gradle task provided by checkstyle. You could add a git hook which checks formatting.

brouter-expressions/src/main/java/btools/expressions/IntegrityCheckProfile.java
zod: I think our workflows are quite different. When using Test-driven development tests are written before the implementation and help while developing. Is the command line tool used after implementation is complete?

afischerdev: Our views are different to testing.
my: 5+5 = 10
your: 10 = 5+5
Both have the result 10. But what is when you start with the result 
10 and get the way 6+4. Result is ok but the way could be wrong.
And what is when the best way is 5+6. You will never find it.
Other sample:
49 Euro ticket, the result is clear but the ways to get there? Or is a 1 Euro per day ticket a better solution? (Sorry for the German sample)
More brouter sample:
We had the PR #474 with test server - pretty good idea to test that way.
Let's take a look at the 'overrideParameter' routine. The result is 1902 for track length. Why that, why not 42 like every on question? You have previously routed to get this value. If we work on the new library now - some day - this test will break. Because of different length. But the intent is still perfect: the override parameter still works. Only the test is wrong.

My workflow is to test every line I write - on the perfect way. But this means for me:
- 1. line has its result in a log file
- 2. line in gpx file
- 3. line needs a diff on two gpx files
- 4. line is only visible on a graphical output test
So I think this is also test driven.
In a pure form and not in the restriction of junit.

zod: I honestly don't understand your references, but I think there is enough information regarding test driven development available.

I actually think that my tests in #474 are bad tests because they check the route length, but that's the only thing that can be observed. Because of the architecture and missing dependcy injection it's impossible to write real unit tests instead of system tests.

PR #495, opened 2023-01-10
Fix regression of travel time calculation

quaelnix:
The combination of bd02587 and 57da34d causes a bug in the travel time calculation in the kinematic model, which can lead to a significant miscalculation in certain cases:
| Old | New |
| :---: | :---: |
| old | new |

Example: https://brouter.de/brouter-web/#map=12/51.8281/10.7136/standard&lonlats=10.789197,51.839793;10.636347,51.816129;10.618328,51.799937&profile=fastbike
___
This pull request fixes the bug by removing the [seemingly undocumented behavior](https://brouter.de/brouter/profile_developers_guide.txt) of falling back into hiking mode (at a walking speed equal to the maximum cycling speed: `rc.maxSpeed`) on way segments with a `costfactor > 4.9`.

### DIFF
--- a/brouter-core/src/main/java/btools/router/StdPath.java
+++ b/brouter-core/src/main/java/btools/router/StdPath.java
@@ -202,26 +202,20 @@ final class StdPath extends OsmPath {
     elevation_buffer += delta_h;
     double incline = calcIncline(dist);
 
-    double wayMaxspeed;
-
-    wayMaxspeed = rc.expctxWay.getMaxspeed() / 3.6f;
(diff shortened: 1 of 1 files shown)


afischerdev:
You are right, I was focused on the words `falling back into hiking mode` and so I got

`6 km/h * exp(-3.5 * (0.10 + 0.05)) ~= 3.6 km/h
`

but it is still a biking profile and has the higher speed. Foot mode had a fixed value of 6 in the past.

PR #498, opened 2023-01-16
Update lib part four - recalculation at end of track

afischerdev:
The calculation for voice hints and speed profile is moved to the end of track generation. 
This only effects tracks with more then two points.
And a recalculation is done for elevation data and track distances.

The output at the moment is not fully correct - the routines for voice hints will follow.
The main discussion is found in #441.

### DIFF
--- a/brouter-core/src/main/java/btools/router/OsmPathElement.java
+++ b/brouter-core/src/main/java/btools/router/OsmPathElement.java
@@ -36,6 +36,10 @@ public class OsmPathElement implements OsmPos {
     return selev;
   }
 
+  public final void setSElev(short s) {
+    selev = s;
+  }
+
   public final double getElev() {
     return selev / 4.;
   }
@@ -60,6 +64,12 @@ public class OsmPathElement implements OsmPos {
     }
   }
 
+  public final void setAngle(float e) {
+    if (message != null) {
+      message.turnangle = e;
+    }
+  }
+
   public final long getIdFromPos() {
     return ((long) ilon) << 32 | ilat;
   }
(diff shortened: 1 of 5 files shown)


quaelnix:
I can tell you from experience that if you continue to update the library in large and messy chunks, you are opening Pandora's box. The old rule "untested code is broken code" almost always applies.

This pull request broke at least two things:
1. The ascent no longer increases monotonically with distance*
2. The travel time computation can be off by at least up to 3 %

The bugs might not be related with each other, since the calculated travel time decreases even though the total ascent increases. While the total distance does not change.

*) If route A is a subset of route B then the total ascent of route B must be greater than or equal to the total ascent of route A. Example:
| | [Route A](http://brouter.de/brouter-web/#map=15/50.2392/9.0400/standard,route-quality&lonlats=9.034353,50.245787;9.037089,50.243861&profile=shortest) | [Route B](http://brouter.de/brouter-web/#map=15/50.2392/9.0400/standard,route-quality&lonlats=9.034353,50.245787;9.040507,50.239814&profile=shortest) |
| :--- |:---: | :---: |
| Total ascent with 2387513 | 25 m |  15 m |
| Total ascent with d081e5e |  25 m |  25 m |

quaelnix:
The first bug is caused by now passing the deltas through the elevation filter: https://github.com/abrensch/brouter/blob/23d55aa40c5e384f75daca9db3f6cb5bf92f0177/brouter-core/src/main/java/btools/router/RoutingEngine.java#L775-L780
in reverse order.
___
The old (correct) behavior can be restored by iterating through the for loop in reverse:
```
diff --git a/brouter-core/src/main/java/btools/router/RoutingEngine.java b/brouter-core/src/main/java/btools/router/RoutingEngine.java
index d69a11d..05fc9b3 100644
--- a/brouter-core/src/main/java/btools/router/RoutingEngine.java
+++ b/brouter-core/src/main/java/btools/router/RoutingEngine.java
@@ -713,7 +713,7 @@ public class RoutingEngine extends Thread {
     short ele_end = Short.MIN_VALUE;
     double eleFactor = routingContext.inverseRouting ? 0.25 : -0.25;
 
-    for (i = 0; i < ourSize; i++) {
+    for (i = ourSize - 1; i > 0; i--) {
       OsmPathElement n = t.nodes.get(i);
       if (n.message == null) n.message = new MessageData();
       OsmPathElement nLast = null;
```

PR #499, opened 2023-01-19
Fix rounding error in calcDistance

quaelnix:
The way BRouter rounds floating-point numbers to integers in the calcDistance function introduces a round-off error accumulation that increases as the average distance between two nodes decreases. This bug caught me completely off guard when I first started working with BRouter.

Here is a comparison of the total distance error distributions of 100k random routes before and after fixing the bug:
brouter_rounding_error

Nürburgring Sprint course:
- https://www.openstreetmap.org/way/27852990:
- sprintcourse
- [https://brouter.de/brouter-web/#map=16/50.3330/6.9438/standard...&profile=shortest](https://brouter.de/brouter-web/#map=16/50.3330/6.9438/standard&lonlats=6.938351,50.329227;6.946246,50.337144;6.946105,50.334522;6.939788,50.333517;6.938351,50.329227&profile=shortest)
- | BRouter: `(int) d + 1` | GPX track length | BRouter: `max(1, round(d))` |
  | ---: | ---: | ---: |
  | 3690 m | 3615 m | 3614 m |

### DIFF
--- a/brouter-core/src/main/java/btools/router/OsmPathElement.java
+++ b/brouter-core/src/main/java/btools/router/OsmPathElement.java
@@ -75,7 +75,7 @@ public class OsmPathElement implements OsmPos {
   }
 
   public final int calcDistance(OsmPos p) {
-    return (int) (CheapRuler.distance(ilon, ilat, p.getILon(), p.getILat()) + 1.0);
+    return (int) Math.max(1.0, Math.round(CheapRuler.distance(ilon, ilat, p.getILon(), p.getILat())));
   }
 
   public OsmPathElement origin;
--- a/brouter-core/src/main/java/btools/router/RoutingContext.java
+++ b/brouter-core/src/main/java/btools/router/RoutingContext.java
@@ -462,7 +462,7 @@ public final class RoutingContext {
         }
       }
     }
-    return (int) (d + 0.5);
+    return (int) Math.max(1.0, Math.round(d));
   }
 
   public OsmPathModel pm;
--- a/brouter-mapaccess/src/main/java/btools/mapaccess/OsmNode.java
+++ b/brouter-mapaccess/src/main/java/btools/mapaccess/OsmNode.java
@@ -100,7 +100,7 @@ public class OsmNode extends OsmLink implements OsmPos {
   }
 
   public final int calcDistance(OsmPos p) {
-    return (int) (CheapRuler.distance(ilon, ilat, p.getILon(), p.getILat()) + 1.0);
+    return (int) Math.max(1.0, Math.round(CheapRuler.distance(ilon, ilat, p.getILon(), p.getILat())));
   }
 
   public String toString() {


polyscias (commented):
I do not know why the original code was using +1 but it looks strange to me, is that because the distance me never be 0? I briefly checked the code but I do not think so.

Would be good to get clarity on this first.

afischerdev:
This problem is already addressed by the lib 1.6.4, but this is not be finalized in this repro at the moment - please see #441.

@polyscias 
> I do not know why the original code was using +1 but it looks strange to me, is that because the distance me never be 0? I briefly checked the code but I do not think so.

Yes that was the reason I think. 
Lib 1.6.4 uses a `(int) (d + 0.5) `and a simple test` if (d == 0) d = 1; ` 
That brings a result of `3617m` for that round. GPX length can only test against its own points and my favorite Win RouteConverter shows the same length with the same file.

Anyway, you have placed your checks in several distance calculations and I will control lib 1.6.4 against that. But it will take a while.

quaelnix:
> That brings a result of `3617m` for that round.

This pull request is not about the length of the GPX track. It is about the length BRouter calculates for each segment and for the final `track-length`.

> Lib 1.6.4 uses a `(int) (d + 0.5)`

That change was apparently sneaked in by a commit called: [add vars for misplaced pts](https://github.com/abrensch/brouter/commit/d2297c0c520b2fb3bb5cdd94342b403fb96d7c1d#diff-cb9c686b2ddc60fdcd577610cae5c73e4edff98d0c52d9333bf2cf7b460892afR465). 

And this only has an effect on the calculated `track-length` because [the pull request you merged a few minutes ago](https://github.com/abrensch/brouter/pull/498) calls `calcDistance` of the RoutingContext class in places where the `calcDistanc` method of the OsmPathElement class used to be called.

This affects the state of the `nogoCost` and has the potential to break a lot.

> and a simple testif `(d == 0) d = 1;`

Where?

afischerdev:
The work for the transfer of lib 1.6.4 tis not ready - please see #441.

> Where?

please see [here](https://github.com/afischerdev/brouter/blob/5d9fdf9afdaf49b7033bbc75a169310cc31678d2/brouter-core/src/main/java/btools/router/OsmPath.java#L333)

quaelnix:
In my honest opinion, the code is hard enough to read as it is, and we should do everything we can to not make things more obscure. I would strongly disagree with separating the sanity check (making sure the distance is always positive) from the rounding itself.

Changing the rounding strategy is also not something that should fall out of the sky in the middle of nowhere, as it also affects other projects (see: https://github.com/nrenner/brouter-web/blob/025eb02a6a0f60bcd826235b09ffca722bc23bd6/js/util/CheapRuler.js#L118-L125) and most likely even things we're not even thinking about right now.

We should also not use different rounding strategies in methods with the same name.

quaelnix:
> For me a value of zero is valid for a distance calculation.

Yes you are right and if I remember correctly abrensch also mentioned in one of the issues that the calcDistance stuff needs cleanup.

> and place the check in the calling function - as OsmNode and OsmPathElement do.

The `calcDistance` helpers in OsmNode, OsmPathElement and RoutingContext are all meant to "sanitize" the raw distance calculated by CheapRuler. And the RoutingContext `calcDistance` method is not free of side effects and so probably should not be called in a different place than it used to be called, or things could break.

PR #500, opened 2023-01-24
Fix regression in elevation filter logic

quaelnix:
25e506d changed the order in which the elevation deltas are passed through the elevation filter, which can lead to the undesirable behavior that appending segments to the end of a route can decrease the calculated total ascent. This fixes the bug by adjusting the elevation filter accordingly.

### DIFF
--- a/brouter-core/src/main/java/btools/router/RoutingEngine.java
+++ b/brouter-core/src/main/java/btools/router/RoutingEngine.java
@@ -772,16 +772,15 @@ public class RoutingEngine extends Thread {
         if (ele_last != Short.MIN_VALUE) {
           ehb = ehb + (ele_last - ele) * eleFactor;
         }
-        if (ehb > 10.) {
-          ascend += ehb - 10.;
-          ehb = 10.;
-        } else if (ehb < 0.) {
-          ehb = 0.;
+        if (ehb > 0) {
+          ascend += ehb;
+          ehb = 0;
+        } else if (ehb < -10) {
+          ehb = -10;
         }
       }
 
     }
-    ascend += ehb;
 
     t.ascend = (int) ascend;
     t.plainAscend = (int) ((ele_start - ele_end) * eleFactor + 0.5);


afischerdev:
Good pointer, rework is needed.

But:
You already placed some samples in the PR #498 and your suggestion for this Route B is working will. 
On other samples as your Ring or my Dreieich I found some differences in result.

|| expected | delivered |
| :--- |:---: | :---: |
[Route B](http://brouter.de/brouter-web/#map=15/50.2431/9.0386/standard,route-quality&lonlats=9.034353,50.245787;9.040507,50.239814&profile=shortest) | 25m | 25m |
[The Ring](https://brouter.de/brouter-web/#map=14/50.3357/6.9296/standard&lonlats=6.938351,50.329227;6.946246,50.337144;6.946105,50.334522;6.939788,50.333517;6.938351,50.329227&profile=shortest) | 52m | 29m |
[Dreieich sample](https://brouter.de/brouter-web/#map=13/49.9869/8.6874/standard&lonlats=8.70014,49.998637;8.715227,49.996196) | 15m | 10m |

I use a more painless version now

```
      {
        ...
        ehb = 0.;
        if (ele_last != Short.MIN_VALUE) {
           ehb = ( ele_last - ele ) * eleFactor;
        }
        if (ehb < 0.) {
          ehb = 0.;
        }
        ascend += ehb;
    }
```

What do you think?

quaelnix:
> What do you think?

That we should not change the method of calculating the total ascent unless it is a significant improvement. Changing it for no good reason would upset many people who have become accustomed to the previous operating principle for many years.

The idea behind the current code seems to be to reject noise from the underlying elevation data and to account for the fact that the kinetic energy that builds up when riding downhill reduces the effort on subsequent climbs up to a certain point.

I agree that the code is far from perfect, but improving it is definitly not a trivial task.

quaelnix:
Any reasonable device (Garmin Edge, ...) or tool (Strava, Komoot, ...) that calculates total ascent from raw elevation data prepares the data in one way or another before using it. Here is an excerpt from the [Strava documentation](https://support.strava.com/hc/en-us/articles/216919447-Elevation-for-Your-Activity#knownissues) for reference:

> We smooth the data before calculating gain and depending on the resulting data, elevation changes may not be enough to pass a threshold that we use for determining whether or not you have gained elevation.

afischerdev:
Yes, I would do that as well. But we are at the end of our work.
And alright, the wording is 'filtered ascend' that should be enough to understand this value in the right way and accept some differences.

PR #507, opened 2023-02-13
Add ascend test

afischerdev:
Add a test for ascend calculation. 
But this is not valid at the moment cause we don't have srmt data in the test files.


afischerdev:
I moved back to commit 33b5da0. It was a mistake to add the other commits here. 
There are plans to add some test SRTM data here later.

afischerdev:
Having worked with SRTM, I currently see no way of generating 'small' SRTM data for testing.
Feel free to open again when you have an idea on that.

zod:
Manually creating a tile with lots of NODATA fields and only a small part with real elevation data encoded using `RasterCoder` as `.bef` should create a rather small file

PR #510, opened 2023-02-26
Update Lib Part Six - Change Export

afischerdev:
The new voice hint rules need an update for the export.

The main changes in voice hint was
- add 180 degree u-turn
- add beeline routing between two points

This cannot be used for all outputs due to unknown commands in the clients. So at the moment we have the following changes:

- Locus GPX output changes from wpt to trkpt route points, using only standard GPX tags
- OsmAnd, Comment Style, Gpsies, Orux work as before
- add old Locus output to keep it for other users/apps (has a 180 degree u-turn but no beeline information)
- there are two hidden turn instruction modes:
  - Cruiser export (8) is like OsmAnd but has the new rules (u-turn and straight line)
  - BRouter internal (9) with routing hints at trkpt tags (contains information only found in JSON export)
  
The Cruiser output is temporary and will be removed when OsmAnd supports 180-u-turns and beelines. It can be used to test against new rules.
BRouter internal is experimental for control. So please don't use it for anything else.

@vodie
You did initiate the Orux output in the past, May be you have more info about u-turn and beeline to get the output improved.

@vshcherb 
For OsmAnd the new voice hint constants are in [VoiceHint](https://github.com/abrensch/brouter/blob/b735cd3e4ea1d4f25b37bdfa2c1384ad24cba69b/brouter-core/src/main/java/btools/router/VoiceHint.java#L13)
It can be tested against the Cruiser output (8). The publishing of the features will be done some day.

The main discussion for the 1.6.4 lib is in #441.

### DIFF
--- a/brouter-core/src/main/java/btools/router/OsmTrack.java
+++ b/brouter-core/src/main/java/btools/router/OsmTrack.java
@@ -21,9 +21,13 @@ import java.io.InputStreamReader;
 import java.io.StringWriter;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
+import java.util.TimeZone;
 
 import btools.mapaccess.MatchedWaypoint;
 import btools.mapaccess.OsmPos;
@@ -45,6 +49,9 @@ public final class OsmTrack {
 
   public boolean showspeed;
   public boolean showSpeedProfile;
+  public boolean showTime;
+
+  public Map<String, String> params;
 
   public List<OsmNodeNamed> pois = new ArrayList<OsmNodeNamed>();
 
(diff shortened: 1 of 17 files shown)


devemux86:
> BRouter internal is experimental for control. So please don't use it for anything else.

Isn't the timode=7 a new mode with turn instructions and other info at trkpt tags?

Cruiser uses it when BRouter 1.6.4 (Beta) is installed, with beeline, etc.

devemux86:
@afischerdev 

Isn't the timode=7 a new mode with turn instructions and other info at trkpt tags?

Cruiser uses it when BRouter 1.6.4 (Beta) is installed, with beeline, etc.

afischerdev:
@devemux86 
Yes, the new output format is moved to the standard locus-style (2). Old variante is found at timode 7 now.

vodie:
@afischerdev 
I did the Oruxmaps export on a request from a user from the velomobilforum.de with his tests in this post: https://www.velomobilforum.de/forum/index.php?threads/brouter-ein-konfigurierbarer-offline-streckenrouter-web-android.38274/post-1146912
I don't know any official documentation or repository with better information.

The turn instructions for Oruxmaps with the service interface on Android worked before I did the commit, so I don't think they use the orux-mode over the service interface. The Oruxmaps instructions are as far as I know only for the GPX-file export to Oruxmaps.

vshcherb:
Sounds ok

PR #526, opened 2023-04-04
App version check

afischerdev:
According to @polyscias [update lookups comment](https://github.com/abrensch/brouter/pull/458#issuecomment-1399432466) a app version check is added.

Rework of the handling if profile param is used.

### DIFF
--- a/brouter-expressions/src/main/java/btools/expressions/BExpressionMetaData.java
+++ b/brouter-expressions/src/main/java/btools/expressions/BExpressionMetaData.java
@@ -18,9 +18,11 @@ public final class BExpressionMetaData {
   private static final String VERSION_TAG = "---lookupversion:";
   private static final String MINOR_VERSION_TAG = "---minorversion:";
   private static final String VARLENGTH_TAG = "---readvarlength";
+  private static final String MIN_APP_VERSION_TAG = "---minappversion:";
 
   public short lookupVersion = -1;
   public short lookupMinorVersion = -1;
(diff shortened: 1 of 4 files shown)


polyscias (commented):
Thanks for taking this up. I am no Java programmer but as far as I can check the changes look good to me.

The usefulness for this is for the coming profile update maybe limited but if this functionality would have been in from the start, a non-compatible profile update would have been less error prone.

I would like to give this code a test (maybe next weekend) and see how it works but I think it would be good to do a release of the App soon after this has been checked in and the profile update only later.

afischerdev:
> I think it would be good to do a release of the App soon after this has been checked in and the profile update only later.

Thanks for the test. Yes, I also think app first.

afischerdev:
@polyscias 
Could you do some tests?
It is hard to test when there is no server to download a new and changed lookups.dat
For this situation I used this logic:
When downloaded a new lookups.dat from server it is first stored in a file named lookups.dat.tmp
And then the checks are done.
You could transfer a manipulated  tmp file from PC and then start BRouter download dialog.
When there already is a tmp file this is checked.
Current version code is 45

polyscias:
I could not find time this long weekend and yesterday did build the brouter App with the patch. Now trying to test it with a servconfig.txt file pointing to my own web server with some test files but that needs still some debug, will update once I have results.

Not sure how much time I have coming days, so better not wait for me if you want to continue.

Issue #369, opened 2021-11-21
Add Profile Config dialog


afischerdev:
For a new version I have a config dialog in stock for profiles. Works a bit like seen on BRouter-Web.

![Screenshot_20211121-114603](https://user-images.githubusercontent.com/23175706/142758950-47aa7069-841c-4f2f-b6c9-1c51565015c1.png)

It saves changes in standard preferences and has one tokenized string that could be used with 'extraParams'. So no remote profile is necessary.

PR #546, opened 2023-05-04
Add an app params dialog

afischerdev:
As suggested in #369 here a dialog to edit profile options - similar to the web client.

This can be used from inside the app with one of the six profile modes.
But it is also possible to connect from an external app to this dialog and edit the parameter from an external profile. 

The `RoutingParameterDialog` contains a small description.


poutnikl:
Interesting. It comes time to prepare my profile templates for interactive parameter tweaking. There may not be needed so many end profiles, derived from them.

Issue #539, opened 2023-04-26
Use of Sonny's LiDAR Digital Terrain Models within Europe


SonnyLidarDTM:
As far as I know Brouter uses internal elevation data which is followed for calculation routes. They are also used tocalculate elevation sums on 3rd-party apps like Locus if there's are no additional elevation files intalled. See for example https://help.locusmap.eu/topic/31440-lidar-digital-terrain-models-dtm-as-zip-possible#comment-126930

So maybe you wanna try **_"Sonny's LiDAR Digital Terrain Models"_** at https://sonny.4lima.de/ as an alternative within Europe?

The accuracy and resolution of its models' elevations compared to satellite-based surveys like ASTER, ALOS, COPERNICUS or SRTM is significantly better. Especially in wooded areas, steep rocky terrain or narrow valleys. Even if BRouter just uses the 3"-version for space saving reasons. File format (.hgt) and file size stays the same.

afischerdev:
@abrensch 
It seems you have some ideas for testing the data. Great. I have already prepared an update, but it will take some days for preparation.

quaelnix:
My home lap contains a 5 km long monotonous ascend in a valley, which does highlight the limitations of the SRTM data:
![brouter_srtm_cmp](https://github.com/abrensch/brouter/assets/122357328/0b4902ef-4c7b-4971-80ca-b16168ca54e2)
![brouter_srtm_sub](https://github.com/abrensch/brouter/assets/122357328/0e39e516-964a-4750-a129-38e3df0668f0)

zod:
I generated segments using Sonny's 3asec DTM for germany and a (slow) routing instance is available at https://brouter.knid.net/

@quaelnix Can you share this route and generate a plot which also includes the lidar data?

quaelnix:
![brouter_srtm_lidar_cmp](https://github.com/abrensch/brouter/assets/122357328/86ac9719-380d-4be7-a845-cef287525111)
![brouter_lidar_sub](https://github.com/abrensch/brouter/assets/122357328/240ad380-7db0-467b-a361-84e0fb4eba15)
![brouter_srtm_sub](https://github.com/abrensch/brouter/assets/122357328/d0a299e6-2362-4f73-9d92-82f8a5663274)

quaelnix:
![brouter_srtm_lidar_cmp](https://github.com/abrensch/brouter/assets/122357328/3f35289c-7e1b-4d2d-a82f-9994f914e42b)
[...]

PR #548, opened 2023-05-06
Add a hgt Reader

afischerdev:
According to the #539 idea here a first version of a hgt reader.
It uses plain hgt or zip with hgt inside.

The PosUnifier has a test routine for single elevations.

I tested with Iceland (is West and above 60 degree).

### DIFF
--- a/brouter-map-creator/src/main/java/btools/mapcreator/ConvertLidarTile.java
+++ b/brouter-map-creator/src/main/java/btools/mapcreator/ConvertLidarTile.java
@@ -14,9 +14,13 @@ import java.util.zip.ZipInputStream;
 public class ConvertLidarTile {
   public static int NROWS;
   public static int NCOLS;
+  public static int ROW_LENGTH;
 
   public static final short NODATA2 = -32767; // hgt-formats nodata
   public static final short NODATA = Short.MIN_VALUE;
+
+  public static final int SRTM3_ROW_LENGTH = 1200; // number of elevation values per line
+  public static final int SRTM1_ROW_LENGTH = 3600; //-- New file resolution is 3601x3601
 
   static short[] imagePixels;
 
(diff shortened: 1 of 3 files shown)


zod:
It's already possible to use hgt files using [ConvertLidarTile](https://github.com/abrensch/brouter/blob/2c707c977bdba6e2d51ac3372506f64a8ae9162f/brouter-map-creator/src/main/java/btools/mapcreator/ConvertLidarTile.java). I don't think it's useful to have two implementations.

Your implementation also uses GPL code which can't be included in a MIT licensend project.

afischerdev:
@zod 
You are right, I missed that. Was only looking in PosUnifier what is used. 
And I remember now I have used that class before for an Iceland map creation . But it was only partly usable.
I'll switch to ConvertLidarTile class and do a little rework for direct use. 
And it works only with SRTM3 files.

brouter-map-creator/src/main/java/btools/mapcreator/ConvertLidarTile.java
```diff
@@ -186,4 +217,91 @@ public static void main(String[] args) throws Exception {
...
+    raster.halfcol = halfCol5;
+    raster.noDataValue = NODATA;
+    raster.cellsize = 1. / (double) ROW_LENGTH;
+    raster.xllcorner = (int) (lon < 0 ? lon - 1 : lon); //onDegreeStart - raster.cellsize;
+    raster.yllcorner = (int) (lat < 0 ? lat - 1 : lat); //latDegreeStart - raster.cellsize;
+    raster.eval_array = imagePixels;
+
+    return raster;
+
+  }
```
zod, line 306: Code duplication. Just write one method which reads from stream and call it either using a stream created from a file or from a zip.

Issue #460, opened 2022-09-07
Round trip suggestions


nrenner:
Automatically generate circular route (loop) suggestions for a given starting point and desired route length.

Potentially with some additional constraints like ascend/difficulty.

I'm collecting links I come accross:
[Round trip suggestions (brouter-web Wiki)](https://github.com/nrenner/brouter-web/wiki/Round-trip-suggestions)

Would go well with (#258)
See also (nrenner/brouter-web#236)

EssBee59:
Very good suggestions!
I am very interested in 2 new functions:
-penalty points for routes with noise (traffic)
-plus points for pleasant routes (green, sights etc..)

afischerdev:
As we are nearly ready with the 'dynamic range' (#747) we could now offer a simple round trip logic:

From the starting point there is a direction and a distance, this will generate two additional points +/-, say 20 degrees from the direction and then routing can continue as normal.

![roundtrip](https://github.com/user-attachments/assets/9982903d-478c-4661-a00b-764703512e0e)

devemux86:
@afischerdev This is very nice!

Do you mean it will require the dynamic range?

And it will generate beelines at the via points?
This may not be desirable to users, as these points are not user-defined.

afischerdev:
@devemux86 
> Do you mean it will require the dynamic range?

Yes, I think so, when point is more then the usual 250 meters away from a way.
But no beeline generation, just find a point.
After first generation a routing client could offer the two via points for moving.

Totorrr:
BRouter engine now has a the round trip mode implemented! 

See also https://github.com/nrenner/brouter-web/issues/236#issuecomment-3371226858

devemux86:
Round trip planning with BRouter (offline) is also available in **[Cruiser](https://github.com/devemux86/cruiser)** app:

<img src="https://github.com/user-attachments/assets/fa4ed5ec-fce5-426d-aaf6-3ab4a7ec7fea" width="50%"/><img src="https://github.com/user-attachments/assets/a6a04eb6-910b-4c53-b8e1-734707329c66" width="50%"/>

Issue #222, opened 2020-01-31
Map matching


nrenner:
Snap a given track to the most likely path taken in the current road network.

A track could be a recorded GPS trace with inaccuracies and noise or a saved previous routing result that no longer matches because of map edits in the meantime, e.g. realigning roads to newer imagery.

Blog posts:
- [Matching GPS traces to a map (OSRM, Mapbox)](https://blog.mapbox.com/matching-gps-traces-to-a-map-73730197d0e2)
- [Digitalizing GPX Points or How to Track Vehicles With GraphHopper](https://karussell.wordpress.com/2014/07/28/digitalizing-gpx-points-or-how-to-track-vehicles-with-graphhopper/)
- [Map Matching in a Programmer's Perspective (Valhalla)](https://github.com/valhalla/valhalla/blob/master/docs/meili/algorithms.md)
- [Barefoot – An Open Source Java library for map matching with OpenStreetMap (BMW Car IT)](http://www.bmw-carit.com/blog/barefoot-release-an-open-source-java-library-for-map-matching-with-openstreetmap/)

Existing Code:
- [Valhalla Meili](https://github.com/valhalla/valhalla/tree/master/valhalla/meili) (C++, MIT License)
- [Open Source Routing Machine (OSRM)](https://github.com/Project-OSRM/osrm-backend) (C++, BSD-2-Clause)
- graphhopper/map-matching and bmwcarit/barefoot are Apache-2.0 licensed (incompatible with MIT)

utagawal:
Would be a very good function to allow also to collect statistics for ways surface from a GPX (useful for gravel/MTB)

PR #549, opened 2023-05-09
Introducing engineMode for future use

afischerdev:
To build a base for future an engine mode is added.
This is related to #460 or #222 or other things that will not working with 'normal' routing mode.


docs/revisions.md
```diff
@@ -2,6 +2,18 @@
 
 (ZIP-Archives including APK, readme + profiles)
 
+### New last version
+
+Android
+
+- Add parameter dialog for profile
+
+Library
```
zod, line 11: Unrelated to this PR. I think we should stick to a common format like [keepachangelog](https://keepachangelog.com/) which suggests a *unreleased* section instead of "new last version" (which should be "new *since* last version"?)

zod (changes requested):
I'm not really sure how this PR is related to the linked issues, because the modes are called "seed" and "other" but the issues request automatic route suggestion and map matching.

As with your last huge PR I would suggest first adding the feature and afterwards exposing the API because it allows you to freely change the implementation without modifying a public API.

afischerdev:
> I'm not really sure how this PR is related to the linked issues, because the modes are called "seed" and "other" but the issues request automatic route suggestion and map matching.

The idea is to open the engine to more functionality than standard routing.
The linked issues are samples to let you know what I'm thinking off. The predefined modes have to change when we have a real new feature. "seed" already exists, but I think it needs changing - anyway we should not ignore it. 

> As with your last huge PR I would suggest first adding the feature and afterwards exposing the API because it allows you to freely change the implementation without modifying a public API.

May be. For me, this way is the better way. This offers you and others the chance to place new features on these terms. Otherwise we develop nice new features and when the first one comes out with it, the others notice that the api implementation is different.

docs/revisions.md
```diff
@@ -2,6 +2,18 @@
...
+
+- Add engineMode for future use
```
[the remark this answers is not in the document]
afischerdev: For me, users also include developers who want to adapt their app accordingly.

afischerdev:
Here a first new function that uses this new entry:

`get elevation for one point`

Returns a wpt gpx with elevation when found.

zod:
I think it's not sensible to extend the `getTrackFromParams` to provide elevation data using this method. It's rather complicated and the function would not return a track but just a single waypoint in a GPX structure, but the API should just return the elevation.

Shouldn't we just extend the AIDL with a new method to retrieve implementation?

afischerdev:
@zod 
Yes, I was thinking of this as well.
We already have a hell of parameter. An extra aidl definition is a smart way but would not work for server or command line situation. 
So I would like to stay with `engineMode`. 
For the parameter hell, I'm planning a small rework to get one entry point for everyone

zod:
We could also add an API endpoint for the server to provide elevation which could be used e.g. by brouter-web to show elevation at POIs. I think brouter-web would rather consume a simple JSON instead of parsing a GPX file with waypoints just for elevation.

For the CLI it should be possible to provide a different command.

I just skimmed through the code and if we would pursue  the `engineMode` way it requires a bit of work (e.g. parameter order in `RoutingEngine` constructor)

brouter-core/src/main/java/btools/router/RoutingEngine.java
```diff
@@ -73,12 +81,19 @@ public class RoutingEngine extends Thread {
 
   public RoutingEngine(String outfileBase, String logfileBase, File segmentDir,
                        List<OsmNodeNamed> waypoints, RoutingContext rc) {
+    this(0, outfileBase, logfileBase, segmentDir,
+      waypoints, rc);
+  }
+
+  public RoutingEngine(int engineMode, String outfileBase, String logfileBase, File segmentDir,
+                       List<OsmNodeNamed> waypoints, RoutingContext rc) {
```
zod, line 89: I think the first parameters of the constructors should be the same and engineMode should be appended.

zod:
I'm quite surprised that it was merged. I still fail to understand why it's a good idea to add reading points elevation data to `RoutingEngine`. This class already does way to much.

What's your reason to not provide a new API endpoint & Android service which provides just elevation data without the GPX stuff?

afischerdev:
@zod 
>What's your reason to not provide a new API endpoint & Android service which provides just elevation data without the GPX stuff?

As discussed before the server also could use this call.
You wanted me not to add an api call without feature. So I added a small function. 
This is a sample for engineMode but main intentions are others.

PR #550, opened 2023-05-09
More PMD rules & fixes

zod:
Android Studio displays a lot of warnings for our current code. I've tried to find some PMD rules which cover the same issues to tackle them and prevent them from reappearing in the codebase.


afischerdev:
@zod 
I have that in mind. But I'm waiting for the #549 to have not to much rework.

May be it is a good way to have such reformatting around a tagged version in the future. So it's many predefined code around.

zod:
It causes just one simple conflict with #549, so I don't think that should delay the merge.

PR #555, opened 2023-05-17
Replace XML parser with pbfparser

### LINKED ISSUE #468: Merge OSM parsers
Currently BRouter contains two parsers for OSM data
* [XML parser](https://github.com/abrensch/brouter/blob/master/brouter-map-creator/src/main/java/btools/mapcreator/OsmParser.java)
* [PBF parser](https://github.com/abrensch/brouter/blob/master/misc/pbfparser/OsmParser.java)

According to the [README of PBF parser](https://github.com/abrensch/brouter/blob/master/misc/pbfparser/README.txt) the XML Parser is just for testing and I found some issues. It doesn't use a XML parser library and therefore rejects some files which are valid XML (e.g. it doesn't parser JOSM exports correctly because of different quotes).

It seems like the PBF parser should be used but it's not integrated into our build because of the osmosis dependency. This causes more work for users. @mjaschen provides a nice [Dockerfile](https://github.com/mjaschen/brouter-routingdata-builder/blob/main/Dockerfile) which handles building BRouter with PBF parser.

We use gradle / maven to get our dependencies and osmosis is available via [maven central](https://central.sonatype.dev/artifact/org.openstreetmap.osmosis/osmosis-osm-binary/0.48.3).

Perhaps we should just remove the XML parser and add osmosis as a dependency for map-creator.

zod:
As explained in #468 BRouter currently has two parsers
* "XML" Parser (doesn't really use a XML library and rejects valid XML) which should be only used for testing but is integrated into the build
* PbfParser which is used in production but requires manual download of osmosis and building

We should just use PbfParser and remove the "XML" parser. This adds a dependency on osmosis/protobuf, but the JAR file size only slightly increased to 2.2 MB. I think this slight increase is worth avoiding the building hassle.

PbfParser has an optimization which polls its input file to allow reading a pbf file while update from osmupdate is still in progress. For small files (e.g. when testing) this causes a delay of at least two minutes, therefore it's now configurable using `avoidMapPolling` (e.g. `java -davoidMapPolling=true`).

Fixes #468


afischerdev:
Worked for me - MapcreatorTest and others on Win10.

Issue #486, opened 2022-11-19
Possible extensions in the Brouter with news tags


EssBee59:
Hello,
I would like to discuss extensions in the brouter!
Many enhancement-requests are already posted, here some examples:
Noise		      https://github.com/abrensch/brouter/issues/476
Green area              https://github.com/abrensch/brouter/issues/258
Round trip               https://github.com/abrensch/brouter/issues/460

In the following I will consider (as example) these 3 enhancements:
-consider the noise / pollution on the route
-consider a river or see on the side of the route
-consider the "green" aspect of the route (within a forest or park…)
The basic idea is to introduce new calculated (or estimated) tags in the RD5 files.
(in a similar way as the existing tag „estimated_traffic_class»)

Of course, the routing engine have also to be extended in order to support these tags, and the profiles have to use them according to the preferred route.

1-	The first challenge is to calculate / estimate the tag values for the concerned highways:
I started some tests and this seems possible(see documentation in pdf file):
[Tag_calculation.pdf](https://github.com/abrensch/brouter/files/10046480/Tag_calculation.pdf)

2-	Extension of the RD5 files with the new tags
3-	Extension of the routing engine to support the new tags (+ lookups.dat)
4-	Extension of profiles
Next steps:
A first look at the results of the spatial-SQL´s is very promising: the calculated values for the new tags seems good and usable.
-a decision, which tags exactly should be implemented, can be made later
-a further challenge is the calculation of the tag value for the planet, as it will take a lot of time! 
So I suggest next to test the impact of the tags « noise » and « river » on the routing within a regional osm-map (as example 30 km * 30 km). 

The tags values are available, but implementing Extension(2) and (3) is prerequisite to start “real” tests.
Is anybody ready to work on this extension / to create a prototype for testing(Extensions 2 and 3)?

afischerdev:
I'm sorry for the delay.

It's not that I haven't heard the wishes.
But all my resources are in the 'final stages' for the 1.6.4 library.
And usually I want to have sampled some code snippets before making a statement/commitment.

So all I can say is: yes, I'm interested, I find it useful, it might start for me after 1.6.4

EssBee59:
> I'm sorry for the delay.
> 
> It's not that I haven't heard the wishes. But all my resources are in the 'final stages' for the 1.6.4 library. And usually I want to have sampled some code snippets before making a statement/commitment.
> 
> So all I can say is: yes, I'm interested, I find it useful, it might start for me after 1.6.4

[...]

PR #556, opened 2023-05-20
Add jdbc import

afischerdev:
This is according to discussion #486 and adds tags via database call to the way generation.
New is 
forest to find nice ways in or near forest
river to file ways along water
noise to avoid bigger streets
town to avoid bigger citys
traffic was used before but is now generated by database call

The classpath in generation needs a jdbc driver
and the OsmFastCutter a call with some database parameter
Please see readme.

There is also a new lookups.dat that contains the new tags - before the node context. This should be neutral to older versions. 
In the next generation lookups.dat (#458) I would prefer these new tags more moved up to get smaller rd5 files.

This needs also an update for the profiles to make use of the new tags.

### DIFF
--- a/brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java
+++ b/brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java
@@ -17,6 +17,7 @@ import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.HashMap;
 import java.util.Map;
 
 import btools.expressions.BExpressionContextNode;
@@ -40,22 +41,16 @@ public class OsmCutter extends MapCreatorBase {
 
 
   Connection conn = null;
-  // PreparedStatement psNoise = null;
-  // PreparedStatement psRiver = null;
-  // PreparedStatement psForest = null;
   PreparedStatement psAllTags = null;
 
   ResultSet rsBrouter = null;
 
   int cntHighways = 0;
   int cntWayModified = 0;
-  int cntNewNoise = 0;
-  int cntNewRiver = 0;
-  int cntNewForest = 0;
-  int cntNewTown = 0;
-  int cntNewTraffic = 0;
 
   String jdbcurl;
+  Map<String, String> databaseField2Tag;
+  Map<String, Integer> databaseFieldsFound;
 
   public static void main(String[] args) throws Exception {
     System.out.println("*** OsmCutter: cut an osm map in node-tiles + a way file");
@@ -79,9 +74,6 @@ public class OsmCutter extends MapCreatorBase {
   private BExpressionContextWay _expctxWay;
   private BExpressionContextNode _expctxNode;
 
-  // private BExpressionContextWay _expctxWayStat;
-  // private BExpressionContextNode _expctxNodeStat;
-
   public void process(File lookupFile, File outTileDir, File wayFile, File relFile, File resFile, File profileFile, File mapFile) throws Exception {
     if (!lookupFile.exists()) {
       throw new IllegalArgumentException("lookup-file: " + lookupFile + " does not exist");
@@ -93,10 +85,6 @@ public class OsmCutter extends MapCreatorBase {
     _expctxNode = new BExpressionContextNode(meta);
     meta.readMetaData(lookupFile);
     _expctxWay.parseFile(profileFile, "global");
-
-
-    // _expctxWayStat = new BExpressionContextWay( null );
-    // _expctxNodeStat = new BExpressionContextNode( null );
 
     this.outTileDir = outTileDir;
     if (!outTileDir.isDirectory())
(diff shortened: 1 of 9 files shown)


abrensch:
I tried to run this branch on the "mapcreator" account of new dev server, but I'm stuck with a very strange effect: Call to OsmFastCutter is returning immediatly, but the process is running on, so script is continuing with the second step (PosUnifier) but that fails...

Probably has noting to to with the jdbc option, more with the new PBF-Parser?

afischerdev:
@abrensch 
I rechecked the jdbc handling.
wrong driver
wrong database
wrong user
wrong password
On all errors the process is killed by System.exit
And next process stops with 'bordernids.dat not found'.

So sounds as this is jdbc problem.
I could change this: remove System.exit with a smoother handling

abrensch:
> So sounds as this is jdbc problem. I could change this: remove System.exit with a smoother handling

A non-zero exit code would be smooth enough :-)

However, I still do not understand why the process keeps running. Looks like a second thread forked or the like (?)

afischerdev:
@abrensch 
Ok I will change to exi(1).
On my run next process stops with 'bordernids.dat not found'. May be no cleaning first?

I was preparing for smoother handling, but not sure if it's a good idea to continue a process when no database call is in progress. The (incorrect) results will therefore only be displayed to you once the entire process has been completed.

brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java
```diff
@@ -165,6 +192,93 @@ private void generatePseudoTags(Map<String, String> map) {
   }
 
 
+// Ess Bee : NEW function to add the new tags (estimated_noise_class , river, forest...)
+
```
zod, line 196: Remove coment

brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java
```diff
@@ -165,6 +192,93 @@ private void generatePseudoTags(Map<String, String> map) {
...
+  private void generateSpecialTags(long osm_id, Map<String, String> map) {
+
+    if (jdbcurl == null) return;
+
+    // is the database allready connected?
+    if (conn == null) {
+
+      String sql_all_tags = "SELECT *  from all_tags where losmid = ?";
+
+      System.err.println("OsmCutter start connection to the database........" + jdbcurl);
```
zod, line 206: That's not an error, should not print to standard error.

brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java
```diff
@@ -165,6 +192,93 @@ private void generatePseudoTags(Map<String, String> map) {
...
+
+      try {
+        conn = DriverManager.getConnection(jdbcurl);
+        psAllTags = conn.prepareStatement(sql_all_tags);
+
+        System.err.println("OsmCutter connect to the database ok........");
```
zod, line 212: That's not an error, should not print to standard error.

brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java
```diff
@@ -165,6 +192,93 @@ private void generatePseudoTags(Map<String, String> map) {
...
+            }
+
+            if (rsBrouter.getString("town_class") != null) {
+              map.put("estimated_town_class", rsBrouter.getString("town_class"));
+              cntNewTown = cntNewTown + 1;
+            }
+
+            if (rsBrouter.getString("traffic_class") != null) {
+              map.put("estimated_traffic_class", rsBrouter.getString("traffic_class"));
+              cntNewTraffic = cntNewTraffic + 1;
```
zod, line 259: Lots of copy-pase code. We should store source tags and BRouter tags in a map and just lookup the mapping. Same would work for those counters.

brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java
```diff
@@ -165,6 +192,93 @@ private void generatePseudoTags(Map<String, String> map) {
...
+    if (jdbcurl == null) return;
+
+    // is the database allready connected?
+    if (conn == null) {
+
+      String sql_all_tags = "SELECT *  from all_tags where losmid = ?";
+
+      System.err.println("OsmCutter start connection to the database........" + jdbcurl);
+
+      try {
```
zod, line 208: This try-block shoudn't be nested in the `if (conn == null)` block

brouter-map-creator/src/main/java/btools/mapcreator/OsmFastCutter.java
```diff
@@ -12,8 +12,8 @@
 public class OsmFastCutter extends MapCreatorBase {
   public static void main(String[] args) throws Exception {
     System.out.println("*** OsmFastCutter: cut an osm map in node-tiles + way-tiles");
-    if (args.length != 11 && args.length != 12) {
-      String common = "java OsmFastCutter <lookup-file> <node-dir> <way-dir> <node55-dir> <way55-dir> <border-file> <out-rel-file> <out-res-file> <filter-profile> <report-profile> <check-profile>";
+    if (args.length != 11 && args.length != 12 && args.length != 13) {
+      String common = "java OsmFastCutter <lookup-file> <node-dir> <way-dir> <node55-dir> <way55-dir> <border-file> <out-rel-file> <out-res-file> <filter-profile> <report-profile> <check-profile> <map-file> <jdbc-url>";
```
zod, line 16: Optional parameters are usually in square brackets e.g. `[jbdc-url]`

misc/scripts/mapcreation/brouter.sql
```diff
@@ -0,0 +1,649 @@
+--  calculation of new tags (estimated_noise_class, estimated_river_class,estimated_forest_class, estimated_town_class, estimated_traffic_class)
+--  Ess Bee version 08/05/2023
```
zod, line 2: We should use git for versioning, not an additional number

afischerdev:
> * `.sql` and `.lua` don't use a consistent format, perhaps run some formatter/beautifier

Not easy to find a formatter, most breaking on postgres functions. I used https://sqlformat.darold.net/ online formatting for now, but also this breaks with errors on database generation. That needs more time.

> * documentation in mapcreation folder: Should we keep it there or move it to docs?

Yes, we should. But at the moment this is more a todo list for finding a right script for generation.
At the end we have a new db_gen.sh and new rd5_gen.sh. May be an db_update.sh
All this is not ready. 

> * absolutly no tests: It requires quite some setup to import OSM data into postgis and process it. For tests we should perhaps use some (in-memory) SQLite database which has an `all_tags` table.

Yes, and I don't have an idea for a change. The scripts need postgres functions. So a SQLite could only hold pregenerated values. But that is a fake test.

> * Is it possible to update the database using osm2pgsql?

I hope so. But it is not developed or tested.

> * Should the database use more (materialized) views instead of tables because we don't need to modify the data?

This is not tested as well. But yes, it is an idea for the future.

abrensch:
I don't get it....
OsmFastCutter keeps putting itself in the background as soon as there is a 13. argument looking slighty like a jdbc-url. Not for "xyz", but for "jdbc:oracle:..." and even without the postgres driver in the classpath
And from the code I would expect that the JDBC-Logic is only initialized when the first Osm-Way is decoded (which would be after half an hour decoding nodes), but it happens immediaty...

afischerdev:
>  I would expect that the JDBC-Logic is only initialized when the first Osm-Way ...

That is the way it works. Please see [this line](https://github.com/afischerdev/brouter/blob/6158c44d82c17df5ae2e8ad2a089b4a0764ea22e/brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java#L263) in nextWay routine. I was able to check it in my logs.
OsmFastCutter only forwards the jdbc argument to OsmCutter ([please see here](https://github.com/afischerdev/brouter/blob/6158c44d82c17df5ae2e8ad2a089b4a0764ea22e/brouter-map-creator/src/main/java/btools/mapcreator/OsmFastCutter.java#L43)). 

But the mapping to "jdbc:oracle:..." sound strange.

abrensch:
> OsmFastCutter keeps putting itself in the background as soon as there is a 13. argument looking slighty like a jdbc-url

Blame on me... the "&" in the jdbc-url is interpreted by the shell... just needs quotes.

zod:
> Yes, and I don't have an idea for a change. The scripts need postgres functions. So a SQLite could only hold pregenerated values. But that is a fake test.

It's pretty common to use mocks/fakes/doubles to test functionality on a unit or module level. This also establishes a clear boundary because you can easily see which data is expected by `generateTagsFromDatabase`.

PR #570, opened 2023-06-12
Also consider cycleway:left:oneway & cycleway:right:oneway in onewaypenalty

Totorrr:
This detects some oneway cases better.


quaelnix:
Works well as far as I can tell.

Totorrr:
Thank you @afischerdev & @quaelnix _/\_

Issue #566, opened 2023-06-05
Profile-Compiler: constant expressions optimizations


abrensch:
Profile-Compiler: constant expressions optimizations

Currently the Profile-Compiler (BExpressionContext.parseFile) creates an (ordered) list of "expression trees" that are not optimized further.

However, chances are that major subtrees always evaluate to a constant value - they can be replaced by that constant value at compile time.

Focus here is not such much the runtime-cost of traversing these constant subtrees - this runtime cost is low as long as the number of distinct tag-value sets is low thus profile evaluation works mainly from the profile-cache.

So the focus of this issue here is to keep the number of distinct tag-value sets low by marking as much tagnames as possible as "unused" (in the sense of "processUnusedTags = false".  So the compiler should be able o mark a tagname as unused though it is used inside a constant subtree.

This optimization is neccecary in order to integrate the new pseudo-tags in the standard profiles without creating a performance impact.

Example for such constant subtrees:

assign   consider_town       = false

assign town_penalty 
   switch consider_town
     switch estimated_town_class=  0
     switch estimated_town_class=1  0.5
     switch estimated_town_class=2  0.9
     switch estimated_town_class=3  1.2
     switch estimated_town_class=4  1.3
     switch estimated_town_class=5  1.4
     switch estimated_town_class=6  1.6 99 0

Here, the compiler should at least be able to mark "estimated_town_class" as unused.

Second level would be to alse replace the expression by:

  assign town_penalty  = 0
  
and third level is to replace also the usage:

  add town_penalty XY
  
by just:

  XY

afischerdev:
@abrensch 
OK, I saw it. And will start with that. But in the next few weeks my time is a bit limited.

abrensch:
> OK, I saw it. And will start with that. But in the next few weeks my time is a bit limited.

Sorry, assignig to you was just by accident. I will look into it, but commit via merge request and welcome a review by you. It's easy to introduce subtle bugs. When looking into the code, I already realized there might by a glitch with numerically evaluated tags not leading to marking the tag as "used".

afischerdev:
I add a small new logic for global expressions.
Please see my [repo profile-compiler](https://github.com/afischerdev/brouter/tree/profile-compiler)
It removes the unused tags from messages list.

abrensch:
I created a merge request for my patch: https://github.com/abrensch/brouter/pull/573

abrensch:
> Would the "second level" catch this?

[...]

PR #573, opened 2023-06-25
optimizing constant expressions in profile parsing

abrensch:
This is the patch for the corresponding issue https://github.com/abrensch/brouter/issues/566

It improves the handling of constant expressions in the profile compiler

Visible effect should be that tags that are unused by configuration (e.g. estimated_forest_class) do not appear in the table if process_unused_tags=false

Tested so far via the unit tests only.

Unit tests had found a problem with injected key-values so I had to change the handling of these. An Injected value is now injected as a Number-Expression-Operand of the assign-expression that assigns the variable.

### DIFF
--- a/brouter-core/src/main/java/btools/router/RoutingContext.java
+++ b/brouter-core/src/main/java/btools/router/RoutingContext.java
@@ -108,16 +108,6 @@ public final class RoutingContext {
 
   public void readGlobalConfig() {
     BExpressionContext expctxGlobal = expctxWay; // just one of them...
-
-    if (keyValues != null) {
-      // add parameter to context
-      for (Map.Entry<String, String> e : keyValues.entrySet()) {
-        float f = Float.parseFloat(e.getValue());
-        expctxWay.setVariableValue(e.getKey(), f, true);
-        expctxNode.setVariableValue(e.getKey(), f, true);
-      }
-    }
-
     setModel(expctxGlobal._modelClass);
 
     carMode = 0.f != expctxGlobal.getVariableValue("validForCars", 0.f);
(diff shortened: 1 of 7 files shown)


brouter-core/src/main/java/btools/router/RoutingContext.java
```diff
@@ -108,16 +108,6 @@ public long getKeyValueChecksum() {
...
-
-    if (keyValues != null) {
-      // add parameter to context
-      for (Map.Entry<String, String> e : keyValues.entrySet()) {
-        float f = Float.parseFloat(e.getValue());
-        expctxWay.setVariableValue(e.getKey(), f, true);
-        expctxNode.setVariableValue(e.getKey(), f, true);
-      }
-    }
-
```
afischerdev, line 120: Please keep this part to add extra parameter.

brouter-expressions/src/main/java/btools/expressions/BExpressionContext.java
```diff
@@ -857,6 +866,7 @@ int getVariableIdx(String name, boolean create) {
       if (create) {
         num = variableNumbers.size();
         variableNumbers.put(name, num);
+        lastAssignedExpression.add(null);
       } else {
```
afischerdev, line 870: When you keep the parameter inserts, this needs a Null check like
if (lastAssignedExpression != null) lastAssignedExpression.add(null);

brouter-core/src/main/java/btools/router/RoutingContext.java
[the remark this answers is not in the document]
abrensch: Got it. I reworked the keyValue injection, but in a way that keeps a single data-path and works at parsetime only.

I also added a unit-test, and extended the "ProfileComparator" to support manual testing.

abrensch:
Repeating my last comment that is hidden in the "resolved conversation":

abrensch:
Repeating my last comment that is hidden in the "resolved conversation": 

Got [your point about injecting keys not defined in the profile]. I reworked the keyValue injection, but in a way that keeps a single data-path and works at parsetime only.

I also added a unit-test, and extended the "ProfileComparator" to support manual testing.

afischerdev:
Great, this worked again for me and my tests.
Only 'consider_traffic' in fastbike profile should be reworked. This is not reachable.

PR #578, opened 2023-07-05
Add support for downloading profiles

jelmer:
Add support for downloading profiles. This is useful when e.g. running the standalone server in docker.

### DIFF
--- a/brouter-server/src/main/java/btools/server/RouteServer.java
+++ b/brouter-server/src/main/java/btools/server/RouteServer.java
@@ -30,12 +30,13 @@ import btools.router.ProfileCache;
 import btools.router.RoutingContext;
 import btools.router.RoutingEngine;
 import btools.server.request.ProfileUploadHandler;
+import btools.server.request.ProfileDownloadHandler;
 import btools.server.request.RequestHandler;
 import btools.server.request.ServerHandler;
 import btools.util.StackSampler;
 
 public class RouteServer extends Thread implements Comparable<RouteServer> {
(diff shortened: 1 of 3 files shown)


abrensch:
I'm not sure what profiles you want to download. Custom-Profiles? We try to prevent this for data protection reasons. If not intended, this i sa least possivble via "../" prefixes with this patch. Or just the default profiles that are available via https://brouter.de/brouter/profiles2 ?

jelmer:
I've got an application that interacts with my standalone brouter instance over the API. I'd like to download just regular profiles that are available on the server (but edited) - not necessarily custom ones.

afischerdev:
The standard profiles are already checked for changes and downloaded to phone. This happens when you update your map tiles. Please see [DownloadWorker](https://github.com/abrensch/brouter/blob/18f2cb548f29ae0df6ec8b98911fbc2aebbfc2e3/brouter-routing-app/src/main/java/btools/routingapp/DownloadWorker.java#L212)

jelmer:
> The standard profiles are already checked for changes and downloaded to phone. This happens when you update your map tiles. Please see [DownloadWorker](https://github.com/abrensch/brouter/blob/18f2cb548f29ae0df6ec8b98911fbc2aebbfc2e3/brouter-routing-app/src/main/java/btools/routingapp/DownloadWorker.java#L212)

I'm not using the app, but the standalone server in a Linux container.

jelmer:
Closing this, since I've migrated away from brouter for my bicycle routing. If somebody else needs it feel free to pick it up.

PR #579, opened 2023-07-06
Update profiles for new db tags

afischerdev:
Since the server works with new database tags we need a profile update.
Also add the idea from [web-client](https://github.com/nrenner/brouter-web/issues/747) issue.
Lookups.dat is replaced and sql-script updated.


devemux86:
Are the new tags usable in the car profiles?

afischerdev:
@devemux86 
Yes, I think so, but I don't have a convertible car for testing.

misc/profiles2/hiking-mountain.brf
```diff
@@ -10,16 +10,22 @@
 
 ---context:global
 
-assign   consider_elevation       true    # %consider_elevation% | Set to false, otherwise less interesting flat roads are chosen  | boolean
+assign   consider_elevation     = false   # %consider_elevation% | Activate to prefer a route with few elevation meters | boolean
```
quaelnix, line 13: Was the change from `true` to `false` intentional?

quaelnix:
Reformatting the code and changing things at the same time makes it extra hard to read the diff. Please don't do that.

---

I think the the new penalties:
https://github.com/abrensch/brouter/blob/e66468b091506833eb5890656a0ef3a538397ab2/misc/profiles2/fastbike.brf#L239-L242
should be added after this line:
https://github.com/abrensch/brouter/blob/e66468b091506833eb5890656a0ef3a538397ab2/misc/profiles2/fastbike.brf#L247

quaelnix:
The idea behind the `min 9999` thing is to ensure that all ways below this line are taken into account by the navigation hint generator. If you add new costs on arbitrary way segments above this line you risk breaking this logic.

afischerdev:
@quaelnix 
You are right, it works but it brings different result values.

EssBee59:
It appears that the way we firstly added the "no_river_penalty" and "no_forest_penalty" can cause irritations:

Example:
When "consider_river" was selected, a basic cost was added when no tag found (no river here) 

assign no_river_penalty
   switch consider_river
     switch estimated_river_class=  3
     switch estimated_river_class=1  2
     switch estimated_river_class=2  1.5
     switch estimated_river_class=3  1
     switch estimated_river_class=4  0.5
     switch estimated_river_class=5  0.2
     switch estimated_river_class=6  0 99 0

Also when no river is found at all on the possible routes, the new routing differed a bit from the original routing without the consider_river option.
To avoid irritations with this effect, I now prefer the following (new) logic:

Assuming the minimum costfactor in the profile is 1
1-multiply ALL the cost parameters by 4. (the min costfactor is now 4, but the routing did not change (only the cost-sum changed)
2-change the "no_river_penalty" and "no_forest_penalty" in "river_advantage" and "forest_advantage" 

example: 
assign river_advantage 
   switch consider_river
     switch estimated_river_class=  0
     switch estimated_river_class=1  -1
     switch estimated_river_class=2  -1.5
     switch estimated_river_class=3  -2
     switch estimated_river_class=4  -2.5
     switch estimated_river_class=5  -2.7
     switch estimated_river_class=6  -3 99 0

It is important to keep the costfactor >= 1, so "-3" should be here the highest possible advantage
As recommended, never select "consider_river" and"consider_forest" together.


  
[racebike-verylowtraffic.brf.txt](https://github.com/abrensch/brouter/files/12319318/racebike-verylowtraffic.brf.txt)

devemux86:
> As recommended, never select "consider_river" and"consider_forest" together.

@EssBee59

Are there other combinations of the new tags that we should avoid using together?
consider_forest, consider_river, consider_town, consider_noise, consider_traffic

EssBee59:
My remark about "never---" was not thought functionally but technically:
With the new profile version above, selecting "consider_river" and"consider_forest" together could lead to a situation where costfacor is lower than 1
(I had trouble with such a situation in the past with a costfactor = 0.05)

In the following new version (attached below), I added at the beginning of costfactor "max 1 .....", and with that change the "never" can be ignored!
To your question:
Theoretically no restriction exists on the use of the options.
But I did not check the result is, if as example you want river and forest at the same time!
For my own, I only use maximum at the same time "consider_elevation" + "consider_traffic" + 1 further consider option.  

[racebike-verylowtraffic.brf.txt](https://github.com/abrensch/brouter/files/12320195/racebike-verylowtraffic.brf.txt)

Stapawe:
> > As recommended, never select "consider_river" and"consider_forest" together.
> 
> @EssBee59
> 
> Are there other combinations of the new tags that we should avoid using together?

The averaging effects of adding penalties (examples in the issue mentioned above), by decreasing cost ratios between ways, is the stronger, the bigger penalties are added to the bigger number of ways, so roads without penalties are rare, so it depends on area.

Issue #551, opened 2023-05-12
Download view in portait


afischerdev:
Is it possible that BInstallerActivity not declared as forced landscape in BRouter [AndroidManifest.xml](https://github.com/abrensch/brouter/blob/master/brouter-routing-app/src/main/AndroidManifest.xml) ?

It is a bit annoying to have to change orientation to download BRouter files.
Doesn't the map work with portrait orientation like all other map apps?

abrensch:
If you do that without being prepared for the hard way of really doing responsive Design, then this will Just Produce Bug Reports..

zod:
The map view would work in portait but android recreates the view during after rotation and I'm not sure if we would recreate all required information.

afischerdev:
We will see. It was just a user request that I didn't want to ignore

PR #582, opened 2023-07-07
Add portrait mode to app

afischerdev:
Based on issue #551, the download view can now also be loaded in portrait mode.
This is only possible once at the beginning. Cannot switch while in use.


devemux86:
@afischerdev have you tried the manifest change mentioned here? https://github.com/abrensch/brouter/issues/551#issuecomment-1545650103

Although it's probably not a normal map view, it could switch while in use?

afischerdev:
@devemux86 
No, it's only active on start, not on use.

Issue #568, opened 2023-06-09
Start marker is to far from a route


0709wiwiwi:
What is wrong ?

https://youtu.be/G82JGCHp6NE

[brouter.gpx.txt](https://github.com/nrenner/brouter-web/files/11709510/brouter.gpx.txt)
https://bikerouter.de/#map=13/24.3099/43.5004/standard&amp;lonlats=43.501053,24.328795;43.490753,24.33788&amp;profile=car-eco

devemux86:
@0709wiwiwi the link does not seem to work, better use regular "&".

quaelnix:
The link works fine, but I don't know what the issue is:
- This works: https://brouter.de/brouter-web/#map=13/24.3099/43.5002/standard&lonlats=43.496781,24.332315;43.508218,24.321775
- This does not: https://brouter.de/brouter-web/#map=13/24.3099/43.5002/standard&lonlats=43.501697,24.327877;43.508218,24.321775

Both times the start marker is on the same OSM way: https://www.openstreetmap.org/way/31462380

devemux86:
Another case discussed here: https://github.com/devemux86/cruiser/discussions/205

- car-fast works: [route link](https://brouter.de/brouter-web/#map=12/24.2360/43.4667/standard&lonlats=43.459525,24.278186;43.499077,24.213858&profile=car-fast)
- car-eco not work: [route link](https://brouter.de/brouter-web/#map=12/24.2360/43.4667/standard&lonlats=43.459525,24.278186;43.499077,24.213858&profile=car-eco)
  - vmax ≥ 95 km/h works

In this case @poutnikl [CarProfiles](https://github.com/poutnikl/Brouter-profiles/tree/master/CarProfiles) seem to work.

poutnikl:
> 
> In this case @poutnikl [CarProfiles](https://github.com/poutnikl/Brouter-profiles/tree/master/CarProfiles) seem to work.

That is funny, as I consider them obsolete, not updating them for years, because of the internal kinematic model, developed by abrensch for his car profiles. OTOH, they can be used as a fallback in case something goes wrong in recent development.

afischerdev:
We have two problems around this discussion - all depend on the large distances in this area.

1. based on the sample above: `car-eco not work`: [route link](https://brouter.de/brouter-web/#map=12/24.2360/43.4667/standard&lonlats=43.459525,24.278186;43.499077,24.213858&profile=car-eco)
[...]

PR #583, opened 2023-07-08
Find matching points in areas with longer distance between way points

afischerdev:
According to issue #568 two changes for get better results on long distances around searching points.

1. Do a second round when a way point is not found to get all square around the search square
2. Add a greater difference value for a finishing check (rc.waypointCatchingRange wasn't enough)

A 3rd idea is drop for now - tile generation: add pseudo points when no points in an areas.

### DIFF
--- a/brouter-core/src/main/java/btools/router/RoutingEngine.java
+++ b/brouter-core/src/main/java/btools/router/RoutingEngine.java
@@ -1328,7 +1328,8 @@ public class RoutingEngine extends Thread {
         }
 
         // recheck cutoff before doing expensive stuff
-        if (path.cost + path.airdistance > maxTotalCost + 100) {
+        int addDiff = 300;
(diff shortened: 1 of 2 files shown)


devemux86:
> waypointCatchingRange wasn't enough

However, using the waypointCatchingRange allows for custom (larger) range as this variable can be changed by apps?

afischerdev:
The problem was on the other side of route generation. The error message was in both cases 'from position not found'. 
In first case this was true and it's useful to start a second round.
In second case the engine doesn't find the end of the route and starts a new round of route searching, where the from point is not found.
The value is not only checked for distance, cost and distance is used. So it may be better to add a new finishing tolerance value than using waypointCatchingRange.

afischerdev:
BRouter runs several rounds of routing before finish. The previous run has an 'airDistanceCostFactor' and we compare the last result with the new costs. so I used the old factor again for compare. Now we can stay with the old tolerance of 100.

PR #591, opened 2023-07-15
Use better named `enums` for voice hints internally

rkflx:
The re-indexing revert is concise and contained in the first patch. Let me know if you prefer to keep or drop the additional cleanup patches on top of that (see commit messages for why I think those might be beneficial).

Additional notes:
- What is the intended meaning of `OFFR`: "off route" (as referred to in the respective comment, added in 23d1812371) or "offroad" (as mapped for Cruiser)? Note that it is actually never used anywhere in the whole codebase. Still I'd like to name the `enum` constant correctly.
- The output of the "BRouter" format (`turnInstructionMode=9`) has not been touched for now. Though in case that format is meant only as a debugging aid to dump the internal names (now `enum` constants), I can still change it to do that. Just let me know.
- If and when to support strings in the GeoJSON/JSON API (either mandatory or as an option) can be decided later, depending on @nrenner's plans.
- Whether or not to go forward with changing the meaning of `TU` for OsmAnd / `turnInstructionMode=3` (before/now: `u-turn-left`, planned: `u-turn-180`), is off-topic here and in the linked issue. Personally I would recommend against it (i.e. rather keep old releases compatible, in new OsmAnd releases only add but do not change hints), but please open another issue if you want to discuss or work on that. Other formats are mapped to fallback solutions anyway (e.g. Orux, old Locus, Gpsies), or already implemented the change (new Locus, Cruiser), AFAICT.

### DIFF
--- a/brouter-core/src/main/java/btools/router/OsmTrack.java
+++ b/brouter-core/src/main/java/btools/router/OsmTrack.java
@@ -1349,7 +1349,7 @@ public final class OsmTrack {
           rc.turnInstructionMode == 9) {
           MatchedWaypoint mwpt = getMatchedWaypoint(nodeNr);
           if (mwpt != null && mwpt.direct) {
-            input.cmd = VoiceHint.Command.BL;
+            input.cmd = VoiceHint.Command.BEELINE;
             input.angle = (float) (nodeNr == 0 ? node.origin.message.turnangle : node.message.turnangle);
             input.distanceToNext = node.calcDistance(node.origin);
           }
--- a/brouter-core/src/main/java/btools/router/VoiceHint.java
+++ b/brouter-core/src/main/java/btools/router/VoiceHint.java
@@ -11,22 +11,22 @@ import java.util.List;
 
 public class VoiceHint {
   enum Command {
-    C, // continue (go straight)
-    TL, // turn left
-    TSLL, // turn slightly left
-    TSHL, // turn sharply left
-    TR, // turn right
-    TSLR, // turn slightly right
-    TSHR, // turn sharply right
-    KL, // keep left
-    KR, // keep right
-    TLU, // U-turn
-    TU, // 180 degree u-turn
-    TRU, // Right U-turn
-    OFFR, // Off route
-    RNDB, // Roundabout
-    RNLB, // Roundabout left
-    BL, // Beeline routing
+    CONTINUE,
+    TURN_LEFT,
+    TURN_SLIGHTLY_LEFT,
+    TURN_SHARPLY_LEFT,
+    TURN_RIGHT,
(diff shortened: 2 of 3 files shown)


afischerdev:
Could you split the PR into re-index and add enums?
The first two commits [82fecf9](https://github.com/abrensch/brouter/pull/591/commits/82fecf95f6f1803cbff5acd172516b6d8897fc8e) and [d98b106](https://github.com/abrensch/brouter/pull/591/commits/d98b1060d48dac056b39fcb423ec967c26792517) bring the expected changes only on json export. 

The enums need more testing. But with a split you get soon a library for your work on brouter-web.

quaelnix:
> The enums need more testing.

Why? The last two commits both only contain non-functional code changes or did I miss something?

afischerdev:
I would like to test against speed and space requirements.

quaelnix:
> I would like to test against speed and space requirements.

@afischerdev the speed difference seems to be negligible on my machine (less than 0.1%, tested on 1k random routes).

quaelnix:
@afischerdev, runtime of [`totaltrack.processVoiceHints(routingContext);`](https://github.com/abrensch/brouter/blob/0b6608eddbeaad443e54825a27af769cd510557a/brouter-core/src/main/java/btools/router/RoutingEngine.java#L589) measured with `System.nanoTime()`:
processVoiceHints
To be honest, I don't understand why we spend time on this.

quaelnix:
@afischerdev, runtime of
https://github.com/abrensch/brouter/blob/0b6608eddbeaad443e54825a27af769cd510557a/brouter-core/src/main/java/btools/router/RoutingEngine.java#L589
measured with `System.nanoTime()`:
processVoiceHints
We should stop spending time on this. There is literally no measureable difference.

quaelnix:
@afischerdev, it was a mistake that we did not merge this pull request without further ado. The improved readability and maintainability would have easily outweighed the possible performance loss of at most a few microseconds per route request.

afischerdev:
@quaelnix 
> it was a mistake

May be, but I didn't close it. 
But I'm with @rkflx there is no sense to discuss again what others have already done. Until we have new facts on better measurement.
And I agreed with you the timing results are not significant. So I don't mind on both ways.

@rkflx 
But I don't understand why you canceled the new idea of string based voice hints. This is possible in both variants.

It was just a file based size measurement. I got three classes: VoiceHint, VoiceHint$1, VoiceHint$Command.
When you assume the extra classes are used only in a single way, you get 
```
trekking 164 wpt =    67.136
car-fast  13 wpt =     8.397
hiking-m 357 wpt =   142.213
```
All of these are guessed values only. The silver bullet would be the size of the ArrayList. But who wants to spend more time doing that?

PR #592, opened 2023-07-16
Update to new beta version

afischerdev:
For the next changes I made an update on the version number to  **1.7.2-beta-1**


afischerdev:
Yes, you are right. I missed the right time for the merge -  again. 
Anyway I would like to make a new release from the re-indexed version. Then time is come.

Therefore I need some other small things before (bundle generation, manual publish run) and I will make an update on OsmTrack not to break on non-static method getClass() for the #457.

Issue #561, opened 2023-05-31
Normalizing the profiles


afischerdev:
We have to change the profiles when the issue #486 goes online and the PR #458 is merged.
So it is a good time to think about normalization of the profiles. 
What I see at the moment:

- our standard profiles have different handling for access rules
- make variable tags (editable in BRouter-Web) available in all profiles
- check if a rule is better to be fully banned or if a heavy penalty should be added

What do you think? Other ideas?

Totorrr:
Hello @afischerdev or @quaelnix, 

could you give me an (even fictional) example for "check fastbike for excluding unused constants"? Are you referring to the constants at the beginning of the file that would not be used (currently, I think there are none) or are you referring to writing the profile in such a way, that for example, when "consider_traffic" is false, some blocks of the profiles relating to traffic would be "bypassed" for better routing engine performance? Or something else? :)

Thank you,

afischerdev:
@Totorrr 
Thank you for looking into the still unresolved problems.

The unused constants are removed by @quaelnix with this [changes](https://github.com/abrensch/brouter/pull/598/commits/c38a9186fd9ce4c50a8a3a1063678fe079fb5be7).

The `consider_traffic` problem is still present. The `estimated_traffic_class` is always used in the fastbike profile. The variable `consider_traffic` is used a little bit different to the other consider_... . An extra switch is needed, something like
```
assign trafficpenalty =
      switch not equal consider_traffic 5
      ...
      0
```
An example could be every route with `estimated_traffic_class` like [this](https://brouter.de/brouter-web/#map=10/47.6696/9.4392/standard&lonlats=9.054709,47.814893;9.749111,47.506744&profile=fastbike)

Totorrr:
Hello @afischerdev,

> The `consider_traffic` problem is still present. The `estimated_traffic_class` is always used in the fastbike profile. The variable `consider_traffic` is used a little bit different to the other consider_... . An extra switch is needed

In the [current](https://github.com/abrensch/brouter/blob/037b3385a74af230feadfe2f180499699e1c66fd/misc/profiles2/fastbike.brf#L162-L164) `fastbike.brf`, I see this: 

https://github.com/abrensch/brouter/blob/037b3385a74af230feadfe2f180499699e1c66fd/misc/profiles2/fastbike.brf#L162-L164

It seems to me that the switch is already there. Is there something that you see me missing @afischerdev? 

Thanks.

afischerdev:
@Totorrr 
[...]

PR #593, opened 2023-07-16
Fix regression in fastbike profile

quaelnix:
https://github.com/abrensch/brouter/pull/579 broke the logic that handled `highway=path` trying to add the missing `avoid_path` logic.

Example: [http://brouter.de/brouter-web/#map=15/51.3734/7.0430/...&profile=fastbike](http://brouter.de/brouter-web/#map=15/51.3734/7.0430/standard&lonlats=7.035675,51.375211;7.049965,51.371661&profile=fastbike)

See: https://github.com/nrenner/brouter-web/issues/756

---

Both `avoid_unsafe` and `avoid_path` dysfunctional. We should either remove these options or implement them properly.

### DIFF
--- a/misc/profiles2/fastbike.brf
+++ b/misc/profiles2/fastbike.brf
@@ -261,8 +261,7 @@ assign costfactor
   switch    highway=cycleway                          1.3
   switch    isresidentialorliving                     switch isunpaved 10 1.2
   switch    highway=service                           switch isunpaved 10 1.2
-  switch    highway=path                              switch avoid_path 2.1 1.1
-  switch or highway=track or highway=road highway=footway
+  switch or highway=track or highway=road or highway=path highway=footway
    switch tracktype=grade1 switch isunpaved 3 1.2
    switch tracktype=grade2 switch isunpaved 10 3
    switch tracktype=grade3 10.0


quaelnix:
> Could you look into the other profiles touched too, e.g. whether anything breaks (trekking.brf)?

The changes in the `trekking.brf` probably did not break something, but they also do not work as one would expect, because the else branch where the `avoid_path` penalty was added is rarely reached on way segments that are of type `highway=path`.

> I'd also revert the addition of UI-facing profile options when they are a no-op 

Sounds reasonable.

quaelnix:
@afischerdev, here is an example route that shows that the new `avoid_path` logic in the `trekking.brf` basically does the opposite of what it is supposed to do: https://brouter.de/brouter-web/#map=12/50.8610/6.9499/standard,route-quality&lonlats=6.957874,50.942837;7.096699,50.732365

> I'd also revert the addition of UI-facing profile options when they are a no-op 

Done.

afischerdev:
Nice to see that we have new maintainers for the profiles.
Because I really don't feel like the supervisor for it.

As I mentioned in #561 I'm a friend of normalization. `fastbike-verylowtraffic` already use a `avoid_path`. And that is why I added it to the others as well. The input wasn't perfect at all. But it was good enough to start discussion.

And so I would like to see it holded in the profiles and optimized. And not thrown away.

BTW there was the constant expressions optimization #566 (PR #573). The `fastbike` profile had the `consider_traffic` logic before. So this  idea doesn't work.

quaelnix:
> As I mentioned in https://github.com/abrensch/brouter/issues/561 I'm a friend of normalization.

Me too, but we should first define what the goal is then discuss different alternatives and then apply them. There is no time pressure with this issue and the rushed `avoid_path` patch sends people across way segments with `surface=ground`, `smoothness=impassable` for no reason.

> so I would like to see it holded in the profiles and optimized.

The problem is that `avoid_unsafe` (e.g. avoid major roads) makes no sense on a profile like fastbike which actively searches for major roads, because you can drive fast there.

Same with `avoid_path` on the trekking profile. `highway=path` is usually the most used type of way on this profile.

> The `fastbike` profile had the `consider_traffic` logic before. So this idea doesn't work.

What do you mean by that?

afischerdev:
>>  The fastbike profile had the consider_traffic logic before. So this idea doesn't work.

> What do you mean by that?

The idea is not to have too much overhead in the routing result. When a constant is not used because of e.g. `consider_traffic = false` it is not need in message block when `processUnusedTags=false`. This does not work for [fastbike](https://github.com/abrensch/brouter/blob/11090210187f35f8283e8c2133ab6db9ae1c776f/misc/profiles2/fastbike.brf#L188).

> Same with avoid_path on the trekking profile. highway=path is usually the most used type of way on this profile.

May be, but what when I like to drive side by side (e.g. with kids), my trekking profile parameter are very optimized to my favours - only today no smaller ways.

> There is no time pressure

Yes, no time pressure, but there should be some pressure otherwise nothing happens. Some likes on an idea is not enough. Changes should follow.
Changing profiles is a thankless task. Everyone has different ideas. The goal here is finding the best way all can live with. And a switch more can add more comfort without changing profiles manually.

quaelnix:
> This does not work for [fastbike](https://github.com/abrensch/brouter/blob/11090210187f35f8283e8c2133ab6db9ae1c776f/misc/profiles2/fastbike.brf#L188).

Yes, but all we need to do to make it work is to also wrap the contents of the `assign trafficpenalty0 =` block into an `if consider_traffic then ... else ...` statement.

> but what when I like to drive side by side (e.g. with kids)

The most likely effect of `avoid_path` in the trekking profile is that you will instead be routed via `highway=track` or `highway=footway`, both of which are most likely worse than any paved `highway=path` in this case.

Just look at the example I posted above. Activation of `avoid_path` replaces a segment of:
```
highway=path surface=asphalt foot=designated bicycle=designated smoothness=excellent
```
with a segment of:
```
highway=track tracktype=grade2 surface=fine_gravel
```

> Some likes on an idea is not enough. Changes should follow.

It's not that I didn't try to find a way to optimize the default profiles according to https://github.com/abrensch/brouter/issues/561, I just failed.

> And a switch more can add more comfort without changing profiles manually.

I agree with that, but the switch must do what it is supposed to do, it must work hand in hand with *all* the other switches, and it must also not change behavior when the switch is not active. The change I want to revert fails in all three categories.

afischerdev:
Ok, it's up to you.

PR #595, opened 2023-07-18
Prevent breaks on future update on version number

afischerdev:
This update aims to avoid breaks in static functions when merging PR #457.

### DIFF
--- a/brouter-core/src/main/java/btools/router/OsmTrack.java
+++ b/brouter-core/src/main/java/btools/router/OsmTrack.java
@@ -839,7 +839,7 @@ public final class OsmTrack {
     return sb.toString();
   }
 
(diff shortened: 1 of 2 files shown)


afischerdev:
> For anyone wondering what this is about (given the sparse description and ambiguous title) and wanting to save some time scratching their head:

Please look after your words. 

There are always more options. 
But I agree your way is more elegant and I shouldn't have done anything.
So I can close this and let others play the ball.

afischerdev:
@rkflx 
I'm sorry, I missed your statement while closing.
The feedback is not the problem, support, quality assurance, improvements, ... are welcome and needed.
The thing is the style of communication, there are different ways of presentation.

PR #601, opened 2023-07-27
Fix bug in exclusion of small water bodies

quaelnix:
This [little pond](https://www.openstreetmap.org/way/124106504) was causing `estimated_river_class=2` on the [Hackländerstraße](https://www.openstreetmap.org/way/302539343):

estimated_river_class

---

It would probably also be worth discussing whether it would not make sense to raise the threshold back to something near the original value of around 8000 square meter:
https://github.com/abrensch/brouter/blob/e66468b091506833eb5890656a0ef3a538397ab2/misc/scripts/mapcreation/brouter.sql#L475

### DIFF
--- a/misc/scripts/mapcreation/brouter.sql
+++ b/misc/scripts/mapcreation/brouter.sql
@@ -77,7 +77,7 @@ FROM
     polygons p
 WHERE
     -- do not consider small surfaces
-    st_area (p.way) > 1000
+    st_area (st_transform (p.way, 4326)::geography) > 1000
     AND p.natural IN ('water')
     OR (p.landuse IN ('forest', 'allotments', 'flowerbed', 'orchard', 'vineyard', 'recreation_ground', 'village_green')
         OR p.leisure IN ('garden', 'park', 'nature_reserve'));
@@ -93,7 +93,7 @@ FROM
     polygons p
 WHERE
     -- do not consider small surfaces
-    st_area (p.way) > 1000
+    st_area (st_transform (p.way, 4326)::geography) > 1000
     AND p.natural IN ('water')
     OR (p.landuse IN ('forest', 'allotments', 'flowerbed', 'orchard', 'vineyard', 'recreation_ground', 'village_green')
         OR p.leisure IN ('garden', 'park', 'nature_reserve'));


afischerdev:
Thanks for your fix.
But please hang on a moment, @EssBee59 will not comment at the moment.
And I have the task of checking in the latest changes.

afischerdev:
@quaelnix 
I don't mind to do that. But I thought maybe @EssBee59  would like to add something else to that.

EssBee59:
It's ok to have different opinions, but I don't understand to create an issue with the label "bug" directly out of it!

I prefer to keep 1000 qm, else a lot of nice water places wouls not be considered, and this would probably lead to the report of new bugs.
Examples:

case1
case2

quaelnix:
> I prefer to keep 1000 qm

This PR fixes the bug that the threshold is currently ~ 1000 square feet instead of 1000 square meters.

quaelnix:
Another example. This [little basin](https://www.openstreetmap.org/way/1095739320) (less than 300 m² in size) is causing `estimated_river_class=4` on [Wiesenstraße](https://www.openstreetmap.org/way/6898220):

afischerdev:
> This PR fixes the bug that the threshold is currently ~ 1000 square feet instead of 1000 square meters.

I'm not sure about this.
A sql like this
`select p.osm_id, p.water, st_area(p.way),  st_area(st_transform (p.way, 4326)::geography) area_4326, st_area(st_transform (p.way, 4326)::geography) / 0.3048 ^ 2 sqft_4326  from polygons p where p.water is not null limit 10;`
brings out e.g this [pont](https://www.openstreetmap.org/way/833830724)  -  [controlled by this](https://www.mapdevelopers.com/area_finder.php?polygons=%5B%5B%5B%5B51.41808394159251%2C8.943861707913259%5D%2C%5B51.41801017345672%2C8.943703175905377%5D%2C%5B51.417972537320395%2C8.94368574154678%5D%2C%5B51.41792737391586%2C8.94373804462257%5D%2C%5B51.418024391544634%2C8.943940551403195%5D%2C%5B51.41808394159251%2C8.943861707913259%5D%5D%2C%22%23AAAAAA%22%2C%22%23000000%22%2C0.4%5D%5D) - switch to satellite view

| osm_id | water | st_area | area_4326 | sqft_4326 |
| :--- | ---| --- | --- | -- |
| 833830724  | pond  |  395.8 |  154.1 | 1659.6 |

Seems the area_4326 is the correct value for sqr meter.

quaelnix:
@afischerdev, yes, I got the square feet wrong, but that doesn't change the fact that the current code never gives a threshold of 1000 square feet unless you are at the aquator.

Essbee59 wants a 1000 square meter threshold, I'm fine with a 1000 square meter threshold and this pull request does get us 1000 square meter threshold, so please let us merge it.

quaelnix:
@afischerdev, yes, I got the square feet wrong, but that doesn't change the fact that the current code never gives a threshold of 1000 square meter unless you are at the equator.

Essbee59 wants a 1000 square meter threshold, I'm fine with a 1000 square meter threshold and this pull request does get us 1000 square meter threshold, so please let us merge it.

EssBee59:
Yes, 1000 is at the equator!
Historically the first calculation were done for Europe only...
Yes I  can / will next change that in the same way as I did by nature_reserve (in Greenland, there was a real need for that)

It seems extremly important for you to be "exact" by a parameter that can just be set with feeling / testing...
I would prefer to concentrate my energy on really critical points (as https://github.com/abrensch/brouter/issues/622)
Regards

quaelnix:
> that can just be set with feeling / testing...

Yes, but the intention was 1000 square meters and not < 300 square meters. That's the whole point of this pull request.

> I would prefer to concentrate my energy on really critical points

Me too, see: https://github.com/abrensch/brouter/pull/614

Issue #55, opened 2016-02-16
Include street names in rd5 files to make voice guidance better in OsmAnd 


RURon:
Hello, 
wonderful piece of software! So far I seem to have found a way to actually use a software for road bike routing! 

Back to topic. I use OsmAnd and use it for voice guidance most of the time. This would be so much better, if the street names were also spoken. I read that this is due to missing road names in the rd5 database. 
Is there any chance that we might have a db that includes those,  preferably as 2nd choice so that everyone who does not need them can save the space? 
Best regards and greetings from Düsseldorf! 
Rodja A.

usrusr:
In addition to better guidance output, this could also be the basis for a future extension to the cost model that offers an alternative to the existing turncost rules that is not based on curvature but on identity.

Currently, brouter tends to select undesirable shortcuts when there is a "direttissima" alternative to a gentle switchback climb or descent and (as far as I understand the code) no amount of turn cost can change that. A cost rule for road identity change could help alleviate that, but as long as the graph does not contain an approximation of identity (uniquity is not required, names would be fine) any experimentations towards that goal would be too big of a changeset.

Other use cases for "identity turncost" would be routing for group rides (where ease of navigation is a primary concern as every missed turn is a crash hazard and any ambiguity during long climbs or descents is problematic) and perhaps also whatever was the root goal behind issue #77 (negative turn cost).

0709wiwiwi:
The following may sound rude, but don't consider it so pse.
No street name support? 
Is this because it is not technically possible or is there a lack of motivation (trigger) to develop and offer it? 

https://forum.kurviger.de/t/no-street-names-in-android-app-navigation-brouter/5086/3?

"**Record street names in rd5 files to make navigation even more nice in any navigation app that uses the Brouter engine app**"

poutnikl:
@0709wiwiwi   Another topic level is API(application programming interface) support for the names. As, the preferred method of LocusMap-BRouter communication is API. 

But if street names were implemented in RD5 and if the support was implemented in LocusMap Gold, which has its own BRouter integrated.........   As LocusMap gold uses AFAIK its own RD5 generation and deployment, I guess it could implement street names into its own RD5 files. But, it would be lot of work and the best if coordinated with BRouter development.

zod:
As you already discovered the segment data (rd5) doesn't contain street names by design (https://github.com/abrensch/brouter/issues/39#issuecomment-119665274).

[...]

PR #602, opened 2023-07-27
No needless dropping of 'estimated_town_class'

quaelnix:
To enable the crossing of a city along rivers, only one line of code in the profile text needs to be adjusted:
```
switch estimated_town_class= 0
```
->
```
switch or estimated_town_class= ( not estimated_river_class= )  0
```

---

This significantly improves the usability of the `estimated_town_class` pseudo tag.

### DIFF
--- a/misc/scripts/mapcreation/brouter.sql
+++ b/misc/scripts/mapcreation/brouter.sql
@@ -675,15 +675,6 @@ WHERE losmid IN (
         WHERE
             forest_class NOT IN ('1'));
 
-DELETE FROM town_tags
-WHERE losmid IN (
-        SELECT
-            losmid
-        FROM
-            river_tags
-        WHERE
-            river_class NOT IN ('1'));
-
 SELECT
     count(*)
 FROM


quaelnix:
@afischerdev, have you thought about it yet?

afischerdev:
As before, I don't mind to do that. But I thought maybe @EssBee59  would like to add something else to that.

afischerdev:
@EssBee59 
What do you think about this?
Or just merge?

EssBee59:
Hello,
As every body is responsable for his profile, I will not comment this.
I am very satistfied with my profiles (without the proposed change, that means river_class=1 is not clearring the town penalty).
With the new tags a lost of parameters had to be set by "feeling", it would cost months of work to discuss each one.
As allready said, I would like to concentrate/ work on real bugs, not on own desires.

quaelnix:
> the proposed change, that means river_class=1 is not clearring the town penalty

The proposed change does not "not clear the town penalty". It only shifts the place where the mingling between the two tags is happening from the preprocessing into to the profile, which increases the usability of the new tags because we no longer enforce one preference (namely your preference) upon users.

> As every body is responsable for his profile

Yes, but without this change users simply cannot decide if they want to intermix the river and town class or not!

> I am very satistfied with my profiles

So what is the problem if you can simply replace `estimated_town_class=` with `or estimated_town_class= estimated_river_class=2|3|4|5|6`?

> I would like to concentrate/ work on real bugs, not on own desires.

You completely ignore that the proposed change increases the freedom of choice for profile developers.

quaelnix:
@afischerdev, I'm not going to accept @EssBee59 answer.

We currently spend a lot of computational time figuring out which way is within a city, only to later discard some of that information.

This pull request is not about "own desires" but about reasonable design decisions and throwing this valuable information away for no good reason is unreasonable.

Especially when [avoiding cities](https://github.com/abrensch/brouter/issues/608) is likely going to be on if not the biggest use-case of the new tags.

afischerdev:
@quaelnix 
At the end I guess it could be you have to accept. @EssBee59 is the maintainer of the database scripts and I would not commit without an OK.

But back to the facts:
What I understand is
- you want to keep the town data in database and do the exclusion river/town in profile
- the other way is to drop town tag from database when river is available

The arguments are understandable.
- users should have more options
- database should be as small as possible
We are living always in this conflict.

The sample issue #608 is not perfect for this discussion:
- it uses car profile which have no new 'green' tags at the moment
- it works in an area where the towns are small and so they don't have a town tag
- but yes, when we break the 50000 population rule - is not the topic at the moment -, the town tag for this villages will be gone

@quaelnix 
But the #608 shows you have other ideas on town tag generation. Please tell us more about 'building density' for town class. 
The idea using maxspeed and town class is interesting. But I didn't understand it. I know lots of streets with maxspeed 50 and lots of building, but no one slows down. And wouldn't effect it the time calculation?

devemux86:
> > database should be as small as possible

> Again, it makes no sense to drop valuable information because of space concern.

Usually some of the first things users expect from routers and they don't find in BRouter
are street names (#55), speed limits, maybe street information in response (e.g. tunnels).

So additional data in the database that will make users' lives easier,
it would be good to consider. :slightly_smiling_face:

quaelnix:
> Please tell us more about 'building density' for town class.

There is a branch on my BRouter fork that generates town tags based on building density: https://github.com/abrensch/brouter/compare/master...quaelnix:brouter:improve-town-and-forest-tags

And there are a couple of comments that contain examples here: https://github.com/abrensch/brouter/issues/486#issuecomment-1656730847

But aside from the considerable performance impact this will also not work properly in countries with poor OSM data, which is why I did not follow through on that concept. My best concept for the town tag so far can be found here: https://github.com/abrensch/brouter/pull/614

I would love to discuss the pros and cons of the different approaches and find the best solution as a team, but there doesn't seem to be much interest in taking the new pseudo tags to the next level. Which is a bit sad to be honest.

afischerdev:
@devemux86 
> Usually some of the first things users expect from routers and they don't find in BRouter
are street names (https://github.com/abrensch/brouter/issues/55), speed limits, maybe street information in response (e.g. tunnels).

I see the street name problem, but I doesn't see the others.
Take this [sample](http://brouter.de/brouter-web/#map=14/47.8445/13.7761/osm-mapnik-german_style&lonlats=13.778834,47.852897;13.785447,47.833491&profile=car-eco) and make all tags visible. You will get maxspeed and tunnel in the json output.

afischerdev:
@quaelnix 
Thanks for the hint. I tried it and it really takes a long time. I still think it's a good idea to do it over the buildings. Maybe the special idea of speeding things up will come along.

I also tried #614, it runs much faster. But it's broken down to 1000 residents per village. I can't say what that means to database. May be @EssBee59  have more experiences here. May be a whole world test could be done.

quaelnix:
> I still think it's a good idea to do it over the buildings.

But take a look at cities like [Sydney, Australia](https://www.openstreetmap.org/#map=12/-33.8674/151.0428), where the majority of buildings are still missing.

> Maybe the special idea of speeding things up will come along.

I'm pretty sure that a SQL expert would be able to speed up our sql code by a factor of 2 or more.

quaelnix:
Try this route with the `Fastbike (low traffic) profile: https://brouter.de/brouter-web/#map=11/50.1237/8.7139/standard&lonlats=8.582151,50.202148;8.863721,50.077662&nogos=8.675754,50.175634,3&profile=fastbike-lowtraffic

Then enable `consider_town` and look at the result, then disable `consider_town` and enable `consider_noise` look at the result and then also enable `consider_town` and again look at the result and you will understand why the needless dropping of the town class is plain stupid.

quaelnix:
Try this route with the `Fastbike (low traffic)` profile: https://brouter.de/brouter-web/#map=11/50.1237/8.7139/standard&lonlats=8.582151,50.202148;8.863721,50.077662&profile=fastbike-lowtraffic

Then toggle `consider_town` and compare the results, then enable `consider_noise` and once again toggle `consider_town` and compare the results and you will immediately understand why the needless dropping of the town class is plain stupid.

afischerdev:
@quaelnix 
Yes, no comfortable way around.

PR #616, opened 2023-08-20
Replace multiple occurrences of toLowerCase()

moving-bits:
Avoids recalculation of `toLowerCase()` for every branch in `addLookupValue()`

### DIFF
--- a/brouter-expressions/src/main/java/btools/expressions/BExpressionContext.java
+++ b/brouter-expressions/src/main/java/btools/expressions/BExpressionContext.java
@@ -578,7 +578,7 @@ public abstract class BExpressionContext implements IByteArrayUnifier {
             if (value.toLowerCase().contains("ft")) {
               float foot = 0f;
               int inch = 0;
-              String[] sa = value.toLowerCase().trim().split("ft");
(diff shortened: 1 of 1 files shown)


afischerdev:
Thank you for contributing

quaelnix:
> Thank you for contributing

Yes, but does this really improve the performance (and or readability) of the code?

moving-bits:
> > Thank you for contributing
> 
> Yes, but does this really improve the performance (and or readability) of the code?

Readability IMHO yes (but that's partly a matter of personal preference), and performance depends on which switch branch gets called.

afischerdev:
From my point of view I can't see much more readability here `valueLowerCase vs value.toLowerCase`.
The benefit I see is the single call of the function.
And may be an other one for the call with `Locale.US` - when this is using an ASCII and not an UTF16 variant.

But the longer I look at this code, I see another optimization:
- move the call value.toLowerCase a few lines up to the 'value prepare section'.
- stay with the name `value`
- remove the trim() calls - blanks have been escaped with '_' and that was removed in 'value prepare section'
- may be add a remove for non breaking space

I'll do that the next days.

PR #619, opened 2023-08-24
New BRouter command line with new parameter collector

afischerdev:
The old BRouter class was working with two positions only. Now it can be called with all available routing options like:

```
java -jar brouter.jar <segmentdir> <profiledir> <engineMode> <profile> <lonlats-list> [parameter-list] [profile-parameter-list] 
or
java -cp %CLASSPATH% btools.server.BRouter <segmentdir>> <profiledir> <engineMode> <profile> <lonlats-list> [parameter-list] [profile-parameter-list]
or
java -jar brouter.jar <segmentdir> <profiledir> <parameter-list>
```

The also included 'seed' variant is not changed and should run as before. 

The new BRouter class comes with a first version of a parameter collector. The idea behind is that BRouter has three variants of parameter collection: BRouter command line, Android app and server. The new collector class should work for all. At the end there is only one place for parameter.

At the moment the server and the Android app are not changed for this collector. Should be done in an other step. First the collector can be tested with the BRouter command line.

There are also two 'new' parameter in RoutingContext `outputFormat` and `exportWaypoints`. They have been there before - only at other places. And the use also takes place in a further step when the parameter handling is changed in the server and in the Android app.

### DIFF
--- a/brouter-core/src/main/java/btools/router/RoutingContext.java
+++ b/brouter-core/src/main/java/btools/router/RoutingContext.java
@@ -216,6 +216,9 @@ public final class RoutingContext {
   public boolean inverseRouting;
   public boolean showTime;
 
+  public String outputFormat = "gpx";
+  public boolean exportWaypoints = false;
+
   public OsmPrePath firstPrePath;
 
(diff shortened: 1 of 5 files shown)


quaelnix:
I like the idea of doing the parameter parsing in a single place.

PR #620, opened 2023-08-26
Add new tags and description profile racebike

afischerdev:
Here an update for the profile `fastbike-verylowtraffic`.
About new tags and an included description.


quaelnix:
This pull request increases the time needed to calculate a route (with default settings) by more than four times.

afischerdev:
@quaelnix 
Yes, you are right. I replaced the wrong file. `racebike-verylowtraffic` was the candidate.

quaelnix:
But isn't it a bit confusing when `fastbike-verylowtraffic.brf` and `racebike-verylowtraffic.brf` coexist?

afischerdev:
@EssBee59 
Could you please help us in this discussion?

EssBee59:
Hello!
As my profiles (racebike-verylowtraffic and trekking_SB) are only addressed in my test instance, I think it is not necessary to put or update them in the repository?
The current state:
I updated both profiles last month, had only few time to test but I did not found problems

afischerdev:
@EssBee59 
Thanks for the clearing.

PR #621, opened 2023-08-26
Replace multiple occurrences of toLowerCase() - continued

afischerdev:
Update with rework as discussed in #616:
removed trim, renamed variable back to `value`, added new parse character.

### DIFF
--- a/brouter-expressions/src/main/java/btools/expressions/BExpressionContext.java
+++ b/brouter-expressions/src/main/java/btools/expressions/BExpressionContext.java
@@ -565,85 +565,85 @@ public abstract class BExpressionContext implements IByteArrayUnifier {
           String org = value;
           try {
             // remove some unused characters
-            value = value.replace(",", ".");
-            value = value.replace(">", "");
-            value = value.replace("_", "");
+            value = value.replaceAll(",", ".");
+            value = value.replaceAll(">", "");
+            value = value.replaceAll("_", "");
+            value = value.replaceAll(" ", "");
+            //value = value.replaceAll(" ", ""); // nbsp
+            value = value.replaceAll("’", "'");
+            value = value.replaceAll("”", "\"");
             if (value.indexOf("-") == 0) value = value.substring(1);
-            if (value.indexOf("~") == 0) value = value.substring(1);
-            if (value.contains("-")) {    // replace eg. 1.4-1.6 m
-              String tmp = value.substring(value.indexOf("-") + 1).replaceAll("[0-9.,-]", "");
-              value = value.substring(0, value.indexOf("-")) + tmp;
+            if (value.contains("-")) {    // replace eg. 1.4-1.6 m but also 1'-6"
+              value = value.substring(0, value.indexOf("-")); // + tmp;
             }
+            value = value.toLowerCase(Locale.US);
+
             // do some value conversion
-            if (value.toLowerCase().contains("ft")) {
+            if (value.contains("ft")) {
               float foot = 0f;
               int inch = 0;
-              String[] sa = value.trim().toLowerCase().split("ft");
-              if (sa.length >= 1) foot = Float.parseFloat(sa[0].trim());
+              String[] sa = value.split("ft");
+              if (sa.length >= 1) foot = Float.parseFloat(sa[0]);
               if (sa.length == 2) {
                 value = sa[1];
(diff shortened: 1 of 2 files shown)


brouter-expressions/src/main/java/btools/expressions/BExpressionContext.java
```diff
@@ -565,85 +565,85 @@ public BExpressionLookupValue addLookupValue(String name, String value, int[] lo
...
-            } else if (valueLowerCase.contains("meter")) {
-              String s = value.substring(0, valueLowerCase.indexOf("m"));
-              value = s.trim();
-            } else if (valueLowerCase.contains("mph")) {
-              String[] sa = valueLowerCase.trim().split("mph");
-              if (sa.length >= 1) value = sa[0].trim();
-              float mph = Float.parseFloat(value.trim());
+            } else if (value.contains("meter")) {
+              value = value.substring(0, value.indexOf("m"));
+            } else if (value.toLowerCase().contains("mph")) {
```
quaelnix, line 630: Duplicate `toLowerCase()` call.

afischerdev: Yes, thank you, very considerate

quaelnix:
What about these errors in the build log?
```
BExpressionContext.java:572: error: unmappable character (0xA0) for encoding UTF-8

            //value = value.replaceAll("�", ""); // nbsp
                                        ^ 
BExpressionContext.java:573: error: unmappable character (0x92) for encoding UTF-8
            value = value.replaceAll("�", "'");
                                      ^ 
BExpressionContext.java:574: error: unmappable character (0x94) for encoding UTF-8
            value = value.replaceAll("�", "\"");
                                      ^
```

brouter-expressions/src/main/java/btools/expressions/BExpressionContext.java
```diff
@@ -565,85 +565,85 @@ public BExpressionLookupValue addLookupValue(String name, String value, int[] lo
...
+            //value = value.replaceAll("�", ""); // nbsp
+            value = value.replaceAll("�", "'");
+            value = value.replaceAll("�", "\"");
             if (value.indexOf("-") == 0) value = value.substring(1);
-            if (value.indexOf("~") == 0) value = value.substring(1);
-            if (value.contains("-")) {    // replace eg. 1.4-1.6 m
-              String tmp = value.substring(value.indexOf("-") + 1).replaceAll("[0-9.,-]", "");
-              value = value.substring(0, value.indexOf("-")) + tmp;
+            if (value.contains("-")) {    // replace eg. 1.4-1.6 m but also 1'-6"
+              value = value.substring(0, value.indexOf("-")); // + tmp;
```
quaelnix, line 577: Won't this fail on something like `30-35 mph` because it drops the unit which is needed for the later conversion?

quaelnix:
I think this is another one of those commits where there are just too many things happening at once. I spent 2 hours on this one and I still do not have a good idea of what consequences these changes might have.

My suggestion would be to split these changes as follows:
1. Commit: Everything related to the `toLowerCase()`  logic
2. Commit: Everything related to the `trim()` logic
3. Commit: Everything related to the `replace(...)` lobic
4. Commit: Change in the `-` logic

It is far too easy to introduce new bugs with changes like this if one is not very careful.

brouter-expressions/src/main/java/btools/expressions/BExpressionContext.java
afischerdev: Yes, in the old version `2m-3m` returns as 2mm. So I drop this.
But anyway this is not a good solution for all. I have a version with numeric check before the append.

afischerdev: >  errors in the build log

Interesting thing. No, I haven't seen. The build was 'green' so I didn't look inside the action report. 
On Windows I didn't get the errors and the rd5 generation error log shows the expected results.
But I checked the lib included in ZIP file (from last action). This is not usable - in this case.

quaelnix: > `2m-3m` returns as 2mm

Good catch! I had not thought of this case at all.

quaelnix:
:+1:

Issue #613, opened 2023-08-16
Rework on voicehint process


afischerdev:
It seems the post process for voicehint generation needs a rework.
Please see [doc for voicehint](https://github.com/abrensch/brouter/blob/8d123e337516664f1bbd05c0802aab30ec698726/docs/features/voicehints.md?plain=1#L21) for the current rules.

Related to this [discussion](https://github.com/devemux86/cruiser/discussions/164#discussioncomment-6738015) in Cruiser forum we have some smaller irritation for the users (I hope I got it correct):

| Sample | Result | Expected Result |
| :--- | :--- | :--- |
| [1. sample](https://brouter.de/brouter-web/#map=18/52.23631/21.08689/standard&lonlats=21.084926,52.235957;21.086953,52.236789&profile=car-eco) | left, continue | left |
| [2. sample](https://brouter.de/brouter-web/#map=18/52.23665/21.08697/standard&lonlats=21.086803,52.23691;21.087645,52.236062&profile=car-eco) | continue, left | left |
| 1. sample with [trekking](https://brouter.de/brouter-web/#map=18/52.23631/21.08689/standard&lonlats=21.084926,52.235957;21.086953,52.236789) | continue, left, continue, continue | left |
| [3. sample u-turn](https://brouter.de/brouter-web/#map=18/52.23622/21.08662/standard&lonlats=21.086803,52.23691;21.086959,52.236654&profile=car-eco) | continue, u-turn-left, continue | u-turn-left |
| [4. sample u-turn](https://brouter.de/brouter-web/#map=18/52.23622/21.08662/standard&lonlats=21.087704,52.23622;21.087656,52.236056&profile=car-eco) | u-turn-left | u-turn-left |

The rule `keep continue when bad way is on higher prio` is working fine. But it has its side effects.

Are there any other examples that should be revised as well?

afischerdev:
An other point for this rework is the handling of two near turn instructions left/right.
Here two examples
[5 meter difference](http://brouter.de/brouter-web/#map=18/49.97455/8.56916/osm-mapnik-german_style&lonlats=8.567633,49.974327;8.569221,49.974934&profile=car-eco) - result for all straight
[7 meter distance](http://brouter.de/brouter-web/#map=19/51.06202/3.94846/standard&lonlats=3.945481,51.062231;3.948759,51.063007) - result for trekking right/left, for car straight

[...]

PR #634, opened 2023-10-05
Update new parameter collector for BRouter app

afischerdev:
Use the new parameter collector also in app context and remove old param logic and routines.


quaelnix:
Lots of duplicate code gone for good! :+1:

... however, I'm a little afraid that this might break something, though I haven't found anything yet.

afischerdev:
@quaelnix 
> however, I'm a little afraid that this might break something, though I haven't found anything yet.

Indeed it has the potential to run into problems.

My ideas on this situation:
- finish the current param collector for BRouter web service
- add a new output logic (move from OsmTrack to extra classes) - this also carries a risk in it

In my opinion, the server is not critical. It could change the library very quickly if something doesn't work properly. And these kind of changes can be easily tested by comparing the results - there should be no change. But for the app an update on trouble will take some days. And this is not easy on testing.

It could be a good way to use a beta app via the Play Store to have more test with users. But we can't publish an update without any user relevant changes.
So I would like to add changes on voicehint #613 as well.

quaelnix:
> add a new output logic (move from OsmTrack to extra classes) - this also carries a risk in it

If it improves the readability of the code, I'm in.

> In my opinion, the server is not critical. It could change the library very quickly if something doesn't work properly.

I think we should try to break neither of the two. We can't know who is using the server in what way.

afischerdev:
> If it improves the readability of the code, I'm in.

It only clears the OsmTrack from formatting and move it to Gpx, Kml, Json, Cvs classes. But code is as it was before. Map matching will need 5x5 OsmTracks to compare each point with next.

> I think we should try to break neither of the two. 

Yes, of course, but how do you achieve this goal?

afischerdev:
@quaelnix 
> I think we should try to break neither of the two.

Is there a new idea to protect the server lib or/and the app from breaking?

zod:
Getting rid of duplicated code is always desirable because it's less code to maintain :) Thanks for your effort!

The usual approach to ensure refactoring doesn't break functionality is to write tests before refactoring which represent the desired behavior. Unfortunately most BRouter classes can't be tested in isolation because of close-coupling with other classes which makes writing unit tests really hard. Usually coverage information is also used to have a metric/information which parts of the code are covered by tests.

afischerdev:
@zod 
My extra tests are no unit tests.
I use the server and fire around 80 requests for old lib against new lib. And compare the results.
For the parameter and the new output logic the tests are easy. The result must be the same. That covers the most features we have. 
Some day I like to do that with an additional Android app as well.

quaelnix:
> Is there a new idea to protect the server lib or/and the app from breaking?

No, and I also do not understand this part of the code well enough to be of any help in this matter. Sorry.

Issue #622, opened 2023-08-28
Trekking profile and cycle-routes


EssBee59:
It seems, the standard trekking profile do not calculate the right costs when a cycle route is defined on a way

2 examples:

http://brouter.de/brouter-web/#map=14/49.9399/8.9582/osm-mapnik-german_style&lonlats=8.954887,49.953097;8.924436,49.887188

http://brouter.de/brouter-web/#map=13/49.9767/8.5679/osm-mapnik-german_style&lonlats=8.578911,49.990673;8.451819,49.947091

[bug_trekking.txt](https://github.com/abrensch/brouter/files/12452816/bug_trekking.txt)

(see cost of 1.000 / km for secondary and tertiary...)

quaelnix:
I would call this a feature and not a bug:
https://github.com/abrensch/brouter/blob/38fc780055e789ee1beb85b5e4439be7a38d8adc/misc/profiles2/trekking.brf#L279-L282
and if you dislike this behavior you can simply toggle the `ignore_cycleroutes` switch?

quaelnix:
The comment in the code says:

> always treated as perfect (=1) 

and the code does do that.

> this is too dangerous

Flip the `avoid_unsafe` switch?

> a mistake by the handling cycle_route possibly exists ?

I don't think that there is a good solution for the dilemma that you cannot make a profile that works perfectly world wide, see the examples in the first post of this thread: https://github.com/abrensch/brouter/issues/479

quaelnix:
> possibly defined for race-bikers who used that way once at a week

- The route relation: https://www.openstreetmap.org/relation/299353#map=13/49.9333/8.9693
- The official description: https://www.rmv.de/c/de/service/freizeitangebote/rhein-main-vergnuegen/rhein-main-vergnuegen-route-5-unterwegs-im-dieburger-land

> Bis auf wenige Ausnahmen verläuft sie getrennt vom Autoverkehr.

> With a few exceptions, it runs separately from car traffic.

It seems obvious that the route relation in OSM is simply wrong.

quaelnix:
The official map clearly shows that the route is not supposed to be on the L 3065 between Babenhausen and Langstadt!

EssBee59:
Of course, errors in OSM are possible.
But I think this is not a unique case.
See quick search with overpass (not giving a perfect result, but helpful)

rel({{bbox}})["route"="bicycle"]->.rels;
[...]

Issue #610, opened 2023-08-10
Adding penalties destroys ratios between roads, so multiplying should be used.


Stapawe:
Let's assume that with (un)paved paths you want, or take your time, to cycle twice as long on an asphalt path than an unpaved path and you have two routes between A and B
- 1.1 km unpaved path, with a cost factor 2 the total cost is 2.2 k, and 
- 2 km paved path with a cost factor 1, in total 2 k so you cycle on the asphalt
But both paths are along noisy roads with added penalty 2.
So:
- 1.1 km unpaved path has a total cost factor 4, with a total cost of 4.4
- 2 km paved path with a cost factor 3, in total 6 k. 
So noise on both routes changes your route to cycle longer time in noise on the unpaved path.

If you say you want to trade off the noise for twice time longer route, then the cost factor should increase twice time, but addition decreases the influence of the penalty with higher starting cost factor.
With multiplying it would be:
- 1.1 km unpaved path, with a cost factor 2*2 the total cost is 4.4 k, and 
- 2 km paved path with a cost factor 1*2, in total 2 4 k so you cycle on the asphalt

I'm impressed by the recent addition of noise and forest penalties and so on. Thanks und danke

quaelnix:
@Stapawe, here is a profile that does it like that: https://github.com/quaelnix/brouter-profiles-db/blob/main/gravel.brf

I haven't had time to try this concept out on the standard profiles. But I agree that it would be worthwhile.

EssBee59:
> * So noise on both routes changes your route to CYCLE LONGER time in noise on the unpaved path.

Hello Stapawe,

Thank for your interesting example!
Assuming, the cost factor reflects lineary the speed of the biker, then is your calculation right!
But, is this allways the case?
As example, replace above the "unpaved path" with a primary highway, and the "paved path" with a track  highway 
 + tracktype=1..
(A profile for racebike + very low traffic would set the basic costfactor in a similar way!)

In that case, considering the noise  I would prefer the primary highway, because the speed on primary is higher as on track.

I think, penalties of each kind (example noise") can change (destroy is not the right term) the ration between roads (in the same way as surface, highway type etc...) ?

Last week I also discovered "side  effects" by adding costs on the costfactor
https://github.com/abrensch/brouter/pull/579#issuecomment-1674315124

A better solution as adding is perhaps possible, I think currently each profile should decides.

Regards and thank for your comment about the new tags, it was a hard job to introduce that!
Ess Bee

[...]

PR #641, opened 2023-10-28
Update the trekking profile

afischerdev:
Here a new version for the trekking profile  suggested by @EssBee59

### DIFF
--- a/brouter-core/src/test/java/btools/router/RoutingEngineTest.java
+++ b/brouter-core/src/test/java/btools/router/RoutingEngineTest.java
@@ -47,7 +47,7 @@ public class RoutingEngineTest {
     String msg = calcRoute(8.723037, 50.000491, 8.712737, 50.002899, "paramTrack", rctx);
     Assert.assertNull("routing failed: " + msg, msg);
 
-    File trackFile = new File(workingDir, "paramTrack1.gpx");
+    File trackFile = new File(workingDir, "paramTrack0.gpx");
     trackFile.deleteOnExit();
     Assert.assertTrue("result content mismatch", trackFile.exists());
   }
--- a/brouter-core/src/test/java/btools/router/RoutingEngineTest.java
+++ b/brouter-core/src/test/java/btools/router/RoutingEngineTest.java
@@ -47,7 +47,7 @@ public class RoutingEngineTest {
     String msg = calcRoute(8.723037, 50.000491, 8.712737, 50.002899, "paramTrack", rctx);
     Assert.assertNull("routing failed: " + msg, msg);
 
-    File trackFile = new File(workingDir, "paramTrack0.gpx");
+    File trackFile = new File(workingDir, "paramTrack1.gpx");
     trackFile.deleteOnExit();
     Assert.assertTrue("result content mismatch", trackFile.exists());
   }
--- a/brouter-core/src/test/resources/paramTrack0.gpx
+++ b/brouter-core/src/test/resources/paramTrack0.gpx
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- track-length = 1570 filtered ascend = 4 plain-ascend = -15 cost=2840 energy=.0kwh time=3m 45s -->
+<!-- track-length = 1497 filtered ascend = 4 plain-ascend = -15 cost=2979 energy=.0kwh time=3m 35s -->
 <gpx 
  xmlns="http://www.topografix.com/GPX/1/1" 
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
  xsi:schemaLocation="http://www.topografix.com/GPX/1/1 http://www.topografix.com/GPX/1/1/gpx.xsd" 
- creator="BRouter-1.7.0" version="1.1">
+ creator="BRouter-1.7.3" version="1.1">
  <trk>
-  <name>brouter_trekking_0</name>
+  <name>brouter_test/../../../../misc/profiles2/trekking_0</name>
   <trkseg>
    <trkpt lon="8.723027" lat="50.000499"><ele>175.25</ele></trkpt>
    <trkpt lon="8.723285" lat="50.000610"><ele>176.75</ele></trkpt>
@@ -16,15 +16,12 @@
    <trkpt lon="8.722781" lat="50.001312"><ele>173.25</ele></trkpt>
    <trkpt lon="8.722027" lat="50.001834"><ele>169.5</ele></trkpt>
    <trkpt lon="8.721982" lat="50.001865"><ele>169.5</ele></trkpt>
-   <trkpt lon="8.722320" lat="50.002050"><ele>171.0</ele></trkpt>
-   <trkpt lon="8.722449" lat="50.002197"><ele>171.25</ele></trkpt>
-   <trkpt lon="8.722497" lat="50.002337"><ele>171.25</ele></trkpt>
-   <trkpt lon="8.722506" lat="50.002463"><ele>171.0</ele></trkpt>
-   <trkpt lon="8.722486" lat="50.002697"><ele>170.25</ele></trkpt>
-   <trkpt lon="8.721892" lat="50.002631"><ele>167.75</ele></trkpt>
-   <trkpt lon="8.721836" lat="50.002624"><ele>167.5</ele></trkpt>
-   <trkpt lon="8.721209" lat="50.002553"><ele>165.25</ele></trkpt>
-   <trkpt lon="8.721118" lat="50.002538"><ele>164.75</ele></trkpt>
+   <trkpt lon="8.721359" lat="50.002265"><ele>166.25</ele></trkpt>
+   <trkpt lon="8.721230" lat="50.002342"><ele>165.5</ele></trkpt>
+   <trkpt lon="8.721155" lat="50.002368"><ele>165.0</ele></trkpt>
+   <trkpt lon="8.721041" lat="50.002409"><ele>164.5</ele></trkpt>
+   <trkpt lon="8.721050" lat="50.002430"><ele>164.5</ele></trkpt>
+   <trkpt lon="8.721050" lat="50.002453"><ele>164.5</ele></trkpt>
    <trkpt lon="8.721021" lat="50.002493"><ele>164.5</ele></trkpt>
    <trkpt lon="8.720994" lat="50.002509"><ele>164.25</ele></trkpt>
    <trkpt lon="8.720960" lat="50.002518"><ele>164.25</ele></trkpt>
@@ -51,13 +48,12 @@
    <trkpt lon="8.712848" lat="50.004896"><ele>153.75</ele></trkpt>
    <trkpt lon="8.712781" lat="50.004859"><ele>154.0</ele></trkpt>
    <trkpt lon="8.712667" lat="50.004765"><ele>154.25</ele></trkpt>
-   <trkpt lon="8.712563" lat="50.004683"><ele>154.5</ele></trkpt>
-   <trkpt lon="8.712154" lat="50.004321"><ele>156.25</ele></trkpt>
-   <trkpt lon="8.712066" lat="50.004245"><ele>156.5</ele></trkpt>
-   <trkpt lon="8.713422" lat="50.003599"><ele>158.5</ele></trkpt>
-   <trkpt lon="8.713452" lat="50.003572"><ele>158.5</ele></trkpt>
-   <trkpt lon="8.713592" lat="50.003347"><ele>158.5</ele></trkpt>
-   <trkpt lon="8.713620" lat="50.003326"><ele>158.5</ele></trkpt>
+   <trkpt lon="8.712845" lat="50.004679"><ele>154.5</ele></trkpt>
+   <trkpt lon="8.714386" lat="50.004209"><ele>156.0</ele></trkpt>
+   <trkpt lon="8.714921" lat="50.003921"><ele>157.0</ele></trkpt>
+   <trkpt lon="8.714950" lat="50.003881"><ele>157.0</ele></trkpt>
+   <trkpt lon="8.714935" lat="50.003833"><ele>157.25</ele></trkpt>
+   <trkpt lon="8.714560" lat="50.003573"><ele>158.0</ele></trkpt>
    <trkpt lon="8.713956" lat="50.003142"><ele>158.5</ele></trkpt>
    <trkpt lon="8.713468" lat="50.002781"><ele>159.5</ele></trkpt>
    <trkpt lon="8.713293" lat="50.002684"><ele>159.75</ele></trkpt>
--- a/brouter-expressions/src/main/java/btools/expressions/BExpression.java
+++ b/brouter-expressions/src/main/java/btools/expressions/BExpression.java
@@ -352,6 +352,10 @@ final class BExpression {
     return this;
   }
 
+  public int getExpressionIndex() {
+    return variableIdx;
+  }
+
   private float max(float v1, float v2) {
     return v1 > v2 ? v1 : v2;
   }
--- a/brouter-expressions/src/main/java/btools/expressions/BExpressionContext.java
+++ b/brouter-expressions/src/main/java/btools/expressions/BExpressionContext.java
@@ -13,11 +13,11 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Random;
 import java.util.StringTokenizer;
 import java.util.TreeMap;
-import java.util.Locale;
 
 import btools.util.BitCoderContext;
 import btools.util.Crc32;
@@ -883,6 +883,20 @@ public abstract class BExpressionContext implements IByteArrayUnifier {
       }
     }
     return num.intValue();
+  }
+
+  public String getVariableKeyValue() {
+    StringBuilder sb = new StringBuilder(200);
+
+    int n = expressionList.size();
+    for (int expidx = 0; expidx < n; expidx++) {
+      float bx = expressionList.get(expidx).evaluate(this);
+      int varidx = expressionList.get(expidx).getExpressionIndex();
+      if (sb.length() > 0) sb.append(' ');
+      sb.append(variableName(varidx)).append("=").append(bx);
+    }
+
+    return sb.toString();
   }
 
   int getMinWriteIdx() {
--- a/brouter-expressions/src/test/java/btools/expressions/EncodeDecodeTest.java
+++ b/brouter-expressions/src/test/java/btools/expressions/EncodeDecodeTest.java
@@ -52,6 +52,6 @@ public class EncodeDecodeTest {
     System.out.println("description: " + expctxWay.getKeyValueDescription(true, description));
 
     float costfactor = expctxWay.getCostfactor();
-    Assert.assertTrue("costfactor mismatch", Math.abs(costfactor - 5.15) < 0.00001);
+    Assert.assertTrue("costfactor mismatch", Math.abs(costfactor - 5.1) < 0.00001);
   }
 }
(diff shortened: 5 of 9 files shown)


zod:
It seems like the updated profile does no longer have a `avoid_unsafe` parameter. The fix for the `overrideParam` testcase should be a valid parameter (which produces a different result) instead of changing the file.

afischerdev:
@zod 
thanks for your review. The trekking profile has an `avoid_unsafe`, but it is located a little bit down.
1. override param in 'RoutingEngineTest' paramTrack comes out as paramTrack0 (first alternative, can't say why the first test was against paramTrack1, don't see a second call)
2. override param in 'RouteServerTest'  also works for me ('avoid_unsafe=0' brings a different results)

afischerdev:
@zod 
I made new tests to get this. 
First I found no paramTrack0.gpx in my local repo. Must have dropped it  But it is needed to build the paramTrack1.gpx
Then I generated the route with old and new profile.

test_alt
test_neu

First one is the old situation - blue the unsafe route, red the safe result and green the test route from repo
Second one is new - blue unsafe, red the safe route.
As you can see the new safe route is near to the test route. That seems to be the reason that there is no paramTrack1.gpx as result

What to do:
I would like to replace the old test route0 with the unsafe route. Both profiles produce it in the same way.
What do you think?

quaelnix:
This PR adds the following without comment:
- `railwaypenalty` (which will affect routing)
- `barrierpenalty` (which will affect routing)
- `uphillcost` penalty of 80 (which will affect routing)
- `consider_traffic` logic that does not fit a trekking profile
- `surface=compacted` to `isunpaved` (which will affect routing)
- `downhillcutoff` and `upillcutoff` logic that avoids steep inclines
- `avoid_path` we previously agreed on to not add: https://github.com/abrensch/brouter/pull/593
- `smoothnesspenalty` we previously agreed on to not add: https://github.com/abrensch/brouter/pull/58

removes the following without comment:
- `stick_to_cycleroutes` logic

changes the following without comment:
- `ignore_cycleroutes` from `false` to `true`
- `tracktype` penalty weights 

And introduces a new logic for subtracting costfactors that can result in the costfactor becoming less than 1, which not only has the potential to completely destroy routing, but also contradicts what is currently being discussed here: https://github.com/abrensch/brouter/issues/610.

---

If this is merged, I will end the collaboration.

EssBee59:
Hello !
Yes, I suggested this profile version to replace the current "trekking", the PR is a kind of a misunderstanding with afischerdev.
Anyway, (perhaps better so!) we have now first reactions to the proposed changes, so that a real discussion on the parameters could start?

My first goal was  to change the default value for "ignore_cycleroute":
see https://github.com/abrensch/brouter/issues/622  
the further changes are not so critical, but as example not considering surface=compacted in the same way as fine_gravel is for me a severe bug.
Yes, I accept the critic that no documentation was provided, thank Quelnix for your list!

I suggest to start (or better to continue) the discussion with https://github.com/abrensch/brouter/issues/622
Till now I am missing arguments to not change the default value?

I hope we can prepare together a solution to this first parameter and go on with the others.   
Regards

afischerdev:
@zod 
Added a new paramTrack file and went back to old test. But for me it is still  more  a test for the generation of a 2nd route then a test for a param change.

afischerdev:
@quaelnix 
You're right, a detailed explanation would have been good. 
When I started I was still unsure whether I should create an issue or a PR. But I decided to do a PR because we already have enough issues about it that end without a result.
But I didn't ask for any further information about it. Sorry.

From my point of view, it would be good if we have a person responsible for the profiles who would bring the issues together. That's why I had the feeling that no one felt responsible for it and nothing was happening. I think it was good that @EssBee59  had the courage to create a new version of the trekking profile.

With a maintainer we could also go back to the normal way, create an issue and a PR for a result.

> If this is merged, I will end the collaboration.

Hard words, but understandable.
I would miss you. We need a corrective, someone who will point out the critical points and encourage changes. But I know that is not always easy going.

EssBee59:
> With a maintainer we could also go back to the normal way, create an issue and a PR for a result.


Yes, with a maintainer defect #622 would probably not hang for weeks!!!


It had be better to start with that, but now I found time to document the changes suggested in this PR profile:



1-cycle_route
===========
Not every  part of a cycle-route is a "safe harbor": 
Constrains exist to create a cycle-route, as example
- it should start and end in the center of a town
- big groupes of cyclist are expected, so that it has to avoid narrow pathes or cycleways!

Statistics from Hesse-Germany show as example, that a big part of the OSM primary (10%) and secondary (20%) ways are used in these cycle-routes, most of them without any cycleway=lane!

In the trekking profile (with the default parameters) these ways get a lower cost as any other ways (track, cycleway etc...), so that the resulting routing is not safe and do not reflect what the Brouter can do!


2-surface=compacted
=================
Fine_gravel ==> cost 1
compacted   ==> cost 2

But compacted (quality just below asphalt!) is better than fine_gravel

The profile make a very limited usage of the surface tag
... assign ispaved = surface=paved|asphalt|concrete|paving_stones|sett
... assign isunpaved = not or surface= or ispaved surface=fine_gravel|cobblestone 

3-uphillcost penalty
===================
The current implementation in the trekking:

assign downhillcost       = 60    # %downhillcost% | Cost for going downhill | number
assign downhillcutoff     = 1.5   # %downhillcutoff% | Gradients below this value in percents are not counted. | number
assign uphillcost         = 0     # %uphillcost% | Cost for going uphill | number
assign uphillcutoff       = 1.5   # %uphillcutoff% | Gradients below this value in percents are not counted.  | number

assign downhillcost       = if consider_elevation then downhillcost else 0
assign uphillcost         = if consider_elevation then uphillcost else 0

With the default value "consider_elevation=true", uphillcost remains =0, and this is a very poor usage of the "elevation" fonction as supported by the Brouter.

example
2 routes are possible to the same destination:

Route A:
20 km  with uphill factor of 7.5 
+ 100 km with a downhill factor of 1.5 

Alternativ Route B:
125 Km with uphill=downhill=0

The user would certainly prefer "Route B" as he choosed "consider_elevation".... but Brouter will suggest "Route A" because no elevation costs are added during his calculation.
(because uphillcost=0 and donwhillcutoff = 1.5)


4- downhillcutoff and upillcutoff logic that avoids steep inclines
==================================================================
Also when "consider_elevation" is changed to "false", bikers are not able to use ways with very steep inclines.
To avoid such ways, a minor change is possible:

assign downhillcost_       =  60    # %downhillcost_% | Cost for going downhill | number
assign downhillcutoff     = switch consider_elevation 1.5 10
assign uphillcost_         =  80     # %uphillcost_% | Cost for going uphill | number
assign uphillcutoff       = switch consider_elevation 1.5 10 



5- consider_traffic
===================
Default value is false, so that the traffic do not impact the routing.
The pseudo-tag "estimatd_traffic_class" is now available world-wide for primary, secondary and tertiary highways, it should be considered for biker routing.
(impact can be changed with the parameter, current setting "very low traffic" can be discussed)

6- smoothnesspenalty
====================
The surface tag is important, but the surface quality can very different between 2 highways with the same type and surface:
- surface is gravel (tracktype2), but very very bad quality ...or excellent 
- surface of a cycleway is asphalt, but with smoothness=horrible  ... and some monthes later after renovation the smoothness get "excellent"
 
When available, the smoothness is a good help!

Remark:
When the tag is considered in the profile, then the users have the posibility by need to add or change its value in OSM!!!
So, it impacts positively not only the own routing, but it is also a benefit for all OSM users. 

7- railway and barrier penalties
================================
OSM delivers detailed tags about railways and barriers, why not to add penalties when the cyclist have to pass them? 

assign railwaypenalty switch railway=level_crossing  155 0

Note: each rail-track is tagged, so the penalty depends on the number of rail-tracks to be crossed-and probably on the number of trains crossing that place per hour/day.

assign barrierpenalty  switch barrier= 0
                       switch barrier=block|bollard 140

Note: Barriers not only cost time, the risk of accidents increases. 

  

Here a new version containing these changes:
[trekking_v11.brf.txt](https://github.com/abrensch/brouter/files/13264221/trekking_v11.brf.txt)


But weaknesses remain (a.e. surface tag), and the maintenance remains difficult.
Better would be a new, well structured, easy to maintain profile!


Regards

zod:
> @zod Added a new paramTrack file and went back to old test. But for me it is still more a test for the generation of a 2nd route then a test for a param change.

I agree that the tests is far from ideal, but unfortunately I didn't find a better way to test if param overriding works. BRouter only creates the 2nd GPX file when the track differs from the existing track. If you would pass `avoid_unsafe=0` in the overrideParam test it would fail.

afischerdev:
@zod 
You don't have a better idea either. And if we test both aspects, it's ok too.

quaelnix:
> Yes, with a maintainer defect https://github.com/abrensch/brouter/issues/622 would probably not hang for weeks!!!

Your assumption of being smarter than the person who created the trekking profile leads to ill-conceived changes and I am simply not going to accept them. The logic you are trying to remove does exactly the right thing despite map errors.

I have already proven that the alleged routing error in the first example you gave in this issue ticket is caused by incorrect map data, and here is the proof that it is no different in the second example:
b486
Source: https://www.google.de/maps/@49.9787262,8.5809106,3a,60.3y,274.65h,88.89t/data=!3m6!1e1!3m4!1s79wZe-swSnTwfujOZnYEbw!2e0!7i16384!8i8192?entry=ttu

EssBee59:
@quaelnix,

thanks for the kind words!
It takes too much time for me to read comments that are confusing or completely out of place.
From now on I will simply ignore your posts.


@ other participants of this PR: 

in the first example Quelnix suspects a mapping error: the cycle_route would not use the secondary hw as mapped, but the parallel path!
(mapping error)

No proof is delivered for this!
I rather think, the route was planed for groups of cyclists biking together "sunday mornings" where the traffic remains low. And the path is this case probably not prefered to the secondary hw. (avoid path for groups)

in the second example Quaelnix delivers a picture taken nearly 1 km outside the route, strange for a proof!
Within the critical route part, no path exists parallel to the primary hw. 

Is it possible, that 10 % of the primary and 20% of the secondary hw in Hess are related BY ERRORS to cycle_route?

afischerdev:
Let us stay by facts. Please.

Routing is very individual and many answers are correct.

Change that appear to be critical:
```
- assign   ignore_cycleroutes       = false  # %ignore_cycleroutes% | Set true for better elevation results | boolean
+ assign   ignore_cycleroutes       = true   # %ignore_cycleroutes% | Set to false to favor the segments on cycleroutes | boolean
```
At first glance I would say it doesn't matter what the starting value is.
But cycle routes can present a few problems for routing:
- as shown in #622, routes can run on roads even though there is a bike path. This doesn't have to be a mistake, routes can be designed for tourist reasons or for sporting reasons (I don't know if there are any rules for that, didn't find one).
- the route relations is not always adjusted if access=no, highway=construction or something similar is suddenly set on a way. There are some issues in that direction
- more personally: normally, when I see a cycle path is available, but the route continues on the road, I decide to use the cycleway. But 'intelligent' apps with re-routing push me back onto the road. Annoying and with the potential to miss the right descent.
- 
@EssBee59 
1. Please tell us how did you find the errors on Hessen data?
2. When I copy the new profile to BRouter-Web profile box and select `ignore_cycleroutes=false`, it didn't follow the route (using first sample of #622). 


May be a feature 'follow cycleroute' could help for trekking profile that follows the direction of a route but uses the saver way. When I use the original profile with `ignore_cycleroutes=false` and `avoid_unsafe=true`, route doesn't change.

quaelnix:
> From now on I will simply ignore your posts.

We can reopen this pull request once you change your mind.

afischerdev:
Here we are again.

@quaelnix 
I don't think it's a good action to just close.
As your pictures show, action is needed.

We can
- enable `avoid_unsave` to work in the original profile
- in @EssBee59 's suggestion recall to follow the routes


Or maybe someone else would like to contribute a completely new trekking profile from their pool.

quaelnix:
> I don't think it's a good action to just close.

None of the proposed changes is acceptable in its current form and as long as @EssBee59 is neither willing to properly test his changes nor to discuss them, this pull request is not going to happen. His behavior is completely unacceptable. Period.

> enable avoid_unsave to work in the original profile

Yes, of course we can discuss if there are ways to improve the `avoid_unsafe` switch. But this has nothing to do with this pull request. And by the way a `highway=primary` section with `maxspeed=40`, `right of way` and asphalt in pristine condition is not unsafe.

> As your pictures show, action is needed.

Yes, someone needs to put the route relation [were it actually belongs](https://www.openstreetmap.org/way/201686001#map=15/49.9380/8.9556) and [properly tag the path that has not been touched in 10 years](https://www.openstreetmap.org/way/106962674#map=17/49.98726/8.57431) and make [this path](https://www.openstreetmap.org/way/252260991) the combined foot and cycleway [it actually is](https://www.google.de/maps/@49.9781713,8.5752883,3a,75y,188.27h,87.68t/data=!3m7!1e1!3m5!1s1cwcTOqdrz3adQMnYQQF4w!2e0!6shttps:%2F%2Fstreetviewpixels-pa.googleapis.com%2Fv1%2Fthumbnail%3Fpanoid%3D1cwcTOqdrz3adQMnYQQF4w%26cb_client%3Dmaps_sv.tactile.gps%26w%3D203%26h%3D100%26yaw%3D150.30319%26pitch%3D0%26thumbfov%3D100!7i16384!8i8192?entry=ttu).

EssBee59:
Hello afficherdev,
> 1. Please tell us how did you find the errors on Hessen data?

The statistics were created using overpass turbo
http://overpass-turbo.eu/

here some examples:
[overpass.txt](https://github.com/abrensch/brouter/files/13333868/overpass.txt)

A further example...

http://brouter.de/brouter-web/#map=9/51.0552/13.7219/osm-mapnik-german_style&lonlats=8.76663,50.017366;13.725973,51.044692

I tested with the current profile, the current profile + ignore_Cycleway=true and with the trekking_v11 including the changes  suggested above


image

quaelnix:
Even with this relaxed query: https://overpass-turbo.eu/s/1zDU, the rate of false positives is extremly high:

- B3: https://www.openstreetmap.org/way/364301602 (`cycleway:both = no`)
- Parallel cycleway: https://www.openstreetmap.org/way/131292244 (`bicycle=designated`)

-> Map is wrong.

quaelnix:
> but as example not considering surface=compacted in the same way as fine_gravel is for me a severe bug.

Another example of a gross misunderstanding of how the trekking profile works. Adding `compacted` to this exclusion list actually removes such ways from the list of unpaved ways, which throws off the tracktype cost logic and results in chosing:
- `tracktype=grade3; surface=compacted` over `tracktype=grade2`: https://brouter.de/brouter-web/#map=16/49.8394/8.6762/standard&lonlats=8.672164,49.839311;8.681874,49.841174
- `tracktype=grade4; surface=compacted` over `tracktype=grade2`: https://brouter.de/brouter-web/#map=17/49.84359/8.69299/standard&lonlats=8.69299,49.844662;8.69378,49.841263

afischerdev:
@quaelnix 
> ...as long as @EssBee59 is neither willing to properly test his changes nor to discuss them...

I think it's sentences like this that make @EssBee59 a little more reserved:
> Your assumption of being smarter than the person who created the trekking profile leads to ill-conceived changes...

Anyway, thanks for your samples, good basis for a discussion.

- Map errors 
  Yes, we don't have to talk about that. But we need a few considerations about how we get from the street to the cycleway.
  As remark only: In the first example I didn't find any route networks
  And in the second example, both have this information.
- smoothnesspenalty
  Yes, I've seen the table for that and `bad` is good enough for trekking bike. 
- compacted
  Yes, that breaks the `tracktype`. But also `cobblestone` as unpaved is not really understandable when I see `sett` as parved. (present in both profiles)

quaelnix:
> I think it's sentences like this that make @EssBee59 a little more reserved:

His complete inability to follow arguments is not new and now it is enough. Not to mention this .txt mess, broken markdown formatting in basically every post and the aparrent refusal to open his own pull requests. Again, I am not going to accept this anymore.

> Yes, we don't have to talk about that.

We need to talk about this because the design goal of the trekking profile was (and still is, in my opinion) to make it resistant to map errors.

> In the first example I didn't find any route networks

What do you mean by that? The [relation in question](https://www.openstreetmap.org/relation/299353) does carry a `network=lcn` tag.


> But also cobblestone as unpaved is not really understandable 

`cobblestone` is **excluded** from the list of unpaved roads!

quaelnix:
> This quote has other context

Yes, sorry, my bad.

> Historical I guess.

No, the (almost exclusive) purpose of these two variables is to feed the `probablyGood` logic, that can be considered a whitelist of ways that are likely kind of decent.

If you start feeding this whitelist with things that might not be good, you're taking it ad absurdum and end up with the routing artefacts I showed above.

You simply cannot look at one line of a profile and say this looks incorrect, lets change it. Doing it like that will almost always fail horribly.

> And I think we have to accept collaborators who are not that equipped.

To be honest, I don't know what to say anymore. If by now @EssBee59 is still not able to grasp the idea behind the cycleroute logic, then yes, maybe this is the end of the discussion. One last example (outside of his Hessian road bike bubble) where the new profile would fail horribly:
- https://brouter.de/brouter-web/#map=12/70.0387/24.9735/standard,Waymarked_Trails-Cycling,route-quality&lonlats=24.948823,69.99743;24.917142,70.080058

> E.g. if you take the example route Frankfurt - Dresden, around 500 km, it takes a lot of time to calculate.

I agree that we should try our best to keep the performance of the trekking profile as high as possible.

afischerdev:
I added a little test for the variables inside a profile.
And a minimal profile with the logic on our discussion: `probablyGood` is defined as true also when we have no surface definition. Because some said this is a cycle way.
But if we have an undefined value (e.g bricks, is added in the next generation lookups.dat) isunpaved becomes true.
You could use this for short results:

`$ gradlew brouter-expressions:test
`

Totorrr:
Hello @afischerdev @EssBee59, 

I suggest we close this MR now. 

Indeed, 

 - as of now, the branch has conflicts, 
 - the MR changes too many aspects of the trekking (main/default) profile at once, which makes clear thinking and peaceful discussion complicated (IMO almost each of the 7 points should have its own merge request), 
 - it seems there is a risk in this PR that some costfactor of this profile could be < 1, which does not follow the router requirements,
 - I am not sure yet that the `adj_param` logic is easy to work with and maintain, 
 - the trekking profile has evolved since then and it has implemented some of the ideas present here, namely, quoting @quaelnix https://github.com/abrensch/brouter/pull/641#issuecomment-1787017489 (not necessarily implemented the same way, but the idea is present) : 

   - 1-cycle_route Not every part of a cycle-route is a "safe harbour": PR #471 and PR #866 
   - 2-surface=compacted: indeed, some surfaces are missing in the `ispaved` and `isunpaved` logic of the profiles. **left to be improved**
   - 3-uphillcost penalty: *sometimes* the optimal route is not chosen (example given), I guess not too often. Also, this is a matter of personal preference, the current implementation says: if I do have to go up and down, I don't care about how I go up (steep or not), but I want ways going down to be the longest possible ways. This could be changed as a user "elevation preference". **left to be improved**
   - 4- downhillcutoff and upillcutoff logic that avoids steep inclines: I disagree with the approach. If you set `consider_elevation` to `false`, you don't want to consider elevation, steep or not. This can be dealt with in the previous point, 
   - 5- consider_traffic: already implemented and disabled by default in trekking, even though, when enabled, it does not benefit from the recent improvements in fastbike yet (PR #844), so not so much fine tuning. May not "fit" or be necessary in trekking, as trekking already tends to avoid "big" highways types (primary...), which, on OSM are supposed reflect the way "importance for the road grid", so "some" traffic correlation should exist here already. 
   - 6- smoothnesspenalty: a tiny bit was introduced PR #840, but it is not advanced, nor optimal. For me, the priority point for **left to be improved**.
   - 7- railway and barrier penalties: this would be a minimal improvement, still could be done. Also, a there is a TODO left by @abrensch from 2015 in `defaultaccess`: "add default barrier restrictions". So, **left to be improved**. 

As of now, I vote for closing this PR, and create smaller PRs for **left to be improved** points. 

ping @zod @rkflx

PR #645, opened 2023-11-21
BRouter translations

devemux86:
- English, Arabic, Catalan, Dutch, French, German, Greek, Italian, Polish, Spanish
- Move fixed strings to donottranslate-strings.xml
- Remove unused strings
- Enable app localization

There are several texts in the code that can be moved to strings.xml and translated as well.

The translations were contributed by Cruiser users:
- https://github.com/devemux86/cruiser/discussions/382

### DIFF
--- a/brouter-routing-app/build.gradle
+++ b/brouter-routing-app/build.gradle
@@ -20,8 +20,6 @@ android {
         minSdkVersion 14
         targetSdkVersion 33
 
-        resConfigs "en"
-
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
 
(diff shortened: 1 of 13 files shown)


afischerdev:
This breaks on git action but runs on a local gradlew build:

```
BInstallerActivity.java:174: Error: Format string 'summary_segments' is not a valid format string

<string name="summary_segments" formatted="false">Size=%s\nFree=%s</string>

```
Any idea?

devemux86:
Interesting that there is a Lint error.

However if we do not use the `formatted="false"` then the build of release apk shows the warning:
`Multiple substitutions specified in non-positional format of string resource string/summary_segments. Did you mean to add the formatted="false" attribute?`

We better try with positional arguments.

devemux86:
@afischerdev please run the workflow again.

Don't they run automatically after the first approval?

afischerdev:
> or split the string.

Yes, but this one has also warnings:

`Do not concatenate text displayed with `setText`. Use resource string with placeholders.`



> Don't they run automatically after the first approval?

I've seen, but can't say why.

devemux86:
@afischerdev please let's try again the build.

devemux86:
@afischerdev is the Lint successful now on these changes?

afischerdev:
@devemux86 fine now.

After merge I'll start the collection for the other hard coded strings.
I hope there are some candidates for remove.

afischerdev:
Thank you for contribution.

PR #647, opened 2023-11-22
A Dockerfile

joe-akeem:
So brouter can be run in Docker 😃


afischerdev (commented):
I'm not firm with Docker, but it looks ok for like it is.

Only I miss is a call for '[generate_profile_variants.sh](https://github.com/abrensch/brouter/blob/master/misc/scripts/generate_profile_variants.sh)'

joe-akeem:
@afischerdev the script was already called by the gradle build but the generated profiles were not used. I updated the Dockerfile and the documentation. Now, if you start the container without mounting a profiles folder, it will use the default profiles including the variants. If you do mount a folder containing custom profiles, it will replace the one with the default profiles. What do you think?

afischerdev:
@joe-akeem 
Thanks for the description, meanwhile I have run the docker build and get the server run.
So I've seen what docker is doing. I'm working on a Windows system, so the profile variant builder script doesn't work, but that is not relevant for docker. 

BTW for running it on Windows I had to invalidate the first line of `gradlew` and `server.sh`

afischerdev (commented):
Fine, I think.

PR #660, opened 2024-01-12
Fix 'sees' typo in consider_river switch

waldyrious:
Replace "rivers or sees" with "rivers, lakers, etc.", based on wording proposed by @gnbl in https://github.com/nrenner/brouter-web/issues/747.

### DIFF
--- a/misc/profiles2/fastbike.brf
+++ b/misc/profiles2/fastbike.brf
@@ -25,7 +25,7 @@ assign allow_motorways       = false  # %allow_motorways% | Set to true to allow
 
 assign consider_traffic      = false  # %consider_traffic% | Activate to avoid traffic | boolean
 assign consider_noise        = false  # %consider_noise% | Activate to prefer a low-noise route | boolean
-assign consider_river        = false  # %consider_river% | Activate to prefer a route along rivers or sees | boolean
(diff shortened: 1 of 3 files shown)


quaelnix:
@afischerdev, the profiles on the server apparently have not been updated for months. Could we sync them?

afischerdev:
@quaelnix 
I'll try that, have never done it before and don't know if my rights will work for that.

But we should do a minimal change: set the default new in
```
trekking.brf
assign   ignore_cycleroutes  = true
```
as discussed e.g. [here](https://github.com/abrensch/brouter/pull/641#issuecomment-1807091999).
That seems to be the main concern. All the other pending things could be discussed later on.

quaelnix:
> But we should do a minimal change: set the default new in

Why should we do that? There is even a dedicated trekking profile for exactly that case the brouter.de server: https://brouter.de/brouter-web/#map=7/51.011/10.015/standard,route-quality&profile=trekking-ignore-cr

I might have overreacted a bit in that pull request, but disabling `ignore_cycleroutes` by default makes zero sense.

afischerdev:
@quaelnix 
You are right. I always forget that feature. 
I only use the plain versions and so I always get a little irritated when I encounter the `ignore_routes` parameter..

PR #689, opened 2024-04-03
Update PMD to 7.0.0

zod:
PMD has finally [released version 7.0.0](https://docs.pmd-code.org/pmd-doc-7.0.0/pmd_release_notes.html) which offers better analysis and new rules.

It required a gradle update but Android Studio kept nagging me to upgrade gradle & AGP (Android Gradle Plugin) anyway...


build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.4.2'
+        classpath 'com.android.tools.build:gradle:8.3.1'
```
devemux86, line 10: Note that many developers use IDEA instead of Android Studio.

Latest IDEA does not still support the Android Gradle plugin 8.3.X.
IDEA works fine with Android Gradle plugin 8.2.X (currently `8.2.2`).

.github/workflows/gradle.yml
```diff
@@ -16,10 +16,10 @@ jobs:
     environment: BRouter
     steps:
     - uses: actions/checkout@v4
-    - name: Set up JDK 11
+    - name: Set up JDK 17
```
afischerdev, line 19: One remark only, the BRouter server still uses JDK 11. 
Therefore, the output lib cannot be used directly.

build.gradle
zod: I guess we don't use any features from 8.3.x, so switching back to 8.2.2 shoudn't cause any issues.

.github/workflows/gradle.yml
zod: If we want to compile for older Java releases we should [target a specific java version](https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_cross_compilation). It seems reasonable to target Java 11 but use JDK 17 to build.

zod:
I've implemented targeting a specific Java version using conventions plugins because the gradle documentation suggested to [avoid cross configuration](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration) using allprojects/subprojects

Currently it targets Java 11 because we already use some functionality from Java 9. It's mostly in test code but the new Hgt processing also uses it. This code won't be run on Android so it should not cause any issues. Going forward we should pick a Java version we really want to target and adopt the build configuration to treat using newer code as error.

afischerdev:
I made some tests:
- export repository as zip and run Android Studio / gradle command line
- export repo as git export and run Android Studio ( gradle command line
- use the generated lib as test-server lib
All tests passed ok.
So I would like to merge after version 1.7.4

afischerdev:
> I'd also like to add a dependency from distZip ...

As you see in the current release, the distribution file is not produced.
I used 
```
distZip {
    dependsOn fatJar
    dependsOn (':brouter-routing-app:assemble')
    archiveFileName = 'brouter-' + project.version + '.zip'
}
```

zod:
Your suggestion would break builds without android SDK because the `:brouter-routing-app:assemble` wouldn't exist. The docker build doesn't add the android SDK and therefore would fail.

afischerdev:
Yes, you are right. This  is only a temporary solution to get an apk inside the zip when Android tasks are present.
My attempts with doLast or Exception were unsuccessful. So no idea at the moment.

afischerdev:
I've masked the dependencies now, that should do the job.

`if (file('../local.properties').exists()) dependsOn (':brouter-routing-app:assemble')
`

zod:
I'm not to happy with this check because not we check for `local.properties` in multiple locations instead of detecting if the target was added, but let's just use it for now and perhaps I'll come up with a different check in a future PR.

afischerdev:
@zod 
I understand that, but we already use this way in [settings.gradle](https://github.com/abrensch/brouter/blob/master/settings.gradle).
My tests with hasProperty('sdk.dir') or similar were not successful. My be there are other ideas?

Issue #628, opened 2023-09-14
Simplify the routing core and data format


abrensch:
This is a major change I have in mind since some time now, but it's a major step and not clear yet whether this can be a patch developed in a branch, or rather is a follow-up project of BRouter

The issue is that the routing core of BRouter is far more complex then it has to be.

That became obvious after I implemented the "DirectWeaver": All the code that was used before to convert the micro-tiles into an intermediate format and all the cache-management to manage this converted data in Memory, this is all obsolete, but still part of the codebase. With the DirectWeaver, only two representations of the data left: the encoded data as stored in the RD5s, and the actual routing graph. Nothing in between anymore.

Another major part of the complexity is the design-decision to distinguish between "Network-Nodes" and "Transfer-Nodes". 2/3 of the Nodes are "Transfer-Nodes", meaning they are only intermediate points along a way with no foreign connections. The idea behind that design decision was that it saves processing time and memory to have a special, simpler treatment for transfer nodes, not processing them through the priority queue and not encoding the OSM-tagging for every single section along a path of transfer nodes. However, nowadays I question this design decision. It's probably still a good idea to have some shortcuts for transfer nodes, but it's definitely a bad idea to encode this distinction into the data-format. One reason is that, in reality, the answer to the question "what is a network node" depends on the routing profile. For a car profile, not every connecting footpath creates a network node. So if there would be some efficient processing for nodes that are effectively transfer-nodes, letting them bypass the priority queue, that should be enough. For the memory footprint of the routing graph, each transfer node would be represented by one single java object (because OsmNode and OsmLink do instance-sharing via "technical inheritance"), so that will be slightly more compared to the current implementation (where the list of transfer node is encoded in a byte-array), but not double it, so that does not justify the complexity introduced by the concept of transfer nodes.

And there are smaller issues blowing up the code and candidates for removal:

- the logical separation of the 5*5 degree RD5's into 25 1*1 degree sections is probably bullshit. There were 2 design considerations behind that: 1) to access a small amount of data, only one (or few) of the 25 micro-tile-indexes has to be read and 2) the data for a small geographc region is better clustered in the physical file, making file access more efficient. Nowadays I think that 1) is irrelevant (at least since I changed from 1/80 degree microtiles to 1/32 degree microtiles...) and 2) can event be improved by just re-ordering the microtiles, without having a seperate logical layer.

[...]

PR #697, opened 2024-05-11
Remove unused traffic simulation code

quaelnix:
Fixes a small part of: #628 by removing the traffic simulation code which is no longer needed after the introduction of Spatial SQL based pseudo tag generation.


zod:
I've haven't really looked at the changes in detail, but I think it's very good to get rid of obsolete code to keep the codebase manageable.

PR #698, opened 2024-05-13
Minor changes for french language

alkino:
The main one is `Gratuit` > `Libre`
Gratuit means free as in free beer.
Libre means free as in available place.

### DIFF
--- a/brouter-routing-app/src/main/res/values-fr/strings.xml
+++ b/brouter-routing-app/src/main/res/values-fr/strings.xml
@@ -11,7 +11,7 @@
     <string name="action_update">Mettre à jour %s</string>
     <string name="action_select">Sélectionner les segments</string>
     <string name="action_cancel">Arrêter le téléchargement</string>
-    <string name="summary_segments">Taille=%1$s\nGratuit=%2$s</string>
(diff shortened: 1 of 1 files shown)


devemux86:
@alkino

Thanks, may I ask about inserting a space before the punctuation marks?

alkino:
Because in french this is not correct to not insert one.

https://en.wikipedia.org/wiki/History_of_sentence_spacing#French_and_English_spacing
```
French spacing inserted spaces around most punctuation marks, but single-spaced after sentences, colons, and semicolons.
```

This is really a *typographic mistake* to not add one.

devemux86:
@alkino then is it needed a [non-breaking space](https://en.wikipedia.org/wiki/Non-breaking_space) to avoid line break with punctuation mark alone in next line?

`&nbsp;`

alkino:
Yes, it would be better.

afischerdev:
May be a `&#160;` could do the job.

afischerdev:
@alkino 
Your PR has not passed the build process.
Please see [here](https://github.com/abrensch/brouter/actions/runs/9059876347/job/24986593755?pr=698#step:6:131)
Do you like to complete it?

PR #717, opened 2024-07-12
Update Android Version

afischerdev:
Google wants us to change the Android version by the end of August.
So here it comes.

This gives you the opportunity to make some additional changes that are only relevant for development:

- new versions for the libraries 
- some compiler params for better control 
- new gradle app name for compatibility with future gradle versions
- replace printstacktrace with more robust logging

### DIFF
--- a/brouter-routing-app/build.gradle
+++ b/brouter-routing-app/build.gradle
@@ -24,7 +24,7 @@ android {
         resValue('string', 'app_version', defaultConfig.versionName)
 
         minSdkVersion 21
-        targetSdkVersion 35
+        targetSdkVersion 34
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
--- a/brouter-routing-app/build.gradle
+++ b/brouter-routing-app/build.gradle
@@ -18,7 +18,7 @@ android {
         namespace 'btools.routingapp'
         applicationId "btools.routingapp"
 
-        versionCode 53
+        versionCode 54
         versionName project.version
 
         resValue('string', 'app_version', defaultConfig.versionName)
--- a/brouter-routing-app/build.gradle
+++ b/brouter-routing-app/build.gradle
@@ -7,7 +7,7 @@ plugins {
 }
 
 android {
-    compileSdkVersion 33
+    compileSdk 34
 
     defaultConfig {
         namespace 'btools.routingapp'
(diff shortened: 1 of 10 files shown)


brouter-routing-app/build.gradle
```diff
@@ -17,10 +22,9 @@ android {
         versionName project.version
 
         resValue('string', 'app_version', defaultConfig.versionName)
-        setProperty("archivesBaseName", "BRouterApp." + defaultConfig.versionName)
 
-        minSdkVersion 14
-        targetSdkVersion 33
+        minSdkVersion 21
+        targetSdkVersion 35
```
devemux86, line 27: Current target API level requirement is 34:
https://developer.android.com/google/play/requirements/target-sdk

35 will not be required until August 2025.

afischerdev: @devemux86 
That is correct. But my Studio tells me 'not targeting the latest versions of Android' when I use API level 34.

Anyway the main problem I see is the `minSdkVersion`.  
With using new libraries, level 21 is required.

When going back with e.g. `androidx.appcompat:appcompat:1.6.1` we could go back to level 19. This will bring an error: `Cannot fit requested classes in a single dex file` for deprecated classes.
But setting `minifyEnabled false` for debug mode takes care of this. And should not bring to much trouble in development because we set all BRouter classes to `keep` in last update.

What do you think? Going back?

BTW: Is there still an IDEA problem with
`classpath 'com.android.tools.build:gradle:8.3.1`

devemux86: @afischerdev 

The latest stable version is compile and target SDK 34.
Android 15 will be released the next months.

> With using new libraries, level 21 is required.

This should be normal, we can update the dependencies and use min sdk 21.

> Is there still an IDEA problem with

Unfortunately we still have to use version 8.2.2 for IDEA.

zod: API 21 was released 10 years ago and is the latest version supported by Google Play Services. This shouldn't affect many devices so I think it's a good idea to increase the minSdkVersion. I think targetSdkVersion should be API 34 because API 35 is only released as preview.

afischerdev: I'll change the version back. I don't mind.
But no, there is no preview Android installed, it's a 35 rev. 1.

devemux86: Android 15 is currently on Beta 4.
https://developer.android.com/about/versions/15/overview

Also we need first to compile with sdk 35 and then advance the target to 35.
There is no need to hurry until August 2025.

The current requirement for August 2024 is just the target 34.

devemux86: IDEA has been updated and now supports the latest Android Gradle plugin `8.5.1`.

PR #725, opened 2024-09-02
Update gradle and docs

afischerdev:
Some updates for gradle and the revisions doc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.2.2'
+        classpath 'com.android.tools.build:gradle:8.5.0'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.5.0'
+        classpath 'com.android.tools.build:gradle:8.5.2'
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.2.2'
+        classpath 'com.android.tools.build:gradle:8.5.0'
```
devemux86, line 9: We can use the `8.5.2` version.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-all.zip
```
devemux86, line 6: Don't know if it is better to use the `-bin` distribution?

build.gradle
afischerdev: Ok

gradle/wrapper/gradle-wrapper.properties
afischerdev: I don't mind. -bin is smaller.

afischerdev:
Find a minor issue during my testing:
The app doesn't have a profile list after the first start and the visit on the InstallerActivity.

Issue #723, opened 2024-08-17
Strange turn instructions?


devemux86:
There is a report in Cruiser forum about wrong turn instructions by BRouter app with car profile:
- https://github.com/devemux86/cruiser/discussions/622

Slight turn right instead of sharp turn right in image 1 with [BRouter link](https://brouter.de/brouter-web/#map=18/50.23150/8.45974/standard&lonlats=8.460445,50.231721;8.459726,50.231851&profile=car-fast).

Unnecessary (?) keep right in image 2 with [BRouter link](https://brouter.de/brouter-web/#map=15/50.2813/8.5915/standard&lonlats=8.582372,50.282182;8.595482,50.280009&profile=car-fast).

<img src="https://github.com/user-attachments/assets/43dca7cd-2358-41a1-aede-59564f303ae9" width="50%"/><img src="https://github.com/user-attachments/assets/5c428be0-7c07-4b4c-ad97-dc10ee6c750d" width="50%"/>

poutnikl:
I assume one of possible reasons may be that the reported light turn may be result of assigning the big portion of turning to arriving and departing roads, not to the crossing, especially if too small value of turnInstructionCatchingRange is used.

EssBee59:
I think, the turn instruction is not really wrong, the situation have to be considered:
-The Brouter do not generate any turn instruction untill the point "Middle" is reached, as the first "right turn" of nearly 90 degre is by following the road.
-Now, from this point a kind of "smooth" right turn is coming. 

<img width="506" alt="image" src="https://github.com/user-attachments/assets/b9dd8e9a-074a-4868-b1bd-ec60ca80dfcb">

afischerdev:
@devemux86 
Thanks for the nice samples. 

I could do a small research for this and I think at the moment both are 'errors' and should be reworked.

1. Slight right: We have an angle collector but this is limited by a contant value of 2m. In this case not very useful. When using a limit of 40m a result of 'turn right' is delivered. But this has be to checked against other rules (e.g. roundabouts).
2. Keep right:  The error is in the variable `hasSomethingMoreStraight`, it is checked against `minAbsAngeRaw` but this is not always filled, so when I check this the expected result of 'no turn instruction' is delivered.

I can do more on this next week and will be back here.

afischerdev:
Well, it is not as easy as I thought on the first view.
[...]

Issue #710, opened 2024-06-17
Problem with communication between BRouter lib 1.7.5 and BRouter-web


afischerdev:
With PR #700 we switched the GeoJSON response
```
application/vnd.geo+json
to
application/geo+json
```
The BRouter-web is still testing the old version

The new lib 1.7.5 is running for
https://brouter.de/brouter-test/
So the problem is visible and could be check against the testserver.

PR #726, opened 2024-09-05
Rework voicehint

afischerdev:
Discussion and test results please see #723

This contains also an update for ServerHandler to enable old requests
please see #710

Set an environment variable like

`export USEOLDSERVERFORMAT="yes"`


afischerdev:
The variable type from `USEOLDSERVERFORMAT` has changed from system to java variable.
So the you could add `-DuseRFCMimeType=true` to your java options. System variable no longer needed.
Please [see](https://github.com/abrensch/brouter/blob/master/misc/scripts/standalone/server.sh)

afischerdev:
Are further tests needed here?

Issue #682, opened 2024-03-27
Car route should not begin/end on a motorway


polyscias:
In most well developed countries it is forbidden to stop on a motorway (highway=motorway/trunk), so when a router is searching for the begin/end of the route is should exclude motorways. Example of things going wrong: 

http://brouter.de/brouter-web/#map=16/51.9858/7.5655/standard&lonlats=7.555976,51.984186;7.550569,51.985726&profile=car-eco

poutnikl:
But the route start/stop points do not necessarily mean you physically start/stop there.

polyscias:
> But the route start/stop points do not necessarily mean you physically start/stop there.

Indeed, drivers should never 100% trust the routes that are calculated and keep thinking themselves. Still it would be good to have the route as good as possible.

> It may mean you already are on a motorway (common case if you are a "navigator" during abroad car trip) or you just want to know how to get on the motorway.

I do not know any car routing device that is integrating brouter and a driver should not operate a smartphone during driving so the only remaining case is where a co-driver uses his/her smartphone.

This argument only holds for the "From", for the "To" I see not a single good reason it should end on a motorway.

Implementation should be relatively simple, find these points is separate functionality and you just have to filter out highway=motorway/trunk running the algorithm.

devemux86:
> It may mean you already are on a motorway (common case if you are a "navigator" during abroad car trip)

Correctly. :+1:

Routing (automatic or not) is necessary for navigation.
Users expect that routes can start anywhere (including motorways).

Routing should not start arbitrarily on random roads other than the road the user is on.

---

[Cruiser](https://github.com/devemux86/cruiser) is a navigator using BRouter for car and motorcycle driving.

poutnikl:
> Indeed, drivers should never 100% trust the routes that are calculated and keep thinking themselves. Still it would be good to have the route as good as possible.

My post was not in context of routing reliability, but about starting/stopping the routing on motorways intentionally.

> 
> I do not know any car routing device that is integrating brouter and a driver should not operate a smartphone during driving so the only remaining case is where a co-driver uses his/her smartphone.

You have probably missed the navigator. I was not speaking about a driver on a motorway. I occasionally perform navigator role on abroad vacation trips with friends and start routing at motorway in real time frequently.

> This argument only holds for the "From", for the "To" I see not a single good reason it should end on a motorway.

The task to get on a motorway is very frequent, as it may be challenging to orientation. Getting already on the motorway, you often do not need navigation for hours.

[...]

Issue #600, opened 2023-07-27
Avoid starting routes on ferry-ways


PartTimeDataScientist:
If a route request is started off a routable track the routing algorithm seems to try to use the next point with `bicycle=yes`. That doesn't make too much sense if that point is in the middle of a ferry-way:
![image](https://github.com/abrensch/brouter/assets/30908795/21b28892-3c8c-4c94-ad28-1303b1991ab3)

Example is here: https://brouter.de/brouter-web/#map=16/51.1081/7.8883/osm-mapnik-german_style,Waymarked_Trails-Cycling&lonlats=7.885628,51.100827;7.878391,51.114132

quaelnix:
If you uncheck the "allow_ferries" profile option, the issue is gone.
> That doesn't make too much sense if that point is in the middle of a ferry-way

I agree with this in principle, but how should we solve the dilemma that starting a route on a ferry is mostly undesirable, but still valid?

PartTimeDataScientist:
If I disallow ferries this setting will be used for the complete route calculation which might be undesired.

I don't know how the algorithm determines the starting point but my proposal would be to look for `bicycle=yes` and `ferry=no` *only* for starting the route.

Anyway, this is arguably an edge case but I wanted to document it nonetheless...

quaelnix:
> I don't know how the algorithm determines the starting point 

Afaik it simply picks the nearest point that lies on a way with costfactor < 9999.

PR #728, opened 2024-09-07
Start way check added

afischerdev:
This is the answer on the issues #600 and #682.

The idea is to place a definition for excludes highways in the profile.
This is scanned at routing start and an array is created with the index of the lookup entry and the lookup value.
The array is then used to exclude the way from the routing.

The test server is prepared for using this feature.
You have to assign a new global rule in the profile - there is no entry for this feature in the profiles at the moment.
```
assign   check_start_way          = true   # %check_start_way% | Activate a test for the starting way | boolean | noStartWay=route,ferry;highway,motorway;highway,motorway_link
```
As you see the comment part of the assign has in this case an array with 4 entries - please note, the web client can not check that at the moment, so you don't see the assigned entry in the option window (I guess, some of the RegEx in [Profile.js](https://github.com/nrenner/brouter-web/blob/6b24a8790eea72bf6e95a63bfaadc0417f36d3fa/js/control/Profile.js) should be changed).
The fourth part of the array contains the excluded entries.
```
noStartWay= - a fix constant for this new feature
route,ferry - entry for route=ferry
; - more entries comes here

```

Here the links for the test server:
[issue 600](https://brouter.de/brouter-test/#map=14/51.1065/7.8808/standard&lonlats=7.885628,51.100827;7.878391,51.114132)
[issee 682](https://brouter.de/brouter-test/#map=15/51.9843/7.5538/standard&lonlats=7.555976,51.984186;7.550569,51.985726)

### DIFF
--- a/brouter-codec/src/main/java/btools/codec/MicroCache2.java
+++ b/brouter-codec/src/main/java/btools/codec/MicroCache2.java
@@ -161,7 +161,8 @@ public final class MicroCache2 extends MicroCache {
           int ilontarget = ilon + dlon_remaining;
           int ilattarget = ilat + dlat_remaining;
           if (matcher != null) {
-            if (!matcher.start(ilon, ilat, ilontarget, ilattarget)) {
+            boolean useAsStartWay = wayValidator.checkStartWay(wayTags.data);
+            if (!matcher.start(ilon, ilat, ilontarget, ilattarget, useAsStartWay)) {
               matcher = null;
             }
           }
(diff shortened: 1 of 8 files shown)


rtwalz:
Just want to say this feature helped me for a project, I really appreciate it :)

afischerdev:
@rtwalz 
Thank you for the info. It is alway good to know that someone is using it - and it is working as intended.

afischerdev:
@rtwalz 
I understand your intention, but don't you think it is better do connect the sideways to junctions? It is hard to say if this sideway is behind a wall or not.

Issue #387, opened 2022-01-01
car routing then direct line


abdullahO2:
Thank you for your efforts in this software, and I tried it and found it useful and wonderful, but in my country (Saudi Arabia) there are large desert areas without roads and sometimes we need to go to a place inside that desert, and BRouter does not support routing for places where there are no roads, so I have a suggestion to put a profile that supports A route to the nearest nearby road and then puts a direct line to the specified waypoint, and this solution is found in many applications such as Navitel and Osmand (example):

![image](https://user-images.githubusercontent.com/61907628/147856008-e4fd52c8-7735-4bd0-ac46-16b191dd7b91.png)

If this solution is possible, please guide me on how to do it, thank you

abrensch:
No, in BRouter-(Web), Waypoints are always matched to the routable-network.

There's a hidden feature for some sort of "beeline-routing", by putting a nogo-circle over 2 Waypoints ( see here for details: https://www.marcusjaschen.de/blog/2021/brouter-direkte-verbindung-zwischen-zwei-wegpunkten/ )

that's good if you need a connection within your route where there's no mapped way. But it doesn't help you for off-road start- or detination-points.

abdullahO2:
Thank you, I was hoping to find this solution for applications that depend on BRouter, like Orux, Locus and others, Regards

polyscias:
It is not that if Brouter would add support for direct lines, Apps like Orux, Locus and OsmAnd would directly support direct lines, these Apps have to communicate their desires to BRouter and such new functionality will need code changes in these Apps.

In the link @abrensch posted you can see BRouter has already a way to do this using No-Go-Area's the problem is on the App side to support this. Note that Apps might also support this but just calling Brouter multiple times, only for the pieces so the route where the user wants routing.

Adding this functionality to the Apps, see also [Allow straight lines #68](https://github.com/nrenner/brouter-web/issues/68) seems to be the hard problem, especially the user interface. So if you want this, you better check these Apps, see if there is already a "direct line" request and if not add one.

abdullahO2:
Thank you very much for explaining that, regards

abdullahO2:
Related discussion:
https://github.com/devemux86/cruiser/discussions/204

abdullahO2:
@afischerdev @abrensch Does the current state of development allow for the development of something for this proposal? This is a video to understand the intended proposal:
[ straight line - normal routing - straight line ](https://www.youtube.com/watch?v=ketZyL-vUCU)
[...]

PR #729, opened 2024-09-12
Enable a dynamic range search for matched way points

afischerdev:
This should enable the #387 issue.

It could be tested on the BRouter test server together with the #728 and #726 
[Sample](https://brouter.de/brouter-test/#map=10/23.6250/43.5354/standard&lonlats=43.503174,23.595478;43.421661,24.40659&profile=car-eco) 

The logic is only valid for the first and the last point. So please test 2 points routes only at the moment.

For a dynamic range test you will need a new profile constant
```
assign use_dynamic_range   = true  # %use_dynamic_range% | Enable distant start/end points | boolean
```

Please remember to test #728 with this entry
```
assign check_start_way      = true   # %check_start_way% | Activate a test for the starting way | boolean | noStartWay=route,ferry;highway,motorway;highway,motorway_link
```

### DIFF
--- a/brouter-core/src/main/java/btools/router/FormatJson.java
+++ b/brouter-core/src/main/java/btools/router/FormatJson.java
@@ -122,7 +122,7 @@ public class FormatJson extends Formatter {
         .append(sele).append("],\n");
       nn = n;
     }
-    sb.deleteCharAt(sb.lastIndexOf(","));
+    if (t.nodes != null && !t.nodes.isEmpty()) sb.deleteCharAt(sb.lastIndexOf(","));
 
     sb.append("        ]\n");
     sb.append("      }\n");
--- a/brouter-core/src/main/java/btools/router/RoutingContext.java
+++ b/brouter-core/src/main/java/btools/router/RoutingContext.java
@@ -76,6 +76,7 @@ public final class RoutingContext {
   public double waypointCatchingRange;
   public boolean correctMisplacedViaPoints;
   public double correctMisplacedViaPointsDistance;
+  public boolean useDynamicDistance;
 
   private void setModel(String className) {
     if (className == null) {
@@ -167,6 +168,8 @@ public final class RoutingContext {
     defaultC_r = expctxGlobal.getVariableValue("C_r", 0.01f);
     // Constant power of the biker (in W)
     bikerPower = expctxGlobal.getVariableValue("bikerPower", 100.f);
+
+    useDynamicDistance = expctxGlobal.getVariableValue("use_dynamic_range", 0f) == 1f;
 
     boolean test = expctxGlobal.getVariableValue("check_start_way", 1f) == 1f;
     if (!test) expctxGlobal.freeNoWays();
--- a/brouter-core/src/main/java/btools/router/RoutingEngine.java
+++ b/brouter-core/src/main/java/btools/router/RoutingEngine.java
@@ -587,7 +587,13 @@ public class RoutingEngine extends Thread {
         mwp.direct = waypoints.get(i).direct;
         matchedWaypoints.add(mwp);
       }
+      int startSize = matchedWaypoints.size();
       matchWaypointsToNodes(matchedWaypoints);
+      if (startSize < matchedWaypoints.size()) {
+        refTracks = new OsmTrack[matchedWaypoints.size()]; // used ways for alternatives
+        lastTracks = new OsmTrack[matchedWaypoints.size()];
+        hasDirectRouting = true;
+      }
 
       routingContext.checkMatchedWaypointAgainstNogos(matchedWaypoints);
 
@@ -968,7 +974,53 @@ public class RoutingEngine extends Thread {
   // geometric position matching finding the nearest routable way-section
   private void matchWaypointsToNodes(List<MatchedWaypoint> unmatchedWaypoints) {
     resetCache(false);
-    nodesCache.matchWaypointsToNodes(unmatchedWaypoints, routingContext.waypointCatchingRange, islandNodePairs);
+    boolean useDynamicDistance = routingContext.useDynamicDistance;
+    double range = routingContext.waypointCatchingRange;
+    boolean ok = nodesCache.matchWaypointsToNodes(unmatchedWaypoints, range, islandNodePairs);
+    if (!ok && useDynamicDistance) {
+      logInfo("second check for way points");
+      resetCache(false);
+      range = -range;
+      List<MatchedWaypoint> tmp = new ArrayList<>();
+      // only first or last checked
+      if (unmatchedWaypoints.get(0).crosspoint == null) tmp.add(unmatchedWaypoints.get(0));
+      if (unmatchedWaypoints.get(unmatchedWaypoints.size()-1).crosspoint == null) tmp.add(unmatchedWaypoints.get(unmatchedWaypoints.size()-1));
+
+      ok = nodesCache.matchWaypointsToNodes(tmp, range, islandNodePairs);
+    }
+    if (!ok)  {
+      for (MatchedWaypoint mwp :unmatchedWaypoints) {
+        if (mwp.crosspoint == null) throw new IllegalArgumentException(mwp.name + "-position not mapped in existing datafile");
+      }
+    }
+    if (useDynamicDistance) {
+      List<MatchedWaypoint> waypoints = new ArrayList<>();
+      for (int i = 0; i < unmatchedWaypoints.size(); i++) {
+        MatchedWaypoint wp = unmatchedWaypoints.get(i);
+        if (wp.waypoint.calcDistance(wp.crosspoint) > routingContext.waypointCatchingRange) {
+          MatchedWaypoint nmw = new MatchedWaypoint();
+          if (i==0) {
+            nmw.waypoint = new OsmNode(wp.waypoint.ilon, wp.waypoint.ilat);
+            nmw.crosspoint = new OsmNode(wp.waypoint.ilon, wp.waypoint.ilat);
+            nmw.direct = true;
+            wp.waypoint = new OsmNode(wp.crosspoint.ilon, wp.crosspoint.ilat);
+          } else {
+            nmw.waypoint = new OsmNode(wp.crosspoint.ilon, wp.crosspoint.ilat);
+            nmw.crosspoint = new OsmNode(wp.crosspoint.ilon, wp.crosspoint.ilat);
+            nmw.node1 = new OsmNode(wp.node1.ilon, wp.node1.ilat);
+            nmw.node2 = new OsmNode(wp.node2.ilon, wp.node2.ilat);
+            nmw.direct = true;
+            wp.crosspoint = new OsmNode(wp.waypoint.ilon, wp.waypoint.ilat);
+          }
+          nmw.name = wp.name;
+          waypoints.add(nmw);
+          wp.name = wp.name + "_add";
+        }
+        waypoints.add(wp);
+      }
+      unmatchedWaypoints.clear();
+      unmatchedWaypoints.addAll(waypoints);
+    }
   }
 
   private OsmTrack searchTrack(MatchedWaypoint startWp, MatchedWaypoint endWp, OsmTrack nearbyTrack, OsmTrack refTrack) {
--- a/brouter-mapaccess/src/main/java/btools/mapaccess/NodesCache.java
+++ b/brouter-mapaccess/src/main/java/btools/mapaccess/NodesCache.java
@@ -177,7 +177,8 @@ public final class NodesCache {
       }
 
       MicroCache segment = osmf.getMicroCache(ilon, ilat);
-      if (segment == null) {
+      // needed for a second chance
+      if (segment == null || (waypointMatcher != null && ((WaypointMatcherImpl) waypointMatcher).useDynamicRange)) {
         checkEnableCacheCleaning();
         segment = osmf.createMicroCache(ilon, ilat, dataBuffers, expCtxWay, waypointMatcher, directWeaving ? nodesMap : null);
 
@@ -282,15 +283,15 @@ public final class NodesCache {
     return existing;
   }
 
-  public void matchWaypointsToNodes(List<MatchedWaypoint> unmatchedWaypoints, double maxDistance, OsmNodePairSet islandNodePairs) {
+  public boolean matchWaypointsToNodes(List<MatchedWaypoint> unmatchedWaypoints, double maxDistance, OsmNodePairSet islandNodePairs) {
     waypointMatcher = new WaypointMatcherImpl(unmatchedWaypoints, maxDistance, islandNodePairs);
     for (MatchedWaypoint mwp : unmatchedWaypoints) {
       int cellsize = 12500;
-      preloadPosition(mwp.waypoint, cellsize);
+      preloadPosition(mwp.waypoint, cellsize, 1);
       // get a second chance
       if (mwp.crosspoint == null) {
         cellsize = 1000000 / 32;
-        preloadPosition(mwp.waypoint, cellsize);
+        preloadPosition(mwp.waypoint, cellsize, maxDistance < 0 ? 5 : 2);
       }
     }
 
@@ -305,7 +306,8 @@ public final class NodesCache {
           mwp.crosspoint = new OsmNode(mwp.waypoint.ilon, mwp.waypoint.ilat);
           mwp.direct = true;
         } else {
-          throw new IllegalArgumentException(mwp.name + "-position not mapped in existing datafile");
+          // do not break here throw new IllegalArgumentException(mwp.name + "-position not mapped in existing datafile");
+          return false;
         }
       }
       if (unmatchedWaypoints.size() > 1 && i == unmatchedWaypoints.size() - 1 && unmatchedWaypoints.get(i - 1).direct) {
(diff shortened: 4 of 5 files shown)


devemux86:
@afischerdev 

> The logic is only valid for the first and the last point. So please test 2 points routes only at the moment.

Testing it I see that besides the first and last point, it also works with the via points?

If we place a via point away from the road, it produces 2 beelines going to and returning from the via point.

I am not sure what users want, a global setting for all points (start + via + end) or more detailed control?

---

@abdullahO2 can you test it on BRouter test website and tell us your opinion?

(instructions are in the [1st post](https://github.com/abrensch/brouter/pull/729#issue-2521976075))

afischerdev:
@devemux86 
> Testing it I see that besides the first and last point, it also works with the via points?

When you test this in the web client this will work because web client sends line by line to BRouter. 
The Android client will break at the via point. I'm working on it - it was my own restriction.

abdullahO2:
@afischerdev and @devemux86

I'm incredibly excited to report that I've tested the new feature on the BRouter test website, and it works flawlessly! I tested a route with two off-road waypoints in a remote desert area, and BRouter seamlessly generated the optimal path, including the necessary beeline segments. This feature will significantly enhance my ability to explore remote areas and plan adventures that were previously difficult to navigate.

I appreciate your responsiveness and hard work in developing this feature so quickly. It's truly remarkable to see it come to life, especially considering the complexities involved.  I want to extend my special thanks to @afischerdev for implementing this complex feature so effectively, and to @devemux86 for his insightful feedback and testing.

To answer @devemux86's question about via points, I find the current implementation, which generates beelines to and from off-road via points, to be ideal for my needs. It strikes a great balance between flexibility and practicality, allowing me to plan routes that include off-road sections without sacrificing ease of use.

I'm so impressed with BRouter and its ongoing development that I'd love to make a donation to support the project. Is there a way to contribute financially? If so, could you please provide me with the relevant information?

Best regards,

Abdullah Abdulrahman

afischerdev:
Now the via points are enabled for the library.
I tested something like this [with 'handmade' and dynamic via points at start, end and in the middle](https://brouter.de/brouter-test/#map=9/23.8670/43.8368/osm-mapnik-german_style&lonlats=43.812801,24.355228;43.306046,24.230062;43.335557,23.679911;43.591001,23.369994;43.767457,23.301897;43.71973,23.22368&straight=1,2&profile=car-eco)

Don't forget to assign the new constant

`assign use_dynamic_range   = true  # %use_dynamic_range% | Enable distant points | boolean`

abdullahO2:
@afischerdev

I've noticed that the dynamic_range feature works within a range of approximately 20km from the nearest road. Attempting to exceed this range results in a message like "Start marker is too far from a route."

Would it be possible to make this distance configurable through the profile settings? This would provide users with greater flexibility in situations where they need to navigate further away from established roads.

Thanks again for your excellent work on this feature!

devemux86:
@abdullahO2 @afischerdev 

Isn't this the `waypointCatchingRange` profile numeric parameter (default 250 m)?

abdullahO2:
@devemux86

I've tried setting `assign waypointCatchingRange = 3000` in the profile, but it didn't seem to change anything. I tested this with the destination point, and the "Destination marker is too far from a route" message still appears when exceeding approximately 20km from the nearest road.

afischerdev:
@abdullahO2 
Do you have some coordinates from your test that we could check against the new lib?

abdullahO2:
> @abdullahO2 Do you have some coordinates from your test that we could check against the new lib?

https://brouter.de/brouter-test/#map=10/17.8696/46.8004/standard&lonlats=46.895592,17.601443;46.720824,17.530452;46.587524,17.680655

afischerdev:
@abdullahO2 
Thanks for your sample.
When I add this it works:

`assign use_dynamic_range   = true`

The profiles are not prepared for this test, you are using the public ones. It will be the next step when this work is done to enable a switch inside the current profiles.

abdullahO2:
If you tried this, it will not work because it is more than 20 km:
https://brouter.de/brouter-test/#map=10/17.8696/46.8004/standard&lonlats=46.895592,17.601443;46.720824,17.530452;46.455709,17.968424

afischerdev:
@abdullahO2 
Thanks for the sample. This helped to find the error.
Please try the new library.

abdullahO2:
@afischerdev

Thank you for the update! I can confirm that the new library resolves the 20km limit issue. 

However, I've noticed a new, unexpected behavior. When requesting a route to a distant location, BRouter sometimes generates a direct line even when there are roads relatively nearby. It seems to bypass the step of finding the nearest road and creating a direct line from there.

Unfortunately, I'm unable to provide specific examples at the moment because the server is currently experiencing a 502 Bad Gateway error. I'll share more details as soon as the server is back online.

afischerdev:
@abdullahO2 
Server is restarted.

abdullahO2:
@afischerdev

I can now provide an example of the new issue I mentioned earlier.

With `assign use_dynamic_range = true` in the profile, please see this route:

[https://brouter.de/brouter-test/#map=8/18.436/45.675/standard&lonlats=46.720379,22.287226;46.604144,18.122351](https://brouter.de/brouter-test/#map=8/18.436/45.675/standard&lonlats=46.720379,22.287226;46.604144,18.122351)

As you can see, BRouter generates a direct line between the two points, even though there are roads relatively nearby. It doesn't seem to be finding the nearest road to the end point and creating a direct line from there, as expected. 

I hope this example helps clarify the issue.

afischerdev:
@abdullahO2 
Ok, the size for the distance check is now more expanded. And the point is marked as 'not reachable' when not inside.

The picture shows all the called data chunks in this area. In the middle, the standard call with smaller data parts is visible.
distance

All this will have new side effects - I guess. Please give it a new try.

abdullahO2:
@afischerdev

Thank you for the update! Expanding the distance check and marking unreachable points is a good step forward. 

However, while this solves the 20km limit issue, I've noticed some unexpected behavior in certain cases. Sometimes, BRouter generates a direct line to distant locations even when there are roads relatively close by. It seems like the expanded distance check might be causing it to bypass nearby roads in favor of a direct line, even when that's not the most logical route.

Here's an example that illustrates this behavior:

[https://brouter.de/brouter-test/#map=8/22.162/48.238/standard&lonlats=48.103582,23.083552;48.784866,20.802156](https://brouter.de/brouter-test/#map=8/22.162/48.238/standard&lonlats=48.103582,23.083552;48.784866,20.802156)

Compared to this more logical route:

[https://brouter.de/brouter-test/#map=8/22.159/48.235/standard&lonlats=48.103582,23.083552;48.806759,20.956385](https://brouter.de/brouter-test/#map=8/22.159/48.235/standard&lonlats=48.103582,23.083552;48.806759,20.956385)

It appears that the expanded distance check might need further refinement to ensure that it prioritizes nearby roads when generating direct lines to distant locations.

I appreciate your continued efforts to improve this feature.

abdullahO2:
@afischerdev

Thank you again for the updates! I've been reviewing the recent changes to `useDynamicRange` and have a suggestion that might further improve its behavior, particularly in cases where closer roads are bypassed in favor of a direct line to a distant location.

I've been experimenting with a modified version of the `preloadPosition` method in `NodesCache.java` that uses a tiered approach to expand the search area when `useDynamicRange` is enabled. The idea is to prioritize closer roads by starting with a smaller search radius and gradually expanding it if no suitable roads are found.

**Suggestion (generated with the help of an AI assistant):**

```java
private void preloadPosition(OsmNode n, int initialCellsize, int initialScale) {
    first_file_access_failed = false;
    first_file_access_name = null;

    int cellsize = initialCellsize;
    int scale = initialScale;

    // Try loading segments with increasing search radius until a match is found or a limit is reached
    while (waypointMatcher != null && ((WaypointMatcherImpl) waypointMatcher).useDynamicRange && !waypointMatcher.isMatched(n) && cellsize <= 1000000 / 32) { 
      for (int idxLat = -scale; idxLat <= scale; idxLat++) {
        for (int idxLon = -scale; idxLon <= scale; idxLon++) {
          if (idxLon != 0 || idxLat != 0) {
            loadSegmentFor(n.ilon + cellsize * idxLon, n.ilat + cellsize * idxLat);
          }
        }
      }

      // Increase search radius and scale for the next iteration
      cellsize *= 2; // Or any other suitable factor
      scale++; 
    }


    if (first_file_access_failed) {
      throw new IllegalArgumentException("datafile " + first_file_access_name + " not found");
    }
  }

  // Add a helper method to WaypointMatcher to check if a waypoint is matched
  interface WaypointMatcher {
    // ... other methods ...

    boolean isMatched(OsmNode waypoint);
  }

  // Implement the isMatched method in WaypointMatcherImpl
  class WaypointMatcherImpl implements WaypointMatcher {
    // ... other methods ...

    @Override
    public boolean isMatched(OsmNode waypoint) {
      for (MatchedWaypoint mwp : waypoints) {
        if (mwp.waypoint == waypoint && mwp.crosspoint != null) {
          return true;
        }
      }
      return false;
    }
  }
```
This tiered approach could potentially help ensure that BRouter always finds the most logical route, regardless of the distance to the nearest road.

Please note that this is just a suggestion generated with the help of an AI assistant. I'm not a software developer, so I might not be fully aware of all the implications of this change. I'd love to hear your thoughts on this and whether you think it's worth exploring further.

Thank you for your continued dedication to improving BRouter!

afischerdev:
@abdullahO2 
thank you very much for the suggestion and the examples. I will try it, but it will take a while. 

At first glance I think it will not work because a test for crosspoint = null is no longer true after the first point pass with dynamic distance. And preload node n is not the point the waypoint could  compete to.

Anyway we need a solution and cellsize has not been treated in this way so far. We will see.

abdullahO2:
@afischerdev

You're very welcome! I appreciate you taking the time to consider the suggestion. I understand that it might not be a straightforward implementation, and I trust your expertise in finding the best solution.

Thank you again for your valuable time and dedication to improving BRouter. Your commitment to ongoing development is truly appreciated.

afischerdev:
@abdullahO2 
Here now the test results:
The long distances beeline is generated when checking the first waypoint.
block_1

The second block doesn't contain the 2nd waypoint. The red line is from an other test to show the way.
block_2

That means to me the routine needs a max distance check to protect from long beelines.

One suggestion was to increase the cell size. This ends up in longer (uncontrolled) routes.
Current sample - red line with a max distance check:
distance

Another suggestion was to check whether a result was already available - sample above.
One problem with this was the running time doubled - in this area. In normal contexts, however, we need this test to avoid working with too much data.

My current library is installed on test server and could be used - while I clean my sources.

abdullahO2:
@afischerdev

Thank you so much for taking the time to investigate this issue further and for sharing your detailed test results. I really appreciate your dedication to finding a solution that works effectively.

I agree that implementing a max distance check is crucial to prevent the generation of excessively long beelines. I'm also open to any other ideas that you think might improve the behavior of `useDynamicRange` without negatively impacting performance.

I understand the trade-offs involved in increasing the cell size, and I agree that it's important to find a solution that balances accuracy with efficiency.

I'm happy to test your current library on the test server and provide feedback as soon as I can. 

Thank you again for your hard work and commitment to improving BRouter. It's truly inspiring!

afischerdev:
@abdullahO2 
Did you find the time to test the last published version on test server?
And if so ready for publishing?

abdullahO2:
@afischerdev

My apologies for the delayed response. I've finally had a chance to test the latest version on the test server, and I'm thrilled to report that the results are fantastic! The improvement is remarkable, and the number of unreachable locations is significantly reduced compared to previous versions. The difference is truly night and day.

I did notice that the server's response time can be slow occasionally. I'm not sure if this is related to the new feature or due to something else entirely.
firefox_30

Overall, I'm very impressed with the latest changes, and I believe this feature is ready for publishing.  Thank you again for your hard work!

afischerdev:
@abdullahO2 
I couldn't find your break. Anyway:
The available memory is very small, but should be alright for a test server.
`java -Xmx512m -Xms512m -Xmn64m ...`

abdullahO2:
@afischerdev

Thanks for looking into the server response time. I understand that the available memory on the test server is limited. I haven't experienced any significant issues affecting my testing, so I wouldn't be overly concerned about it for now.  If I encounter any further problems, I'll be sure to let you know.

devemux86:
@afischerdev Shouldn't the variable be added to other profiles too, like fastbike, etc.?

afischerdev:
@abdullahO2 
> If I encounter any further problems, I'll be sure to let you know.

That is a good way. So I will merge it.
First I added the `assign use_dynamic_range   = true ` to some profiles. 

@devemux86 
At the moment I ignored the profiles fastbike, fastbike-verylowtraffic, vm-forum-liegerad-schnell, vm-forum-velomobil-schnell because they are not really off road prepared.

abdullahO2:
@afischerdev

Great! Thanks for merging the changes.

I've encountered one more scenario where BRouter isn't finding a route, even with `use_dynamic_range` enabled.  Here's an example:

[https://brouter.de/brouter-test/#map=11/27.1120/36.0681/standard&lonlats=35.982285,27.14042;43.956041,26.153823](https://brouter.de/brouter-test/#map=11/27.1120/36.0681/standard&lonlats=35.982285,27.14042;43.956041,26.153823)

I'm getting the following error: "Error: cannot find a route for given points. Maybe try to move them closer to roads?"

Could you please take a look at this example when you have a moment?

afischerdev:
@abdullahO2 

When I add a 

`assign use_dynamic_range   = true
`

to your sample with trekking profile it works for me.
When using it with a car profile then we have more rules in the profile:
Again it has to do with access rules, but in this case the node access.

access

It works when you unable this test in node area like this:
```
assign caraccess 1
assign xcaraccess ...
```
The rule for `use_dynamic_range` is to find the nearest way. This is done without any tests for access.
And again this will not find a route when you select `avoid_toll`.
I'll try to find a way for this. But it will take a while.

abdullahO2:
@afischerdev

Thanks for investigating this and explaining the issue with access rules. I understand that finding the nearest way with `use_dynamic_range` doesn't currently consider access restrictions.

I appreciate you looking into a solution for this, especially the interaction with `avoid_toll`. I'm happy to help with testing whenever needed.  Just let me know.

afischerdev:
@abdullahO2 
I said:
> This is done without any tests for access.

This is only true for the first generation of the called waypoints. In next steps an island check is done and this breaks for the funded extra point.

Please have a look at this picture:
gate
It shows one road and two gates on it - blue line, generated with trekking profile, could be done also by the car profiles when node access is ignored, not a good solution.

In BRouter data this is not one road like in OSM. BRouter cutted this at the gates in three segments and works with them. So the blue line hits the road and can't go on because of the next segment. I didn't found a way to generate a route for this special case.

But see the easier example with [one gate](https://brouter.de/brouter-test/#map=13/27.0973/35.9917/standard&lonlats=35.976965,27.114902;43.956041,26.153823&profile=car-eco). We hit the red road - car profile will still block because of the gate. 
But now we have only on blocker and for this situation I found a workaround, the BRouter doesn't use the founded point it uses the end points of the segment - green line.

I don't think this is the end of this special beeline feature. But for now we did a small step.
It could be tested on test server.

afischerdev:
@abdullahO2
Sorry, added this in #742

abdullahO2:
@afischerdev

Thank you for the update and the follow-up on this issue!

I've tested the latest changes on the test server, and while the overall functionality seems improved, I've noticed a new behavior where BRouter generates three different routes for a single request.

Here's an example:

[https://brouter.de/brouter-test/#map=12/25.8921/46.1989/standard&lonlats=38.714447,30.773754;46.19339,25.886352](https://brouter.de/brouter-test/#map=12/25.8921/46.1989/standard&lonlats=38.714447,30.773754;46.19339,25.886352)
image
image

Could you please take a look at this when you have a chance?

afischerdev:
@abdullahO2 
Thanks for your nice example. More protection was needed. Please see #747.
Available on test server.

abdullahO2:
@afischerdev

Thank you for the update! The previous issue seems to be resolved. However, I've encountered two new problems during testing:

1. **Missing Direct Line:** In some cases, the direct line between the nearest road and the destination point is missing.  Interestingly, this issue disappears if the starting point is changed.  Here's an example:

   [https://brouter.de/brouter-test/#map=11/26.8242/45.4265/standard&lonlats=38.904941,30.760435;45.528047,26.858384](https://brouter.de/brouter-test/#map=11/26.8242/45.4265/standard&lonlats=38.904941,30.760435;45.528047,26.858384)

image

2. **Issues with Dynamic Range < 60km:** When `dynamic_range` is enabled and the distance is less than approximately 60km, there are sometimes issues with the generated route. Here are a couple of examples:
https://brouter.de/brouter-test/#map=9/30.0409/41.5720/standard&lonlats=41.458946,29.910146;40.847165,30.069095
https://brouter.de/brouter-test/#map=9/30.8999/40.6494/standard&lonlats=40.760891,30.872524;40.357041,30.404793

image


Could you please take a look at these when you have a chance?

afischerdev:
@abdullahO2 
I have to control the long distance samples - will take a while.
I'm on an other [sample](https://www.openstreetmap.org/#map=19/30.230164/43.300106) which has an isolated road and no way out of it.
So I disabled the new logic on test server.
This brings back the beeline for your first sample.

abdullahO2:
@afischerdev

Thank you for the update. I understand you're working on controlling the long-distance samples, and I appreciate the time and effort you're putting into refining this feature. It's been a long journey, and I'm grateful for your persistence.

Disabling the new logic on the test server and reverting to the beeline for the first sample seems like a reasonable temporary workaround for now.  Thank you for implementing that.  I'll continue testing and provide any further feedback as needed.

afischerdev:
@abdullahO2 
Your long distance samples should work now.
The new logic - continue at start/end of a way - is still unabled.
I try to find an other logic for that, too many problems with that.

devemux86:
@afischerdev Does the work now happen in https://github.com/abrensch/brouter/pull/747?

Or it is not valid anymore and will not be merged?

Issue #734, opened 2024-09-17
"Exit left" / "Exit right" instructions


devemux86:
Looking at the available turn instructions in [VoiceHint](https://github.com/abrensch/brouter/blob/master/brouter-core/src/main/java/btools/router/VoiceHint.java), we do not have "**Exit left**" and "**Exit right**" instructions.

They are very useful when we exit motorways. Now they are replaced by turn (slight) left / right instructions.

afischerdev:
These would be very specific instructions, but they are indeed clearer
I made a first test. Just replace keep left/right with exit left/right on leaving motorways.
Something like this:
![exit_right](https://github.com/user-attachments/assets/95829066-3faf-4c05-bce7-7cd87c1e741f)

[testtrack0.2.gpx.txt](https://github.com/user-attachments/files/17083501/testtrack0.2.gpx.txt)

The problem will be the clients and their handling of the instructions.
In the test I changed all 'exit left/right' back to 'keep left/right' to be conform with the clients. 
But I guess that is not what you want.

devemux86:
@afischerdev thanks for the example.

I do not understand what you mean with the change back?

GPX nav trk format has the advantage that is declares the instructions as strings.
Clients that already support exit instructions (like Cruiser) can use it immediately.

Nevertheless, "Exit left / right" are 2 new instructions.
So we will need new ids in `VoiceHint` and they must be included in all BRouter response types (like `geojson`).
Then clients must update their BRouter response parsing to handle the new instructions.

<img src="https://github.com/user-attachments/assets/104b5020-f632-4dda-8977-74374effa395" width="50%"/>

afischerdev:
@devemux86 
> Nevertheless, "Exit left / right" are 2 new instructions.

Yes, that is what I did. And as you see it is working for the Locus export (timode=2). But for all other clients the new tag `ER / EL` have to go back to `KR / KL`.

devemux86:
> But for all other clients

Do you mean other `turnInstructionMode`?

Also are exit instructions available in BRouter test website?

devemux86:
@afischerdev 

> But for all other clients

Do you mean other `turnInstructionMode`?

Also are exit instructions available in BRouter test website?

afischerdev:
@devemux86 
> Do you mean other turnInstructionMode?

Yes, and other exports.

> Also are exit instructions available in BRouter test website?

Sorry, no. Not at the moment, it was just a quick local test.

devemux86:
[...]

PR #742, opened 2024-11-19
New voicehints for motorway exit

afischerdev:
This is the PR for #734

The problem with this new voicehints is, not all clients are ready for using.
First rule: the export contains what the clients can read. Means

- Locus already has this tags of Exit Left / Right
- all other become the old values - we can change later

For testing select you favorite turnInstructionMode.
Please use for a direct test export the button between close and export.
When using JSON export you could select `turnInstructionMode = 2` to get the new index.

For OsmAnd exists a special `turnInstructionMode = 8`. This is not reachable via the profile dialog, you have to set this inside the profile editor. It can be used to prepare OsmAnd to the new voicehints before BRouter delivers the new tags on `turnInstructionMode = 3`.

A Test is possible on the test server: [route](https://brouter.de/brouter-test/#map=12/53.3562/10.0729/osm-mapnik-german_style&lonlats=10.085098,53.311964;10.053516,53.390419&profile=car-fast) 

@devemux86 @vshcherb @vodie @menion
Please let us know when you change your rules in your apps, so we can go on to normalize the voicehint situation.

### DIFF
--- a/brouter-core/src/main/java/btools/router/VoiceHint.java
+++ b/brouter-core/src/main/java/btools/router/VoiceHint.java
@@ -29,6 +29,8 @@ public class VoiceHint {
   static final int EL = 17; // exit left
   static final int ER = 18; // exit right
 
+  static final int END = 100; // end point
+
   int ilon;
   int ilat;
   short selev;
@@ -155,6 +157,8 @@ public class VoiceHint {
         return timode == 2 || timode == 9 ? "ER" : "KR";
       case OFFR:
         return "OFFR";
+      case END:
+        return "END";
       default:
         throw new IllegalArgumentException("unknown command: " + cmd);
     }
--- a/brouter-core/src/main/java/btools/router/VoiceHint.java
+++ b/brouter-core/src/main/java/btools/router/VoiceHint.java
@@ -26,8 +26,8 @@ public class VoiceHint {
   static final int RNLB = 14; // Roundabout left
   static final int TU = 15; // 180 degree u-turn
   static final int BL = 16; // Beeline routing
-  static final int EL = 17; // Beeline routing
-  static final int ER = 18; // Beeline routing
+  static final int EL = 17; // exit left
+  static final int ER = 18; // exit right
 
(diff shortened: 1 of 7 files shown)


devemux86:
@afischerdev 

> Please use for a direct test export the button between close and export.

Can you explain what is the difference between the middle "cloud" button and the export button?

afischerdev:
@devemux86 
The export uses the web client internal export, the cloud button exports without reworking.

devemux86:
@afischerdev 
I have seen that GPX navigation tracks can only be exported with the middle "cloud" button.
(and turnInstructionMode 2)

brouter-core/src/main/java/btools/router/VoiceHint.java
```diff
@@ -26,6 +26,8 @@ public class VoiceHint {
   static final int RNLB = 14; // Roundabout left
   static final int TU = 15; // 180 degree u-turn
   static final int BL = 16; // Beeline routing
+  static final int EL = 17; // Beeline routing
+  static final int ER = 18; // Beeline routing
```
devemux86, line 30: @afischerdev The comments here should mention the exit instructions.

devemux86:
@afischerdev Thanks for this implementation. :+1: 

Does the test server need a restart?
Routing returns "502 Bad Gateway".

afischerdev:
Just restarted

afischerdev:
Just add a logic that also find motorway exits when there are no more entries.
Therefore I added an new END voicehint type and throw it away after calculation.
This is also already available on testserver.

devemux86:
@afischerdev

Should the clients do something with this?
Or it is handled internally and is mapped to the EL / ER instructions?

afischerdev:
@devemux86 
It was only an internal update to get more exit nodes.
As before the client only has to handle the new voicehints.

devemux86:
@afischerdev 

> As before the client only has to handle the new voicehints.

You mean the 2 new EL + ER instructions.
(not the END)

Issue #743, opened 2024-11-20
Routing up a one way street.


bartp06:
Hello! I encountered an issue with routing, I was routed up a one-way street without contraflow for cyclists.
Please see the attachment:
![obraz](https://github.com/user-attachments/assets/2396970d-b823-4a5c-a507-e8ec205a611d)

I checked the tags in OSM and they seem correctly mapped to me, the affected segment has the oneway=yes tag, while the others up north indeed have a oneway:bicycle=no tag.

Here is the osm tag of the affected way 945735356.

As well as a brouter link with the way in frame for testing. https://brouter.de/brouter-web/#map=19/52.22313/21.00875/standard&lonlats=21.01558,52.212268

Thank you for reading through!

jakobwalter:
The routing works correctly as far as I can tell. You will just need to adjust the penalties. For example in the trekking profile:

`assign onewaypenalty =
       if ( badoneway ) then
       (
         if      ( cycleway=opposite|opposite_lane|opposite_track       ) then 0
         else if ( cycleway:left=opposite|opposite_lane|opposite_track  ) then 0
         else if ( cycleway:right=opposite|opposite_lane|opposite_track ) then 0
         else if ( oneway:bicycle=no                         ) then 0
         else if ( cycleway:left:oneway=no                   ) then 0
         else if ( cycleway:right:oneway=no                  ) then 0
         else if ( junction=roundabout|circular              ) then 60
         else if ( highway=primary|primary_link              ) then 50
         else if ( highway=secondary|secondary_link          ) then 30
         else if ( highway=tertiary|tertiary_link            ) then 20
         else 4.0
       )
       else 0.0`

The street in question is residential, so the added one-way penalty is just 4. If you change this to 10 you will not go through the one-way street anymore.

maciejs:
I believe that this is not handled properly.  The assumption of such a low penalty is explained by the comment posted above the cited calculation:

```
#
# handle one-ways. On primary roads, wrong-oneways should
# be close to forbidden, while on other ways we just add
# 4 to the costfactor (making it at least 5 - you are allowed
# to push your bike)
#
```

BUT, in this case it is not true, because it is not allowed to walk on this segment (because when there is a sidewalk it is obligatory to use it), so it is also not allowed to push a bicycle against traffic. In this case, the penalty should be calculated as for the forbidden segment, to possibly redirect the cyclist to the sidewalk, where it is allowed to push the bike.

Totorrr:
[...]

PR #745, opened 2024-12-04
Update trekking.brf

maciejs:
#743

### DIFF
--- a/misc/profiles2/trekking.brf
+++ b/misc/profiles2/trekking.brf
@@ -135,10 +135,9 @@ assign bikeaccess =
 # calculate logical foot access
 #
 assign footaccess =
-       if bikeaccess then true
-       else if bicycle=dismount then true
+       if bicycle=dismount then true
        else if foot= then defaultaccess
-       else not foot=private|no
+       else not foot=private|no|use_sidepath
 
 #
 # if not bike-, but foot-acess, just a moderate penalty,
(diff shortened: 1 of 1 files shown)


polyscias:
Repeating here what I wrote in #746:

I do think it is not a good idea to set the penalty for foot=use_sidepath to 10000, i.e. the same as no.

Not all Openstreetmap data is correct and especially for short segments around crossings I have seen foot=use_sidepath set incorrectly.

10000 means that brouter will never include the way, instead I suggest to set a high penalty like 100 so the algorithm will still have the option to chose it if there is foot=use_sidepath but no suitable sidepath.

Where for bicycle this is less a problem, as the fall back is pushing the bike i.e. foot with this foot change there is no fall-back.

quaelnix:
This change only affects a small subset of one-way streets, but it works well.

polyscias:
If "it works well" is your criteria, 100 also works well:

Screenshot_20241207_142814

quaelnix:
I think 10000 is acceptable in this case, but I would also agree with 100, but the routing & code example you showed belongs to: https://github.com/abrensch/brouter/pull/745

maciejs:
@polyscias @quaelnix Thank you for your suggestions, I have updated the value

quaelnix:
@afischerdev, are you ok with this change?

maciejs:
> @quaelnix I don't mind. But the result looks different to the [sample above](https://github.com/abrensch/brouter/pull/745#issuecomment-2525114880)

I'm not sure if this was an actual change proposal or just for demo, @polyscias could you clarify?

Issue #744, opened 2024-11-26
Please add respect foot=use_sidepath for pedestrian routes


maciejs:
According to [OSM Wiki](https://wiki.openstreetmap.org/wiki/Tag:foot%3Duse_sidepath#Status),

> Unlike [bicycle](https://wiki.openstreetmap.org/wiki/Key:bicycle)=[use_sidepath](https://wiki.openstreetmap.org/wiki/Tag:bicycle%3Duse_sidepath), this tag was never properly proposed and accepted. Due to its considerable use however (often in combination with [bicycle](https://wiki.openstreetmap.org/wiki/Key:bicycle)=[use_sidepath](https://wiki.openstreetmap.org/wiki/Tag:bicycle%3Duse_sidepath)), routing engines should treat it as a negative access value (i.e., equivalent to [foot](https://wiki.openstreetmap.org/wiki/Key:foot)=[no](https://wiki.openstreetmap.org/wiki/Tag:foot%3Dno) or a bit weaker) to prevent pedestrian users from being routed along ways with parallel compulsory sidepaths.  

Currently, however, a profile such as [hiking-mountain.brf](https://github.com/abrensch/brouter/blob/master/misc/profiles2/hiking-mountain.brf) does not respect this value.

afischerdev:
Do you have an example for that?

maciejs:
https://brouter.de/brouter-web/#map=19/50.13740/19.39749/standard&lonlats=19.396266,50.137436;19.397892,50.137288&profile=hiking-mountain

afischerdev:
@maciejs 
Great, thanks

polyscias:
As indicated in #745 and #746, I think it is better to use 100 instead of 10000 for foot=use_sidepath, that works also fine:

![Screenshot_20241207_142814](https://github.com/user-attachments/assets/704a29df-1842-4ebb-9007-f11fc6432a5b)

maciejs:
@polyscias Thank you for your suggestion, I have updated the value

PR #746, opened 2024-12-05
Update hiking-mountain.brf

maciejs:
#744

### DIFF
--- a/misc/profiles2/hiking-mountain.brf
+++ b/misc/profiles2/hiking-mountain.brf
@@ -164,9 +164,9 @@ assign bikeaccess
 
 assign footaccess     or any_hiking_route
                       or issidewalk
-                      or and bikeaccess  not foot=no
+                      or and bikeaccess  not foot=no|use_sidepath
                       or bicycle=dismount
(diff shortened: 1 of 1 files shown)


quaelnix:
It seems to work reasonably well.

polyscias:
I do think it is not a good idea to set the penalty for foot=use_sidepath to 10000, i.e. the same as no.

Not all Openstreetmap data is correct and especially for short segments around crossings I have seen foot=use_sidepath set incorrectly.

10000 means that brouter will never include the way, instead I suggest to set a high penalty like 100 so the algorithm will still have the option to chose it if there is foot=use_sidepath but no suitable sidepath.

Where for bicycle this is less a problem, as the fall back is pushing the bike i.e. foot with this foot change there is no fall-back.

quaelnix:
@polyscias, I was also sceptical at first, but this change will only affect the initialcost of the way and thus does not completely remove it from the routing graph and works really well as far as I can tell.

maciejs:
@quaelnix  Thank you for your suggestion, I have updated the change

quaelnix:
@polyscias, what do you think about the updated version?

Issue #749, opened 2024-12-21
get available profiles from api


pcace:
Hi there, is there any way to get the available profiles from the api? 
 - so for example https://brouter.damsy.net/api/brouter?getProfiles or something like this?
Thanks a lot!

PR #752, opened 2024-12-21
Add functionality to list available profiles and download a single profile

pcace:
This pull request adds new functionality to the BRouter server to support listing available profiles and downloading a single profile. it addresses: https://github.com/abrensch/brouter/issues/749

The following changes have been made:

**New Parameters:**

getprofiles: Returns a list of available profiles in JSON format.
getprofiles/profilename.brf: Returns the content of the specified profile.

**Updated Documentation:**

The README section in ServerHandler.java has been updated to include the new parameters and example URLs.

Example URLs:

List available profiles:
`http://localhost:17777/brouter/getprofiles`

Download a specific profile:
`http://localhost:17777/brouter/getprofiles/trekking.brf`

Please review the changes and let me know if there are any questions or further improvements needed.

Thank you!

### DIFF
--- a/brouter-server/src/main/java/btools/server/RouteServer.java
+++ b/brouter-server/src/main/java/btools/server/RouteServer.java
@@ -9,6 +9,7 @@ import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
+import java.io.FileInputStream;
 import java.net.InetAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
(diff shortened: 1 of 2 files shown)


afischerdev:
@pcace 
Thanks for your PR. 
But sorry, I don't understand the idea behind. 

When you call for the folder content you will get also profiles that are not intended for routing only for database generation (all, dummy, softaccess). And there could be some other profiles from the server owner, that are not public and so they are not in the visible list box at the server.

When you call for a single profile, what is the need to have this profile? 
When you like to use this profile on the server just call `...&profile=profile_name&...`
When you like to build your own profiles you should use the originals from this git repository. And go on with your needs.
And there are other profiles that are not reachable over the server logic - download only over github.

pcace:
Hi, no worries, i was expecting this a little bit - i was just breefly looking for the functionality to get the profiles from the server because i wanted to have the ability here: https://gpx.studio/app to use my own profiles. And since in this project the profiles are hardcoded, i thought it might be a good idea to have a way to ask the server what profiles it has. (so i changed it on my fork here: https://github.com/pcace/gpx.studio/commit/2919bec9949b2d253059eabe1caac43bc495891e)

To be honest, i had no idea that there can be profiles which are not intended for routing.

I am totally fine with you closing this here! 

cheers and have a great christmas!

zod:
Well brouter-web also needs those profiles but expects the server to provide them as download to allow customizing. There was also a similar PR in the past  (#578), but the older PR and your PR both have the same weakness: They are vulnerable to directory traversal which would allow extracting files outside of the profiles directory (or custom profiles).

In general I think it makes sense to provide a list of profiles and eventually downloading them for customization, but without those pitfalls.

PR #754, opened 2024-12-28
Updated database scripts

afischerdev:
Updated to the latest database scripts.
Thanks to @EssBee59 to prepare that.


jakobwalter:
How are we deciding which landuse is desirable? I think allotments aren't the most desirable to bike through to be honest. Furthermore, if vineyards are in there, I see no reason not to add meadow, farmland and grass. Also leisure == golf_course could be considered.

PR #755, opened 2025-01-13
Get elevation expanded, added get info

afischerdev:
The elevation function to get an elevation for a single point wasn't ready in my eyes.
Previously, only the first and last points of a segment were used. This now changes by using all segment points and searching for the nearest one.

Additionally, the routine can also provide the segment info.

It is recommended to play with the profiles to get a desired segment information.

trekking.brf
trekking 

car-eco.brf
car


devemux86:
> Additionally, the routine can also provide the segment info.

@afischerdev Is this like the "Data" tool on the BRouter website that lists the tags of route segments?

afischerdev:
@devemux86 
The tags are included in GPX and JSON format, just as a string with `tag=value ...`.

Issue #661, opened 2024-01-20
Some thoughts on rerouting – app only


afischerdev:
The app already includes a simple and effective rerouting function.
This means that only a fraction of the time is required for a recalculation

How it works:
BRouter app saves a RAW track from the last routing request.
This applies to the RAW points (way starts, way ends only), the routing destination and checksums for the profile, the Nogo instructions and the profile parameters used.

Try:
Plan a route of approx. 50 to 150 kilometers. The feeling will be something like 'this is taking a long time'.
If the starting point is now moved, the route should be finished much faster - rerouting in action.
This function can be used to start a recalculation of a running route.

Problem 1:
Reference is made to the old route. And not as expected with the 'heading=270' statement.

![fra_erf_start1](https://github.com/abrensch/brouter/assets/23175706/f7d205dd-bc12-433d-ad36-7efe362d935d) ![fra_erf_start2](https://github.com/abrensch/brouter/assets/23175706/f00529c7-afcf-4261-918a-754d8e3b9281)

This was already discussed under #589 and one solution was a weighted no-go point behind the 'old' route.
The problem with this is the rule old route is not the same as new route, the checksum of the Nogos becomes different.

The change:
This Nogo point is only formed after the route check from the direction.

Problem 2:
Maybe just in my person - when planning happens, it doesn't start straight away. Means the last saved RAW track is not valid for the route started from database.

The change:
There is a new EngineMode that only creates the RAW track. BRouter is prepared at the start time and can then be used later.
Needs no other output.

Problem 3:
The current logic only applies to a start - destination segment and is not usable when via points are used.
![fra_erf_1via](https://github.com/abrensch/brouter/assets/23175706/30d20208-0417-48bd-bb10-7599e677fa52)

The change:
Connecting the individual segments to one overall RAW track.

Test:
So that all of this can be tested, I also activated the RAW track in the BRouter class.

Example of 3 segments [start route](https://brouter.de/brouter-web/#map=8/50.495/10.344/standard&lonlats=8.693065,50.107079;9.662488,50.030603;10.327292,50.376915;11.023287,50.986153)
Rerouting shortly [after start](https://brouter.de/brouter-web/#map=10/50.1053/8.6450/standard&lonlats=8.717823,50.10046;9.662488,50.030603;10.327292,50.376915;11.023287,50.986153)
[...]

PR #759, opened 2025-01-21
New Roundtrip function

afischerdev:
Here comes the new engineMode for roundtrips.

Parameters to play with:

direction - angle around start point (default random)
roundtripDistance - distance from start point (default 1500m)
roundtripDirectionAdd - changes the angle added to the direction (default 20 degrees)
allowSamewayback - go back the same way (default false)

exportWaypoints - export the generated points (default false)
profile:correctMisplacedViaPoints - correct the results (default false)
profile:correctMisplacedViaPointsDistance - distance for correction,  0 = don'r check distance (default 40m)

some samples:

direction = 90

direction 360 correctMisplacedViaPoints off
360_nosnap

direction 360, correctMisplacedViaPoints on
360_snap

direction 180 roundtripDirectionAdd 45deg

direction 360 allowSamewayback true
360_back

### DIFF
--- a/brouter-core/src/main/java/btools/router/RoutingEngine.java
+++ b/brouter-core/src/main/java/btools/router/RoutingEngine.java
@@ -670,13 +670,25 @@ public class RoutingEngine extends Thread {
 
     switch (preferredRandomType) {
       case AreaInfo.RESULT_TYPE_ELEV50:
-        Collections.sort(ais, Comparator.comparingInt(AreaInfo::getElev50Weight));
+        Collections.sort(ais, new Comparator<AreaInfo>() {
+          public int compare(AreaInfo o1, AreaInfo o2) {
+            return o2.getElev50Weight() - o1.getElev50Weight();
+          }
+        });
         break;
       case AreaInfo.RESULT_TYPE_GREEN:
-        Collections.sort(ais, Comparator.comparingInt(AreaInfo::getGreen));
+        Collections.sort(ais, new Comparator<AreaInfo>() {
+          public int compare(AreaInfo o1, AreaInfo o2) {
+            return o2.getGreen() - o1.getGreen();
+          }
+        });
         break;
       case AreaInfo.RESULT_TYPE_RIVER:
-        Collections.sort(ais, Comparator.comparingInt(AreaInfo::getRiver));
+        Collections.sort(ais, new Comparator<AreaInfo>() {
+          public int compare(AreaInfo o1, AreaInfo o2) {
+            return o2.getRiver() - o1.getRiver();
+          }
+        });
         break;
       default:
         return (int) (Math.random()*360);
--- a/brouter-core/src/main/java/btools/router/RoutingEngine.java
+++ b/brouter-core/src/main/java/btools/router/RoutingEngine.java
@@ -490,7 +490,6 @@ public class RoutingEngine extends Thread {
       double searchRadius = (routingContext.roundTripDistance == null ? 1500 :routingContext.roundTripDistance);
       double direction = (routingContext.startDirection == null ? -1 :routingContext.startDirection);
       double directionAdd = (routingContext.roundTripDirectionAdd == null ? ROUNDTRIP_DEFAULT_DIRECTIONADD :routingContext.roundTripDirectionAdd);
-      //if (direction == -1) direction = getRandomDirectionFromRouting(waypoints.get(0), searchRadius);
       if (direction == -1) direction = getRandomDirectionFromData(waypoints.get(0), searchRadius);
 
       if (routingContext.allowSamewayback) {
@@ -503,7 +502,6 @@ public class RoutingEngine extends Thread {
         onn.name = "rt1";
         waypoints.add(onn);
       } else {
-        //buildPointsFromAngle(waypoints, direction, directionAdd, searchRadius, true);
         buildPointsFromCircle(waypoints, direction, searchRadius, routingContext.roundTripPoints == null ? 5 : routingContext.roundTripPoints);
       }
 
@@ -518,34 +516,6 @@ public class RoutingEngine extends Thread {
       logException(e);
     }
 
-  }
-
-  void buildPointsFromAngle(List<OsmNodeNamed> waypoints, double startAngle, double addAngle, double searchRadius, boolean withNogoCenter) {
-    int[] pos = CheapRuler.destination(waypoints.get(0).ilon, waypoints.get(0).ilat, searchRadius, startAngle - addAngle);
-    OsmNodeNamed onn = new OsmNodeNamed(new OsmNode(pos[0], pos[1]));
-    onn.name = "rt1";
-    waypoints.add(onn);
-
-    pos = CheapRuler.destination(waypoints.get(0).ilon, waypoints.get(0).ilat, searchRadius, startAngle + addAngle);
-    onn = new OsmNodeNamed(new OsmNode(pos[0], pos[1]));
-    onn.name = "rt2";
-    waypoints.add(onn);
-
-    onn = new OsmNodeNamed(waypoints.get(0));
-    onn.name = "to_rt";
-    waypoints.add(onn);
-
-    if (withNogoCenter) { // add a nogo area
-      pos = CheapRuler.destination(waypoints.get(0).ilon, waypoints.get(0).ilat, searchRadius/2, startAngle);
-      OsmNodeNamed n = new OsmNodeNamed();
-      n.name = "nogo" + (int) (searchRadius/3);
-      n.ilon = pos[0];
-      n.ilat = pos[1];
-      n.isNogo = true;
-      n.radius = (int) (searchRadius/3);
-      n.nogoWeight = Double.NaN;
-      routingContext.setWaypoint(n, false);
-    }
   }
 
   void buildPointsFromCircle(List<OsmNodeNamed> waypoints, double startAngle, double searchRadius, int points) {
@@ -717,141 +687,6 @@ public class RoutingEngine extends Thread {
 
     int angle = ais.get(0).direction;
     return angle - 30 + (int) (Math.random() * 60);
-  }
-
-  int getRandomDirectionFromRouting(OsmNodeNamed wp, double searchRadius) {
-
-    long start = System.currentTimeMillis();
-
-    int preferredRandomType = 0;
-    boolean consider_elevation = routingContext.expctxWay.getVariableValue("consider_elevation", 0f) == 1f;
-    boolean consider_forest = routingContext.expctxWay.getVariableValue("consider_forest", 0f) == 1f;
-    boolean consider_river = routingContext.expctxWay.getVariableValue("consider_river", 0f) == 1f;
-    if (consider_elevation) {
-      preferredRandomType = AreaInfo.RESULT_TYPE_ELEV50;
-    } else if (consider_forest) {
-      preferredRandomType = AreaInfo.RESULT_TYPE_GREEN;
-    } else if (consider_river) {
-      preferredRandomType = AreaInfo.RESULT_TYPE_RIVER;
-    } else {
-      return (int) (Math.random()*360);
-    }
-
-    MatchedWaypoint wpt1 = new MatchedWaypoint();
-    wpt1.waypoint = wp;
-    wpt1.name = "start_info";
-    List<MatchedWaypoint> listStart = new ArrayList<>();
-    listStart.add(wpt1);
-
-    List<OsmNodeNamed> wpliststart = new ArrayList<>();
-    wpliststart.add(wp);
-
-    List<OsmNodeNamed> listSearch = new ArrayList<>();
-
-    for (int a = 45; a < 360; a +=90) {
-      int[] pos = CheapRuler.destination(wp.ilon, wp.ilat, searchRadius * 2, a);
-      OsmNodeNamed onn = new OsmNodeNamed(new OsmNode(pos[0], pos[1]));
-      onn.name = "via" + a;
-
-      MatchedWaypoint wpt = new MatchedWaypoint();
-      wpt.waypoint = onn;
-      wpt.name = onn.name;
-      listStart.add(wpt);
-    }
-
-    for (int a = 0; a < 360; a +=90) {
-      int[] pos = CheapRuler.destination(wp.ilon, wp.ilat, searchRadius, a);
-      OsmNodeNamed onn = new OsmNodeNamed(new OsmNode(pos[0], pos[1]));
-      onn.name = "via" + a;
-      listSearch.add(onn);
-    }
-
-    RoutingEngine re = null;
-    RoutingContext rc = new RoutingContext();
-    String name = routingContext.localFunction;
-    int idx = name.lastIndexOf(File.separator);
-    rc.localFunction = idx == -1 ? "dummy" : name.substring(0, idx+1) + "dummy.brf";
-
-    re = new RoutingEngine(null, null, segmentDir, wpliststart, rc, BROUTER_ENGINEMODE_ROUNDTRIP);
-    rc.useDynamicDistance = true;
-    re.matchWaypointsToNodes(listStart);
-    re.resetCache(true);
-
-    int numForest = rc.expctxWay.getLookupKey("estimated_forest_class");
-    int numRiver = rc.expctxWay.getLookupKey("estimated_river_class");
-
-    OsmNode start1 = re.nodesCache.getStartNode(listStart.get(0).node1.getIdFromPos());
-
-    double elev = (start1 == null ? 0 : start1.getElev());
-
-    List<AreaInfo> ais = new ArrayList<>();
-
-    listStart.remove(0);
-    for (int a = 0; a < 4; a++) {
-      rc.ai = new AreaInfo(a * 90 +90);
-      rc.ai.elevStart = elev;
-      rc.ai.numForest = numForest;
-      rc.ai.numRiver = numRiver;
-
-      List<OsmNodeNamed> wplist = new ArrayList<>();
-      wplist.add(new OsmNodeNamed(listStart.get(a).waypoint));
-      if (a == 3) wplist.add(new OsmNodeNamed(listStart.get(0).waypoint));
-      else        wplist.add(new OsmNodeNamed(listStart.get(a+1).waypoint));
-
-      rc.ai.polygon = new OsmNogoPolygon(true);
-      rc.ai.polygon.addVertex(wp.ilon, wp.ilat);
-      rc.ai.polygon.addVertex(wplist.get(0).ilon, wplist.get(0).ilat);
-      rc.ai.polygon.addVertex(wplist.get(1).ilon, wplist.get(1).ilat);
-
-      List<OsmNodeNamed> wpsearchlist = new ArrayList<>();
-      wpsearchlist.add(listSearch.get(a));
-      if (a == 3) wpsearchlist.add(listSearch.get(0));
-      else        wpsearchlist.add(listSearch.get(a+1));
-
-      rc.useDynamicDistance = true;
-      re = new RoutingEngine(null, null, segmentDir, wpsearchlist, rc, 0);
-      rc.processUnusedTags = true;
-      rc.turnInstructionMode = 9;
-      re.doRun(0);
-
-      ais.add(rc.ai);
-    }
-
-    logInfo("execution time get area = " + (System.currentTimeMillis() - start) / 1000. + " seconds");
-
-
-    // for (AreaInfo ai: ais) {
-    //  System.out.println("\n" + ai.toString());
-    //}
-
-    switch (preferredRandomType) {
-      case AreaInfo.RESULT_TYPE_ELEV50:
-        Collections.sort(ais, new Comparator<>() {
-          public int compare(AreaInfo o1, AreaInfo o2) {
-            return o2.getElev50Weight() - o1.getElev50Weight();
-          }
-        });
-        break;
-      case AreaInfo.RESULT_TYPE_GREEN:
-        Collections.sort(ais, new Comparator<>() {
-          public int compare(AreaInfo o1, AreaInfo o2) {
-            return o2.getGreen() - o1.getGreen();
-          }
-        });
-        break;
-      case AreaInfo.RESULT_TYPE_RIVER:
-        Collections.sort(ais, new Comparator<>() {
-          public int compare(AreaInfo o1, AreaInfo o2) {
-            return o2.getRiver() - o1.getRiver();
-          }
-        });
-        break;
-      default:
-        return (int) (Math.random()*360);
-    }
-
-    int angle = ais.get(ais.size()-1).direction;
-    return angle - 45 + (int) (Math.random()*90);
   }
 
 
--- a/brouter-core/src/main/java/btools/router/RoutingEngine.java
+++ b/brouter-core/src/main/java/btools/router/RoutingEngine.java
@@ -716,7 +716,7 @@ public class RoutingEngine extends Thread {
     }
 
     int angle = ais.get(0).direction;
-    return angle - 45 + (int) (Math.random()*90);
+    return angle - 30 + (int) (Math.random() * 60);
   }
 
   int getRandomDirectionFromRouting(OsmNodeNamed wp, double searchRadius) {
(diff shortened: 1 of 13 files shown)


devemux86:
@afischerdev Thanks for this long awaited feature! :+1: 

> roundtripDirectionAdd - add an angle to the direction (default 20degree)

Can you explain more how this parameter works?
Does it add an angle _tolerance_ to the direction (± 20°)?

---

Some routers have a seed parameter to get different tours.
See the [GraphHopper documentation](https://docs.graphhopper.com/openapi/routing/getroute#routing/getroute/t=request&in=query&path=round_trip.seed).

Is this something to consider for BRouter or it generates the same route for the same parameters?

devemux86:
> roundtripDistance - distance from start point (default 1500m)

@afischerdev What is this distance? Not length of the route (as in other routers)?
Distance of the start point from the longest point of the route or something else?

afischerdev:
@devemux86 
> Does it add an angle tolerance to the direction (± 20°)?

may be better: It replaces the default of 20 with e.g. 45 degree.

> Is this something to consider for BRouter or it generates the same route for the same parameters?

Both are correct.
If you define a `direction` of 180 degrees, it will be used for all calls.
If you do not define the `direction` value, however, a random `direction` will be generated. Therefore, you will get different routes for each call with the same parameter.

devemux86:
@afischerdev Is it also available in BRouter test server?

afischerdev:
@devemux86 
> What is this distance?

We have two via point and this is the distance start point to via point.

> Is it also available in BRouter test server?

No, I could port it, but we currently don't have a web client that could take advantage of this new feature.
May be later on ;-)

devemux86:
> direction - angle around start point (default random)
roundtripDirectionAdd - changes the angle added to the direction (default 20 degrees)

@afischerdev So, if we set `direction` 90 degrees, why would we want to use `roundtripDirectionAdd` and change the direction again to a different value?
What is its usefulness?

afischerdev:
@devemux86 
> why would we want to use roundtripDirectionAdd

The direction is something like a main angle we want to place the route. E.g. North of parking space
But we want to have a more or less round trip and with the `roundtripDirectionAdd` we open the left and right area of the direction.

Please see:
360_snap_angle

devemux86:
> with the `roundtripDirectionAdd` we open the left and right area of the direction.

@afischerdev Thanks, now it is clear! :+1:

devemux86:
> with the `roundtripDirectionAdd` we open the left and right area of the direction.

@afischerdev Thanks, also for the image, now it is clear! :+1:

devemux86:
> allowSamewayback - go back the same way (default false)

@afischerdev It would be interesting if this option could also be available outside of round trips in normal route planning.

devemux86:
> direction 360, correctMisplacedViaPoints on

@afischerdev What `correctMisplacedViaPointsDistance` did you use in the 3rd image?

Even if we use a large `correctMisplacedViaPointsDistance`, we may have via points that go back and forth on the same road with **u-turns** that we want to avoid.

Is it because the distance is a strict requirement in the calculation?
Or can calculate a bit longer round trips and avoid such u-turns?

devemux86:
> Even if we use a large `correctMisplacedViaPointsDistance`, we may have via points that go back and forth on the same road with u-turns that we want to avoid.

@afischerdev I mean if we can avoid round trips like the following with u-turns on the via points:

afischerdev:
Some mere results:

The current version (add angle = 20, no misplaced correction) produces this routes (direction 0 to 315, step 45 deg):



the new version runs with an add angle of 45, missplaced correction and a nogo point in angles center



Not perfect but a little bit better results.

Source coming soon. I will change the new same way back function.

afischerdev:
The last commits enabled a same way back function for both: normal routing and round trip routing. This will also work when via points are used. But it doesn't necessarily have to be the same route back.
And it changes the defaults: angle add 45 deg, more points to go back a line (snapp when two lines are concatenated).

devemux86:
@afischerdev I tried the `allowSamewayback` in normal routing and it seems to do strange things.

allowSamewayback=1 should be like the old routing where we allow the same way back?
Then allowSamewayback=0 is the new option we meant to implement in normal routing?
But checking the code if allowSamewayback=1 then it adds no-go areas, which is wrong?

devemux86:
@afischerdev Also it seems the waypoints (by user) are modified by the routing engine?

e.g. we have 2 user waypoints Start + End on a highway with distance a few kilometers.
In `RoutingEngine` line 207 a new waypoint (with radius) is added with the start's coordinates.
So the route goes start > end > start with the double length of the expected route.

afischerdev:
@devemux86 
> allowSamewayback=1 should be like the old routing where we allow the same way back.
Then allowSamewayback=0 is the new option we meant to implement in normal routing.

May be I didn't understand `allowSamewayback`. 
My idea was to go back a route, that means we have waypoints at start
```
wp from
wp to
```
with same way back on we have
```
wp from
wp via1 (the old 'to')
wp to (same as from)
```

>  if allowSamewayback=1 then it adds no-go areas

No, the nogo point is added only for roundtrip.

devemux86:
> My idea was to go back a route

Sorry, that's not what I meant. And that's the app's job to design a go back route.

What I mean is an option that in normal routing can avoid using the same roads.
For example in above [image](https://github.com/abrensch/brouter/pull/759#issuecomment-2609331262), it does not u-turn but continues forward and returns through a different road.

---

@afischerdev What is a waypoint with radius but with `isNogo` false?
If it is not a no-go, what is it?

afischerdev:
@devemux86 
>  avoid using the same roads

This is done by using 

```
profile:correctMisplacedViaPoints=1 - switch the check on
profile:correctMisplacedViaPointsDistance=250  - stops going back on around 250 m, =0 don't stop, the limit for go back points is at the moment MAX_STEPS_CHECK = 250
```

allowSamewayback=1 - Ok, if not like, just don't use.

afischerdev:
@devemux86 
> What is a waypoint with radius but with isNogo false?
If it is not a no-go, what is it?

The standard for a waypoint is  isNogo=false. Please see the [OsmNodeNamed](https://github.com/abrensch/brouter/blob/master/brouter-core/src/main/java/btools/router/OsmNodeNamed.java)

This class is used for both nogo and not nogo points. And we start with a list of OsmNodeNamed for the waypoint collection.

brouter-core/src/main/java/btools/router/RoutingEngine.java
```diff
@@ -447,6 +478,79 @@ public void doGetInfo() {
...
+        waypoints.add(onn);
+      } else {
+        int[] pos = CheapRuler.destination(waypoints.get(0).ilon, waypoints.get(0).ilat, searchRadius, direction-directionAdd);
+        MatchedWaypoint wpt2 = new MatchedWaypoint();
+        wpt2.waypoint = new OsmNode(pos[0], pos[1]);
+        wpt2.name = "via1_" + (int) (direction-directionAdd);
+
+        OsmNodeNamed onn = new OsmNodeNamed(new OsmNode(pos[0], pos[1]));
+        onn.name = "via1";
+        onn.radius = 1000;
```
devemux86, line 512: @afischerdev Here why do we set a `radius`?
As this a via point, not a no-go.

afischerdev: @devemux86 
The radius is also used for distance to the nearest road calculation.
I think it is from the beginning of this project, where the dynamic distance wasn't there and I tried to have a slightly larger radius.
But this is redundant now. I will remove it

devemux86:
@afischerdev So the recommendation when generating round trips is the following?
```
profile:correctMisplacedViaPoints=1
profile:correctMisplacedViaPointsDistance=250
```

afischerdev:
@mcliquid 
I think I jumped to short with my answer. The profile can help with better route finding for elevation or 'green area'.
But in this round trip calculation the random start brings you anywhere. And the profile has no chance to find a proper route when you start e.g. in an East direction and the forest is in the West.

The data for your wishes is available in the routing tiles. But they are only out when a route is calculated. This could mean that routing in all directions is necessary to find an optimum. I don't like that very much. So I hope to find a shorter way.

afischerdev:
@devemux86 
Yes, same rule as before. But I would use `profile:correctMisplacedViaPointsDistance=0` to get out the maximum points.

afischerdev:
@devemux86 
I'll have a look.

afischerdev:
@devemux86 
This should bring better results. But it is not yet optimal. I hope with a 'green filter' it will start more in the North.

Last version:
190_beeline

New version
190_nobeeline

afischerdev:
In this version I changed to a new function for finding travelling points. More like a semi circle, with 4 extra points and without nogo in center.

0-315_new_circle

afischerdev:
There is also an initial idea for a weighted starting direction for the round trips.
First build four points around the starting point.
start
Then create a route between these points to get the route segments and description -  just first path finding is enough.
And collect some values from the description. Something like this (area direction, start point elevation, counted ways):
```
Area 90 210.5m ways 685
Area ways <50m  518 75%
Area ways <100m 634 92%
Area ways more  51 7%
Area ways green 294 42%
Area ways river 0 0%

Area 180 210.5m ways 745
Area ways <50m  451 60%
Area ways <100m 708 95%
Area ways more  37 4%
Area ways green 336 45%
Area ways river 0 0%

Area 270 210.5m ways 967
Area ways <50m  546 56%
Area ways <100m 957 98%
Area ways more  10 1%
Area ways green 533 55%
Area ways river 634 65%

Area 360 210.5m ways 751
Area ways <50m  636 84%
Area ways <100m 751 100%
Area ways more  0 0%
Area ways green 369 49%
Area ways river 27 3%
```
Now play with the results. Looking for the user selected profile parameters, like `consider_elevation, consider_forest, consider_river`

This has its limits when working with the Azores example, the space becomes very tight.
start2

All this works a little bit against timing.
And I see a problem in keeping this data ready for the next routing call and not calling the tests again.

More work is needed, so no source here at the moment.
Ideas are welcome.

devemux86:
> First build four points around the starting point.

@afischerdev Interesting idea. So instead of starting at a location and do a tour around returning again to the start.
_We start at the center of the tour_, go to point 1, do a tour 1-2-3-4-1 and then return from 1 to the start at the center?

How about the calculation speed?
Currently (with the no-go area) calculating large round trips (> 300 km) can take a lot of time.

Also is this to become an extra algorithm option?
We should not remove current round trips, because this is how users know them from other routers and expect to work.

afischerdev:
@devemux86 
Sorry, not that idea. 
Just a pre process before round trip routing starts. The idea was to find the direction not from random 0 - 360 degrees but from more green or less high direction, e.g. random 315 to 45 deg for a more green area in the North.

devemux86:
> In this version I changed to a new function for finding travelling points. More like a semi circle, with 4 extra points and without nogo in center.

@afischerdev This version seems faster and can produce more circular routes.

Can we have the extra points number as an parameter in the API?

[OpenRouteService](https://maps.openrouteservice.org/) has a similar parameter with documentation:
_"**Points**: The number of points to use on the route. Larger values create more circular routes."_

devemux86:
> Sorry, not that idea.

@afischerdev No problem. :slightly_smiling_face: 
I didn't mean to change the implementation, just to understand if the round trip concept changes.

afischerdev:
@devemux86 
> Can we have the extra points number as a parameter in the API?

Yes, that should be no problem

devemux86:
@afischerdev 

> assign add_beeline          = true    # default is false

Can this new parameter from https://github.com/abrensch/brouter/issues/762 be added to this pull requests or separately?

afischerdev:
@devemux86 
I've added it in a separate request. Seems to be clearer.

afischerdev:
The next update will bring a first function to detect the surroundinga of a round trip area.
How this works:
To determine the optimal routing direction, four quadrants are examined. The collection counts the ways for the conditions of forest, river or an elevation difference less than 50 meters. And when you are looking for e.g forest, so the quadrant with the most way of forest information comes back with a random value of its angles.
test360
This is the North direction and will deliver back an angle between 315 to 45 deg. And with this direction the round trip generation could be done.

This is not ready and comes with some problems.
- At the moment the examination is done on every routing call. There should be a temporary save of the results.
- It works well in a shorter distance range (< 5 km = ways around 15 km), but it becomes less accurate and slower with larger ranges.

brouter-core/src/main/java/btools/router/RoutingEngine.java
```diff
@@ -559,6 +564,126 @@ void buildPointsFromCircle(List<OsmNodeNamed> waypoints, double startAngle, doub
...
+    logInfo("execution time get area = " + (System.currentTimeMillis() - start) / 1000. + " seconds");
+
+
+    // for (AreaInfo ai: ais) {
+    //  System.out.println("\n" + ai.toString());
+    //}
+
+    switch (preferredRandomType) {
+      case AreaInfo.RESULT_TYPE_ELEV50:
+        ais.sort(Comparator.comparingInt(AreaInfo::getElev50Weight));
```
devemux86, line 672: @afischerdev `List.sort` requires Android 7.

Better use `Collections.sort` to be compatible with all Android versions.

brouter-core/src/main/java/btools/router/RoutingParamCollector.java
```diff
@@ -206,6 +206,11 @@ public void setParams(RoutingContext rctx, List<OsmNodeNamed> wplist, Map<String
           rctx.roundtripDistance = Integer.valueOf(value);
         } else if (key.equals("roundtripDirectionAdd")) {
           rctx.roundtripDirectionAdd = Integer.valueOf(value);
+        } else if (key.equals("roundTripPoints")) {
```
devemux86, line 209: @afischerdev The other round trip parameters have "roundtrip" in lower case:
`roundtripDistance` and `roundtripDirectionAdd`.

The `roundTripPoints` has the "trip" starting in upper case. We could make all of them similar?

brouter-core/src/main/java/btools/router/RoutingEngine.java
afischerdev: Ok, I'll change

brouter-core/src/main/java/btools/router/RoutingEngine.java
```diff
@@ -669,13 +670,13 @@ int getRandomDirectionByRouting(OsmNodeNamed wp, double searchRadius) {
 
     switch (preferredRandomType) {
       case AreaInfo.RESULT_TYPE_ELEV50:
-        ais.sort(Comparator.comparingInt(AreaInfo::getElev50Weight));
+        Collections.sort(ais, Comparator.comparingInt(AreaInfo::getElev50Weight));
```
devemux86, line 673: @afischerdev `Comparator.comparingInt` also requires Android 7.

devemux86: @afischerdev BRouter app has `minSdkVersion 21` so IDEA should inform for all these?

afischerdev: Using CLI with gradlew for building my test lib.
In the Android Studio same source, `comparingInt` is marked (api 24). When changed the Comparator, the `Collections.sort` is marked and recommended to use List.sort(). After doing this, it's marked as wrong api 21.
Intelligent development systems are sometimes funny.

devemux86:
@afischerdev  Did you find better results with `buildPointsFromAngle` or `buildPointsFromCircle`?

Will there be a parameter to select the calculation method?

afischerdev:
@devemux86 
I prefer the circle method, the chance to get a round and not a spider route seems to be better. So at the moment no plans for a switch.

But I'm working on a second method to get the starting angle from direct reading database. Has a better timing on small ranges but takes its time on bigger rounds. Both not optimal.

devemux86:
@afischerdev Thanks, the biggest complaint I've seen from users who have tried it,
is the u-turns that are created at some via points, that go and return at the same road.

Some routers have a parameter to force go forward on the via points and not return with u-turn.
I don't know if we can solve this somehow with BRouter engine?

afischerdev:
@devemux86 
Hm, interesting thing.
We can cut this at the moment, but go on is a new idea. I'll see how we can use it. But will take a while.

afischerdev:
@devemux86 
I prefer the circle method, the chance to get a round and not a spider route seems to be better. So at the moment no plans for a switch.

But I'm working on a second method to get the starting angle from direct reading database. Has a better timing on small ranges but takes its time on bigger rounds. Both not optimal.

afischerdev:
Added a new method of collecting area values. It reads direct from rd5 files.
But as you will believe, it take its time:
- ~ 0.1 seconds on 1500 m radius
- ~ 0.5 seconds on 15000 m
- ~ 3.0 seconds on 50000 m

Here the scanned areas for 15000 m
rt_15000

It is not ready at the moment and needs optimization. 
But I would like to offer it for discussion.

devemux86:
@afischerdev Something seems wrong in `getRandomDirectionFromData` and `getRandomDirectionFromRouting`?
```java
rc.localFunction = "dummy";
```
This is without full path, so after calling `new RoutingEngine` then in `ProfileCache.parseProfile` line 38 the:
```java
profileDir = new File(rc.localFunction).getParentFile();
```
returns `null` and then the `lookups.dat` file cannot be found, so route calculation fails.

devemux86:
@afischerdev Perhaps it's better to use:
```java
rc.localFunction = routingContext.localFunction;
```

afischerdev:
@devemux86 
Thanks for the test.
Are you sure you have `dummy.brf` on board? It is not excluded [here](https://github.com/abrensch/brouter/blob/c72fc5f74df5a380b9d7121b53a6b92089f4acf1/brouter-routing-app/build.gradle#L139) and the BRouter apk should have it in profile folder.
I used the dummy profile just to get ways and there tags. Conditional selection is not needed here - I think.

devemux86:
@afischerdev If have not built the BRouter app.
I use directly the code including this pull request.

I have the `dummy.brf`. The routing and get elevation work.
The round trip does not work with this random direction call.

devemux86:
@afischerdev If have not built the BRouter app.
I use directly the code including this pull request.

I have the `dummy.brf`. The routing and get elevation work.
The round trip does not work with this random direction call.

But it works if we use:
```java
rc.localFunction = routingContext.localFunction;
```

afischerdev:
@devemux86 
Thanks again for the pointer. I'd like to stay with the dummy profile.

afischerdev:
@devemux86 
> Some routers have a parameter to force go forward on the via points and not return with u-turn.

Now we go on after a round trip via point. 
But it doesn't protect in all cases, so a `correctMisplacedViaPoints` is still recommended.

devemux86:
@afischerdev This works much better now! :+1:

devemux86:
> Now we go on after a round trip via point.

@afischerdev Can we have the avoid u-turns on via points as an option in normal A-B routing as well,
as other routers provide?

afischerdev:
Just to clear the situation a little bit here two routes without snapping function. The red one has no new 'go on function', the underlying blue one shows in point 1 and 2 (ends in a roundabout) the result of going forward on next node.
And the result of snapping function in an extra picture.
rt_nosnap rt_snap

afischerdev:
@devemux86 
> Can we have the avoid u-turns on via points as an option in normal A-B routing as well, as other routers provide?

I think it is possible, but I'm not sure if it is a good idea. It will bring you away from your original route in an other direction. `correctMisplacedViaPoints` is still my friend.

devemux86:
@afischerdev 

> `correctMisplacedViaPoints` is still my friend.

But this only works when calculate route in 1 step with all via point included? Which can be slow for long routes.
BRouter website calculates much faster by doing separate calculations for each route section between via points.

devemux86:
> I think it is possible, but I'm not sure if it is a good idea. It will bring you away from your original route in an other direction.

@afischerdev It will be a user option, only activated if users really want to avoid u-turns.

Some routers provide it as a global on/off option, other routers provide it as option on each via point.

devemux86:
@afischerdev I included BRouter round trip function in the latest Cruiser (Beta):
- https://github.com/devemux86/cruiser/discussions/734

Users started testing BRouter round trips and we have a report about "target island detected for section":

- https://github.com/devemux86/cruiser/discussions/707#discussioncomment-12487022

afischerdev:
@devemux86 
> BRouter website calculates much faster by doing separate calculations for each route section between via points.

How do you calculate that?
Here a small example to [verify your statement](https://brouter.de/brouter-web/#map=13/49.9729/9.3505/standard&lonlats=9.300059,49.973627;9.3263,49.972918;9.350897,49.974574;9.375695,49.969383)
When I test this, the brouter lib is called 3 times with around 30 ms per call.
When I now export a json file and use the cloud button (one call for the lib) time is ~ 60 ms.

BRouter does the same as BRouter web, it calculates three segments and put them together. But it has only one call to load the profile.

afischerdev:
Here the example result as picture.
You see for the 3rd point a short  u-turn. when you go on here you has a loop as the new result.
no_snap

afischerdev:
> target island on segment

Means you are on a point that has no connection to the others - like on an island.

devemux86:
> Means you are on a point that has no connection to the others - like on an island.

@afischerdev Thanks, I understand that this is the technical reason.

However users see a failed round trip calculation. Can we improve this in any way?

afischerdev:
@devemux86 
> he 2nd time the combined result is always faster

Yes, BRouter web sends only the two segements again where the via point moves.
But it could also send the two new segments in one request (having the missplaced correction on the via point) and merge this to the old route.
We discussed that for the library in #661 and stopped #674. But its still an option.

afischerdev:
@devemux86 
> However users see a failed round trip calculation. Can we improve this in any way?

I'll try that. But I need a sample.
This [example](https://brouter.de/brouter-test/#map=15/52.4668/13.1975/standard&lonlats=13.192105,52.452584;13.191358,52.46811;13.196981,52.484887) on the test server shows a correct result, no error message - it has `allow_ferries = false`.
When you use this on the BRouter web you will get an error message - but not the target island exception.

devemux86:
@afischerdev Round trips already use dynamic range, so they have extended search range.

If "target island" is detected, perhaps the algorithm can continue and snap the via point to the next nearby valid road?

afischerdev:
@devemux86 
> If "target island" is detected, perhaps the algorithm can continue and snap the via point to the next nearby valid road?

Yes, that is what you see on the test server. 
The nearst road is on the island, but that is not connected to others. So the via crosspoint (the used for routing point) is on land.

afischerdev:
There is a small improvement for collecting area data.
On bigger collection the center is ignored and the border of the quadrants.

with 50.000 m or 15.000 m
50_save 15_save

afischerdev:
Another small improvement for collecting area data.
Save and read the data search results for the area info. 
If there are no significant changes, the old results are reused. 
This will only work for the BRouter app.

brouter-routing-app/src/main/java/btools/routingapp/BRouterWorker.java
```diff
@@ -46,6 +46,7 @@ public String getTrackFromParams(Bundle params) {
 
     RoutingContext rc = new RoutingContext();
     rc.rawTrackPath = rawTrackPath;
+    rc.rawAreaPath = rawTrackPath.substring(0, rawTrackPath.lastIndexOf(File.separator)+1) + "rawAreaInfo.dat";
```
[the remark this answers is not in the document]
afischerdev: @devemux86 
Thanks for the report. Hm, on a short test I had no problem with getinfo. 
But it looks like if would be nice to have a check for null here.

afischerdev:
Next improvement: Collect area info per rd5 file. Helpful when the data tiles overlap, e.g. round trip at 10.001 / 49.998.

afischerdev:
The result angle of the random function seems to be a little bit to often near the border to the next quadrant. Something like shown here
rt_15000
So I pushed it away a little.

afischerdev:
I would like to close this. 
But first, the source code needs to be cleaned up:

Could we drop the [getRandomDirectionFromRouting](https://github.com/afischerdev/brouter/blob/ea3399d761ce705c40748c46790a70995c5e2f89/brouter-core/src/main/java/btools/router/RoutingEngine.java#L493) function?
And could we drop the [buildPointsFromAngle](https://github.com/afischerdev/brouter/blob/ea3399d761ce705c40748c46790a70995c5e2f89/brouter-core/src/main/java/btools/router/RoutingEngine.java#L506C11-L506C31) function as well?

Or are there other ideas for a round-trip strategy?

Issue #762, opened 2025-02-16
Change of the meaning for 'dynamic range'


afischerdev:
The dynamic range was introduced with PR #729 and #747 and enabled the issue #387 - this also carries the description.

The work on the round trip (#759) shows that the dynamic range could be useful in other contexts as well.

But this reduces the meaning to the original meaning of the words: 
`find the next point that is connected to a way`.

The current meaning has the addition 
`and add a beeline`

To play with the problem I added a test library to the test server.

New is the parameter for the profile in global section:

```
assign add_beeline          = true    # default is false
```

Please see this [sample 2 points](https://brouter.de/brouter-test/#map=12/37.6907/-25.4990/standard&lonlats=-25.506737,37.719793;-25.522968,37.645937) or [sample 3 points](https://brouter.de/brouter-test/#map=14/37.6985/-25.5252/standard&lonlats=-25.506737,37.719793;-25.506955,37.684186;-25.521312,37.723427)

Now if you add the new beeline parameter above, you will see it in the result.

To get all examples used in #387 running there is one more parameter:

```
# car profile
assign use_offroad          = true       # default is false
```
This is because we used `use_dynamic_range` also to enable some other roads
```
assign caraccess ...
    switch and highway=track use_offroad     1
    switch and highway=road use_offroad      1
```
Should be clearer now.

devemux86:
@afischerdev Thanks, very nice improvement. 👍 

Will you also include it in a pull request?

afischerdev:
@devemux86 
Yes, that is the plan. Because it is also needed for roundtrip #759 to avoid ambiguity.

PR #764, opened 2025-02-25
Change dynamic range meaning

afischerdev:
Update for #762 to enable dynamic range for all waypoint matches.

### DIFF
--- a/brouter-core/src/main/java/btools/router/RoutingContext.java
+++ b/brouter-core/src/main/java/btools/router/RoutingContext.java
@@ -77,7 +77,8 @@ public final class RoutingContext {
   public boolean correctMisplacedViaPoints;
   public double correctMisplacedViaPointsDistance;
   public boolean useDynamicDistance;
-
+  public boolean buildBeelineOnRange;
+  
   private void setModel(String className) {
     if (className == null) {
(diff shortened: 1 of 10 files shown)


devemux86:
@afischerdev Very nice improvement! :+1: 

With `use_dynamic_range` now default on and `add_beeline` off,
route planning with BRouter is now much better and easier without the annoying "marker is too far from a route".

devemux86:
@afischerdev And now we don't have to set a large value in `waypointCatchingRange`.

So the `waypointCatchingRange` is now only useful when `use_dynamic_range` is off?

afischerdev:
@devemux86 
This variable is still in use.
As before the beelinie is added when the distance is more then `waypointCatchingRange`
E.g. [this](https://brouter.de/brouter-test/#map=17/37.70738/-25.50812/standard&lonlats=-25.517515,37.721379;-25.506798,37.706501) against [this](https://brouter.de/brouter-test/#map=17/37.70738/-25.50812/standard&lonlats=-25.517515,37.721379;-25.506615,37.706242). both are with 
`assign add_beeline true
`

devemux86:
@afischerdev Thanks for the examples, now it is clear. :+1:

Issue #773, opened 2025-04-01
Prevent u-turns on via points


devemux86:
Some routers have a parameter to force go forward on the via points and not return with u-turn.

I don't know if this is possible to implement in BRouter engine?

_(this idea was originally posted in round trip pull request https://github.com/abrensch/brouter/pull/759)_

devemux86:
`correctMisplacedViaPoints=1` and `correctMisplacedViaPointsDistance=0` can help in round trips.

But in regular routing it can change the result, like removing the large section with the via point (image 2).

While we really want to go to the via point, but prevent the u-turn and instead continue forward,
like the result when adding the 2nd via point (image 3).

A discussion about this with a gpx route example exists here:
- https://github.com/devemux86/cruiser/discussions/759

<img src="https://github.com/user-attachments/assets/d5dbabc6-ee74-4363-9b28-f8f991a9c1d6" width="33%"/><img src="https://github.com/user-attachments/assets/a9ab9cb9-1cad-49f2-a3f2-b7bd1b9f782e" width="33%"/><img src="https://github.com/user-attachments/assets/2112d046-7743-4fb5-8822-6d33690b8c2c" width="33%"/>

devemux86:
> Some routers have a parameter to force go forward on the via points and not return with u-turn.

Here is the documentation of such a parameter from other routers:
- **GraphHopper**: If true, u-turns are avoided at via-points.

- **OpenRouteService**: Forces the route to keep going straight at waypoints restricting uturns there even if it would be faster.

- **OSRM**: Forces the route to keep going straight at waypoints constraining uturns there even if it would be faster.

afischerdev:
@devemux86 
> I don't know if this is possible to implement in BRouter engine?

I still think it's possible, but I'm not sure if it's a good idea to use it.
Here my results:
![Image](https://github.com/user-attachments/assets/bae69398-eb4f-4f35-9af8-d28e6205dc29)
I used the default BRouter car profile go get this. And as you can see, the route is almost the same as in the example.

There must be strange rules in the other profile to cause such a strong detour. I would recommend to have a look here first.

Other [example](https://brouter.de/brouter-web/#map=15/49.9734/9.3577/standard&lonlats=9.307661,49.973215;9.350314,49.974806;9.390043,49.963253) from #759:
A `go on rule` here would bring an extra loop - I can't believe that is intended.
[...]

PR #774, opened 2025-04-02
Added a switch to avoid u-turns

afischerdev:
As discussed in #773, here comes the change.

It could be activated by `profile:continueStraight=1` or `assign continueStraight = true`.

The default is false.
For round trips this was the default setting, but now you can decide.

### DIFF
--- a/brouter-core/src/main/java/btools/router/RoutingEngine.java
+++ b/brouter-core/src/main/java/btools/router/RoutingEngine.java
@@ -995,7 +995,7 @@ public class RoutingEngine extends Thread {
       } else {
         seg = searchTrack(matchedWaypoints.get(i), matchedWaypoints.get(i + 1), i == matchedWaypoints.size() - 2 ? nearbyTrack : null, refTracks[i]);
         wptIndex = i;
-        if (engineMode == BROUTER_ENGINEMODE_ROUNDTRIP) {
+        if (routingContext.continueStraight) {
           if (i < matchedWaypoints.size() - 2) {
             OsmNode lastPoint = seg.containsNode(matchedWaypoints.get(i+1).node1) ? matchedWaypoints.get(i+1).node1 : matchedWaypoints.get(i+1).node2;
             OsmNodeNamed nogo = new OsmNodeNamed(lastPoint);
--- a/brouter-core/src/main/java/btools/router/RoutingEngine.java
+++ b/brouter-core/src/main/java/btools/router/RoutingEngine.java
@@ -1137,7 +1137,12 @@ public class RoutingEngine extends Thread {
         indexfore++;
 
         if (routingContext.correctMisplacedViaPointsDistance > 0 &&
-          wayDistance > routingContext.correctMisplacedViaPointsDistance) break;
+          wayDistance > routingContext.correctMisplacedViaPointsDistance) {
+          removeVoiceHintList.clear();
+          removeBackList.clear();
+          removeForeList.clear();
+          return false;
+        }
       }
 
 
--- a/brouter-core/src/main/java/btools/router/RoutingContext.java
+++ b/brouter-core/src/main/java/btools/router/RoutingContext.java
@@ -77,6 +77,7 @@ public final class RoutingContext {
   public double waypointCatchingRange;
   public boolean correctMisplacedViaPoints;
   public double correctMisplacedViaPointsDistance;
+  public boolean continueStraight;
   public boolean useDynamicDistance;
   public boolean buildBeelineOnRange;
 
@@ -125,6 +126,8 @@ public final class RoutingContext {
 
     correctMisplacedViaPoints = 0.f != expctxGlobal.getVariableValue("correctMisplacedViaPoints", 1.f);
     correctMisplacedViaPointsDistance = expctxGlobal.getVariableValue("correctMisplacedViaPointsDistance", 0.f); // 0 == don't use distance
+
+    continueStraight = 0.f != expctxGlobal.getVariableValue("continueStraight", 0.f);
 
     // process tags not used in the profile (to have them in the data-tab)
     processUnusedTags = 0.f != expctxGlobal.getVariableValue("processUnusedTags", 0.f);


afischerdev:
For testing use on the test server this [example](https://brouter.de/brouter-test/#map=15/49.9735/9.3487/standard&lonlats=9.307661,49.973215;9.350314,49.974806;9.390043,49.963253). But you will not see a result on screen. BRputer web calls the route segment by segment. 
Only in export (cloud symbol at center) the result is visible.

devemux86:
@afischerdev Thanks, this is a very nice improvement! :+1:

afischerdev:
One more [example](https://brouter.de/brouter-test/#map=18/49.61052/8.85700/standard&lonlats=8.858837,49.610223;8.857131,49.610614;8.855957,49.611008):
This is a dead end street. The route finds the way out.
As above edit the profile und export route for control.

devemux86:
Another example:
Avoid u-turns: Off|Avoid u-turns: On
-|-
|

0709wiwiwi:
@afischerdev @devemux86 
Just had a very short fast test. (I did not further study this case).  (Today other things to do.)

https://brouter.de/brouter-test/#map=17/49.60962/8.85634/standard&lonlats=8.858757,49.609236;8.857244,49.608596;8.855957,49.611008

Download the navigation track.
Misses the expected in out traject plus expected u-turn.
(I think (quick test) the u-turn is out of the max 100m Misplaced Via points range tool.)
[Wahlen 0.6km (2).gpx.txt](https://github.com/user-attachments/files/19599033/Wahlen.0.6km.2.gpx.txt)

devemux86:
@afischerdev 

`correctMisplacedViaPointsDistance` is now default `0`, [RoutingContext.java#L127](https://github.com/abrensch/brouter/blob/master/brouter-core/src/main/java/btools/router/RoutingContext.java#L127).
Also in test website? i.e. always correct without distance.

Why this change from the previous default `40 m`?
This can lead to unwanted removal of all side route sections, as seen [above](https://github.com/abrensch/brouter/pull/774#issuecomment-2777649867) and below:

devemux86:
@afischerdev 

`correctMisplacedViaPointsDistance` is now default `0`, [RoutingContext.java#L127](https://github.com/abrensch/brouter/blob/master/brouter-core/src/main/java/btools/router/RoutingContext.java#L127).
Also in test website? i.e. always correct without distance.

Why this change from the previous default `40 m`?
This can lead to unwanted removal of all side route sections, as seen [above](https://github.com/abrensch/brouter/pull/774#issuecomment-2777649867) and below:

Without correction|With correction
-|-
|

afischerdev:
The intention of `correctMisplacedViaPoints` is to have a clear route without any detours. So what is the benefit of 40 meter detour? 
If you like to have a stop at the via point it is not a misplaced one. You could add a label to that point and have only the other protected, like `...;8.857244,49.608596,"stop";...`

via_c_40 via_c_stop

We could define it the other way round, a point more then 40m is not a missplaced point. This depends on the zoom level at which you plan a route.

afischerdev:
Small change: the correctMisplacedViaPointsDistance works now as a tolerance value. If ways are longer than the value, they are retained.
Default is still 0. So the 'tentacle of the round trips' are removed.
Available at the testserver. (source will follow)
Please try with this [example](https://brouter.de/brouter-test/#map=17/49.60962/8.85634/standard&lonlats=8.858757,49.609236;8.857244,49.608596;8.855957,49.611008)
```
assign   correctMisplacedViaPoints  true
assign   correctMisplacedViaPointsDistance 50 
```

@0709wiwiwi 
Please provide samples in this [format](https://brouter.de/brouter-test/#map=14/51.0512/3.9963/standard&lonlats=3.982497,51.044751;3.999205,51.051093;3.98686,51.059118). So I had to pick out the points and create a request.
Result: my point 72 is the end point and has no u-turn. But may me I'm different from you test.

devemux86:
@afischerdev If we use `continueStraight=1` in the API call, the round trips do not work with the exception:
```java
ParseException .../brouter/profiles2/profile.brf at line 1: For input string: "1@correctMisplacedViaPoints"
```

If we don't use the `continueStraight=1` in the API call, the round trips work, but we have again the u-turns.

It was reported here:

- https://github.com/devemux86/cruiser/discussions/765#discussioncomment-12734309

brouter-core/src/main/java/btools/router/RoutingEngine.java
```diff
@@ -995,7 +995,7 @@ private OsmTrack tryFindTrack(OsmTrack[] refTracks, OsmTrack[] lastTracks) {
       } else {
         seg = searchTrack(matchedWaypoints.get(i), matchedWaypoints.get(i + 1), i == matchedWaypoints.size() - 2 ? nearbyTrack : null, refTracks[i]);
         wptIndex = i;
-        if (engineMode == BROUTER_ENGINEMODE_ROUNDTRIP) {
+        if (routingContext.continueStraight) {
```
devemux86, line 998: @afischerdev Here to solve the problem in https://github.com/abrensch/brouter/pull/774#issuecomment-2780728835, we could use:
```java
if (engineMode == BROUTER_ENGINEMODE_ROUNDTRIP || routingContext.continueStraight) {
```

afischerdev: That is not a bug it's a feature. The idea was to make the `continueStraight` switchable and this should also effect the round trips. Otherwise you alway have on round trip this feature on.

I made around 200 random round trips with continueStraight on. 
And again 100 random with off. 
And 100 with undefined value. No problems found.

afischerdev: And please notice this:
> ParseException ... profile.brf ...

This is thrown before routing.

devemux86: @afischerdev Have you tested on Android?

The following bundle does not work with `continueStraight=1`.
If we remove the `continueStraight=1` then it works.

```java
Bundle[{exportWaypoints=1, timode=2, allowSamewayback=0, roundTripPoints=5, profile=trekking, alternativeidx=0, roundTripDistance=6333, lonlats=1.384443,43.444626, trackFormat=json, maxRunningTime=600, extraParams=continueStraight=1@correctMisplacedViaPoints=1&correctMisplacedViaPointsDistance=0&allow_steps=1&allow_ferries=1&ignore_cycleroutes=0&stick_to_cycleroutes=0&avoid_unsafe=0&consider_noise=0&consider_river=0&consider_forest=1&consider_town=0&consider_traffic=0&consider_elevation=0, engineMode=4}]
```

afischerdev: Please look here:
> ...extraParams=continueStraight=1@correctMisplacedViaPoints=1...

separator should be a `&`

devemux86: @afischerdev What a silly mistake, sorry for the confusion and thanks! :slightly_smiling_face:

Issue #519, opened 2023-03-26
More on missplaced via points


afischerdev:
Here comes an idea for an update on the misplaced via point logic.

The current lib prevents simple spikes by going back and testing for equal connections. 
Please see point two in this [sample](https://brouter.de/brouter-web/#map=15/50.0004/8.7052/standard&lonlats=8.69946,50.001146;8.701859,50.000416;8.704949,49.999023;8.711557,49.999933;8.715055,50.000705).

New is the recalculation on a via point section with a polygon circle and finding the in and out points. And use these points for a new `findtrack` session to get areas on roundabouts or other non-trivial situations.

![dreieich](https://user-images.githubusercontent.com/23175706/227789146-2e725a2e-cd45-444e-ae6a-6915addff61a.png)

This also covers older issues like [#441 comment](https://github.com/abrensch/brouter/pull/441#issuecomment-1179509565)

![Braunschweig](https://user-images.githubusercontent.com/23175706/227789722-5c5bbc32-e834-424e-8970-d023006bbc72.png)

What do you think?

It take a little bit more on time. And it's not fully tested. 
But if interested have a look into the [repository](https://github.com/afischerdev/brouter/tree/missmatch-update).

polyscias:
> What do you think?

Great idea. 

I find it hard to review the code but if it does do good things to slightly misplaced via points it would be good to have!

afischerdev:
@0709wiwiwi 
You are right, may be this is not a clear sample.
1. the trekking profile goes the wrong way to via point 1. please see #485
2. in this case the missplacing is recalcualted by going back the line.

When you switch to car routing (or change your bike profile) it will go the correct way around and you become a nice circle tour with missplacing.

![Screenshot_20230330-104720](https://user-images.githubusercontent.com/23175706/228783728-5dc31cd2-59dc-467d-a65b-bc45bb5f5658.png)

afischerdev:
@polyscias 
Thank you for your code review. 
It's good to know that someone is keeping an eye on it.
And I believe this it is not the morning paper for reading.
I think I have to add some more comments in the code. 

[There](https://github.com/afischerdev/brouter/tree/missmatch-update) is a merge with the last updates from main BRouter.

PR #785, opened 2025-05-11
Mismatched point extention for roundabouts

afischerdev:
Here a variant of the issue #519 

This should work with four situations on aroundabouts.

via_rds_1 via_rds_2
via_rds_3 via_rds_4

And the more complicated situation:
via_rd_1 via_rd_2
via_rd_3 via_rd_4

Please note that the correction of the via point is moved to the start of the roundabout. This should make it easier to move the second line later on.


afischerdev:
This is now testable on the test server.
The BRouter-web interface doesn't know about it and will show the standard concatenation. But you should get results when you use the server export funktion - cloud symbol on the export dialog.

PR #795, opened 2025-05-31
Update Android app with lookups.dat version check

afischerdev:
To protect the first start after an update in the lookups.dat version, the files
```
serviceconfig.dat
timeoutdata.txt
```
inspected for none inbuilt profiles.


afischerdev:
With the last change in #458 (T1-hiking enabled again) this changes are unnecessary.
What do you think?

devemux86:
@afischerdev Agreed if compatibility remains.

PR #797, opened 2025-06-02
Add node:traffic_calming=choked_table to lookups.dat

ProBackup-nl:
Nodes with both vertical and lateral deviation. Therefore different cost assignment, thus as a new line.
Tag `choked_table` is in the [wiki](https://wiki.openstreetmap.org/wiki/Tag%3Atraffic_calming%3Dchoked_table).


afischerdev:
Please stop adding lookups direct, it should be discussed in #458.
As you see this has a long history. And if ok then I add this in my repository. So we have only one new version to check. 

And value `choker_table` is already used as alias

afischerdev:
added to #458

PR #798, opened 2025-06-02
Add node:traffic_calming=double_dip to lookups.dat

ProBackup-nl:
Seen mainly in Europe, especially in The Netherlands. Especially as traffic calming in cycleways. The double dip is there often seen around Utrecht, Groningen, Amsterdam and The Hague. It's intended to reduce the speed of speed pedelecs, mofas, mopeds. Horrific for velomobiles (without suspension).

Taginfo: https://taginfo.openstreetmap.org/tags/traffic_calming=double_dip


afischerdev:
added to #458

PR #799, opened 2025-06-02
Add node:surface=asphalt to lookups.dat

ProBackup-nl:
Intentended use is for traffic calming. Asphalt tables often are of the sinosoid shape. For the combination of table and sinus shape, the vertical deviation has a lot less impact.


afischerdev:
added to #458

PR #803, opened 2025-06-07
Waypoint distance error

afischerdev:
There was a report on an unexpected result in waypoint positioning.
When waypoint is exact on a node1 or a node2 it stopped searching for more and use the current way for routing.
via_1
Original from [Cruiser Forum](https://github.com/devemux86/cruiser/discussions/799#discussioncomment-13387145)


devemux86:
@afischerdev Users have tested it and it seems to work correctly now.

afischerdev:
@devemux86 
Thanks for the tests and the info.

Issue #807, opened 2025-06-18
changing Profile files with brouter app


becksy82:
Hi,

I tried to change brouter Profile files with the brouter App. But the App doenst Change the files.

See this from osmand. Ist a summary of the Problem.

https://github.com/osmandapp/OsmAnd/issues/22782#issuecomment-2980702403

I am on Android 15 with S22

BR

afischerdev:
When you like to use custom profiles in BRouter app, you need to use a new name for it. E.g. make a copy of `trekking.brf` named `mytrekking.brf`. Put your changes there and 'book it' via the server mode to bicycle_short and/or bycycle_fast.

When you update your BRouter data, the internal BRouter profiles are also updated.

becksy82:
I tried it. Doesnt work.

afischerdev:
@becksy82 
Can't say why. This is what my OsmAnd installation sends to BRouter app:
```
v = bicycle
fast = 1
lats = [D@52f3f5c
lons = [D@7943265
trackFormat = gpx
nogoLats = [D@926b3a
nogoLons = [D@3df83eb
nogoRadi = [D@31ef548
turnInstructionFormat = osmand
acceptCompressedResult = true
```
First entries is v and fast. With this BRouter calls the `bicycle_fast` entry from above.
Please control your BRouter server modes. Should look like this:

![Image](https://github.com/user-attachments/assets/c5c027b9-ad92-4b31-a18c-1a924c8610d0)

becksy82:
![Image](https://github.com/user-attachments/assets/13eda06a-f821-45b0-be99-a2e898d461da)



This is mine. But of i Change the Profile settings with the brouter App. 

![Image](https://github.com/user-attachments/assets/7bf1bb86-21b0-4e66-bd62-4e6f3a32e34d)

The Profile file will Not changed.

afischerdev:
The second screen shows parameter changes. You will find this in the serviceconfig.dat (folder ./Android/media/btools.routingapp/brouter/modes)
This changes not the profile text but the call of the profile with different parameters.
You can only control this with routes using different parameters. 
E.g. select `allow_ferries` - is easier to test - and do routes with and without it.

vodie:
You can't change the profile with the BRouter app. You can only set some parameters for the routing call by mode (no [...] in the OsmAnd profile name), but this doesn't work if you call the profile by name in OsmAnd with [...]. If you want to call a customized profile by name you have to edit the profile with an text editor.

[...]

PR #808, opened 2025-06-19
Apk: Use profile parameter

afischerdev:
In the discussion for #807 about profile and parameter we found that parameter changes made direct in BRouter are not used in routing.

I will examine `repeat:profile` function later on.


afischerdev:
The parameters are only used to build a hash for the timeoutdata.
But when routing again via the <repeat:profile> function the parameter values are gone.

PR #812, opened 2025-07-08
Apk: Update for Android 16

afischerdev:
Google wants us to update the Android target version.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.11.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.3-bin.zip
 networkTimeout=10000
 validateDistributionUrl=true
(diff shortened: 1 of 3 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.11.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.3-bin.zip
```
devemux86, line 3: @afischerdev It's best to use the latest official recommendation, currently Gradle `8.13`:
https://developer.android.com/build/releases/gradle-plugin

Also it's better to upgrade Gradle using command line so all Gradle files are updated correctly:
(shell scripts `gradlew` and `gradlew.bat` and the `gradle-wrapper.jar`)
```groovy
./gradlew wrapper --gradle-version 8.13
```

afischerdev:
@devemux86 
Thanks for your remarks.
Yes, I'm using Android Studio. But not in my git folder, there is only a CLI used.
Android 16 is the last version so I would like to stay with that. 
But I switched back to the gradle build 8.10.1, but the dist 8.13  plugin.

PR #823, opened 2025-08-05
Update Android Gradle plugin

devemux86:


brouter-util/src/main/java/btools/util/StackSampler.java
```diff
@@ -47,6 +47,7 @@ public void run() {
     }
   }
 
+  @SuppressWarnings({"deprecation", "RedundantSuppression"}) // Android
```
devemux86, line 50: Suppress warning as `Thread.getId()` was deprecated in API level 36.
And we cannot check here with Android code `Build.VERSION.SDK_INT`.

https://developer.android.com/reference/java/lang/Thread#getId()

Issue #702, opened 2024-05-18
RFC: Improvement to Logging


mjaschen:
Following #699 I'd like to discuss some more improvements to the log output of BRouter server.

There are three things I'd like to address:

1. separation between *stdout* and *stderr* outputs
1. structure of the log messages
1. introduction of (optional) structured logging

These are just some ideas scribbled down. Before going any deeper, I'd like to hear your thoughts on this.

## Separation between stdout and stderr outputs

Currently BRouter emits messages to *stdout* and *stderr.*

Sometimes messages related to each other are printed to different streams (e.g. error message printed to *stdout,* followed by a stacktrace printed to *stderr*).

A clear separation between regular log messages (i.e. routing requests or profile uploads) and informational/error messages (e.g. startup messages, usage information, errors, stacktraces, debug output if enabled) would be beneficial.

## Structure of the log messages

I've tried to describe the structure of the log messages in the [README](https://github.com/abrensch/brouter?tab=readme-ov-file#run-the-brouter-http-server).

I think there's some room for improvement here, mainly removing redundant information which has to be filtered out in later processing anyway.

I'd like to propose a structure that's very close to the current format, with some small changes:

```text
<timestamp> <session count> <client ip> <processing time> <request query string> [...]
```

| Field                  | Description                                                   | Changes                                                                                                                                               |
| ---------------------- | ------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
| `timestamp`            | The current timestamp in ISO 8601 format, unchanged           |                                                                                                                                                       |
| `session count`        | `new` for newly detected client or counter (positive integer) | no more capping at 999,<br>remove padding                                                                                                                                  |
| `client ip`            | IP address of client                                          | remove `ip=` prefix                                                                                                                                   |
| `processing time`      | processing duration in _ms_                                   | remove `ms=` prefix                                                                                                                                   |
| `request query string` | the plain request query string                                | remove HTTP method (`GET`),<br>remove HTTP version (`HTTP/1.1`),<br> remove path components (`/brouter?`), i.e. print just the request parameter pairs separated by `&`,<br>enclose in query string quotes `"` |

An example of a log message would look like this:

```text
2024-05-14T21:11:26.499+02:00 new 127.0.0.1 189 lonlats=13.377485,52.516247%7C13.351221,52.515004&profile=trekking&alternativeidx=0&format=geojson
[...]

PR #830, opened 2025-08-29
[WIP] Replace logging with slf4j/logback

jgillich:
This change replaces all logging in brouter-core with slf4j.

Why? The current logging is a mess. `RoutingEngine` implements a simple logger that doesn't preserve log levels or timestamps. Other core classes use `System.out` instead. To enable debug logs, there are sometimes static variables and sometimes config is read from system properties. To me this all seems confusing and difficult to configure, therefore I propose to replace it all with the simple, standardized and pluggable library that is slf4j.

slf4j is merely an abstraction and does not provide logging itself. This gives the user of the library the freedom to use any library of their choosing. The server, app and tests use logback to do the actual logging. Logback also supports structured logging as proposed in #702 (cc @mjaschen), I have modified many of the log messages to use the key-value syntax.

RoutingEngine used have an option to to write its log to a `debug.txt`. I've removed this because I believe a library should generally not access the disk directly. Per-profile logging could be implemented by passing a `ILoggerFactory` to RoutingContext, but I haven't implemented this yet because I wonder one thing: Do we even need this file? If we attach the profile name as structured data, one can simply grep the general log file or use a filter in an advanced system like Logstash etc. Having to locate the individual file each time honestly sounds more complicated to me. Let me know what you all think.

### DIFF
--- a/brouter-core/src/main/java/btools/router/AreaReader.java
+++ b/brouter-core/src/main/java/btools/router/AreaReader.java
@@ -1,5 +1,8 @@
 package btools.router;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.DataInputStream;
@@ -27,6 +30,7 @@ import btools.mapaccess.OsmNodesMap;
 import btools.mapaccess.PhysicalFile;
 
 public class AreaReader {
+  private Logger logger = LoggerFactory.getLogger(AreaReader.class);
 
   File segmentFolder;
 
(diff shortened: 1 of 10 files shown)


devemux86:
Why can't we use the `java.util.logging` API, which works on all platforms
and avoid loading the library with third-party dependencies?

jgillich:
JUL is not particularly popular, it's [10x slower](https://stackoverflow.com/questions/11359187/why-not-use-java-util-logging) and it can't do structured logging. Is there any platform that SLF4J wouldn't work on? I'm not sure but I doubt it

zod:
I'm not sure if the performance issues of JUL would affect us but I think it's a good idea to replace the print statements with real logging.

PR #842, opened 2025-09-29
More realistic bicycle weights

Totorrr:
Slightly adjust different weights in bicycle profiles, reflecting what I believe is closer to reality, and taking bike hints into consideration in some cases to adjust weights more precisely.


Totorrr:
Hello @quaelnix, what would be your feedback on this one please? 

It seems to me, having read some of your past comments, that you value imperfect-map resistant routing that approaches real-world experience, so, I'd appreciate hearing from you on this one. 

You may also check my 3 other PRs if you'd like. Thanks.

PR #852, opened 2025-11-17
trekking: add whether there is a usable cycleway in the isbike bike hint (v2)

Totorrr:
Follow-up of PR #838

Reminder: 

> Some cycleways are not always properly detected, because some tagging schemes are not as straightforward as having bicycle=yes defined on the way. The isbike bike hint we use would be closer to reality, and thus offering better routing, if we'd consider some more complicated tagging schemes, like "is there a usable cycleway on this road?".

### DIFF
--- a/misc/profiles2/trekking.brf
+++ b/misc/profiles2/trekking.brf
@@ -85,7 +85,7 @@ assign badoneway =
 
 # set isbike considering access, local cycle route or the presence of a usable cycleway on the highway
 assign isbike =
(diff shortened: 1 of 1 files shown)


afischerdev:
Could someone please test the profile?

Totorrr:
> Could someone please test the profile?

Hello @devemux86, would you be willing to try a few routings with this profile to test if it works nicely? That would be great!

afischerdev:
I'm sorry for the delay. Thank you for your cooperation.

Issue #854, opened 2025-11-25
Enable pseudo tags for car profiles


afischerdev:
It is hard to add some of the new pseudo tags to a car profile. The only entry point for variable values is the `maxspeed` parameter.
However, the logic of, for example, `consider river` is: all paths that are not near a river are penalized. And when you add a penalty to `maxspeed` this will affect your route duration. 

The code is not ready yet, but the feature for the moment is available for testing.

Please see this [sample](https://brouter.de/brouter-test/#map=14/50.0978/8.6776/standard&lonlats=8.664816,50.088858;8.699614,50.104079&profile=car-fast) for a river test.

The link is set to `car-fast`. This is not the original but a replacement with a contact to a new KinematicModel class. This contains now an entry point for `costfactor` in a similar use as bike profiles.
When you place the penalty on costfactor, it should not bring a duration problem.
Please see the text inside the profile.

Another [example](https://brouter.de/brouter-test/#map=13/42.4860/1.5376/standard&lonlats=1.490831,42.476783;1.630685,42.53952&profile=car-fast) for use with `avoid_motorways`, `avoid_trunk` and `avoid_primary`.

Please play around with this examples and test if this fits your needs. 
The next summer will come with a cabrio ride.

The current car profile doesn't contain all pseudo tags. Which do we need for car driving?
**Note:** 
On the testserver the tests for issue #848 and the PR #853 are running as well.

devemux86:
@afischerdev 

> The current car profile doesn't contain all pseudo tags. Which do we need for car driving?

This profile can be used and modified for all vehicles, cars and motorcycles, so we at least need:
- consider_town
- consider_forest
- consider_river

afischerdev:
@devemux86 
Thanks for the ideas,
They are online on test server now - car_fast profile.

devemux86:
@afischerdev Thanks, very interesting results.

The pseudo tags may need stronger costs?

For example this [route](https://brouter.de/brouter-test/#map=10/50.0620/8.6565/standard&lonlats=8.682208,49.707328;8.399171,50.436511&profile=car-fast) has very little difference with consider_town and still passes through the city.

afischerdev:
@devemux86 
> The pseudo tags may need stronger costs?

Maybe, I did not care about the values, just copy it from trekking profile. But If you have a look at the same route done with the f`astbike less traffic` profile, the values are a little bit different. We need to experiment a little with these values ​​to find the best compromise.

devemux86:
[...]

PR #860, opened 2025-12-12
Kinematic class update

afischerdev:
According to the #854 here a test class and test profile.
This can still be tested on the test server.

This is **only** a test class, at the end should be the replacement of the origin KinematicPath class.
But I guess this needs further testing.

### DIFF
--- a/brouter-core/src/main/java/btools/router/KinematicWeightModel.java
+++ b/brouter-core/src/main/java/btools/router/KinematicWeightModel.java
@@ -11,13 +11,13 @@ import btools.expressions.BExpressionContextNode;
 import btools.expressions.BExpressionContextWay;
 
 
-final class KineNewModel extends OsmPathModel {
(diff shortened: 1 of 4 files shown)


devemux86:
@afischerdev 

Can we set avoidance levels with the costs?
Do the costs have a minimum and maximum?

For example, avoid motorways by 50% or we can only set them on and off?

afischerdev:
@devemux86 
As in `avoid motorway` you can set a value, but can not say if this will fit 50 percent.

```
assign costfactor 
  ...
  else if is_avoided_motorway then 10
  else if is_avoided_trunk then 10
  else if is_avoided_primary then 10
  ...
```

afischerdev:
Currently, I want to keep both – the old and the new kinematics classes. This gives more time for testing.

devemux86:
> Currently, I want to keep both – the old and the new kinematics classes. This gives more time for testing.

@afischerdev This is the best. :+1:

PR #863, opened 2025-12-16
profiles: correc. isunpaved ignoring smoothness if surface undefined

Totorrr:
Tiny bug correction


afischerdev:

PR #864, opened 2025-12-16
profile: refactor, refine and simplify the logic handling oneways

Totorrr:
Currently this logic is spread in `badoneway`, `isbike` and `onewaypenalty` variables, that can lead to repetition or confusion.

With this change, new intermediate variables like `hascycleway` and `hasbikerouteoraccess` may allow more precise routing rules. If Brouter engine has lazy loaging, it may also improve performance, for example, in `onewaypenalty` don't re-check every possible cycleway:xxx one by one if `hascycleway` is false anyway.

Ths `isbike` behaviour is changed in the sense that it still checks the presence of a cycleway, but it does not check any more if it's a oneway that we can legally use in the current direction. If leaves this to be handled within `onewaypenalty`.

### DIFF
--- a/misc/profiles2/fastbike.brf
+++ b/misc/profiles2/fastbike.brf
@@ -133,15 +133,23 @@ assign badoneway =
          else oneway=yes|true|1
        else oneway=-1
 
+assign hascycleway = not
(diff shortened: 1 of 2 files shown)


afischerdev:

PR #865, opened 2025-12-16
Profile lower pedestrian cost if cycleway

Totorrr:
This may allow passing safely on more pedestrian ways.

Depends on PR #864 (that is, commit 0dfc858d)


afischerdev:

Issue #658, opened 2024-01-09
Hiking profile ignores `acces=no`


Helium314:
Brouter routes along steps with `access=no`, which it definitely should not:
https://brouter.de/brouter-web/#map=18/48.26936/16.33782/standard&lonlats=16.339615,48.269619;16.336176,48.26888&profile=hiking-mountain

Or is there any reason to ignore this (imo very straightforward) restriction?

poutnikl:
The issue goes rather to me and my Hiking-Poutnik repository, as as the Brouter Hiking-Mountain is based on it.

I have checked if it is some error, as it is some time since I made last editing. It is not an error, but on purpose, by the profiles design. As you can check at [Hiking-Poutnik - master - Hiking.brf](https://raw.githubusercontent.com/poutnikl/Hiking-Poutnik/master/Hiking.brf), there is assigned   
 `assign accesspenalty switch footaccess 0 9998`

As you can see at [Brouter-profiles - wiki - Glossary](https://github.com/poutnikl/Brouter-profiles/wiki/Glossary#c), the Costfactor 9998 is the highest costfactor still allowing passage. It means that e.g. 10 m of such a way is equivalent about 100 km of the way considered as optimal.

The rationale behind this is that if there is any other possible way, Brouter avoids an OSM way with such a high costfactor.  But, if there is no other way, routing fails with error there is no possible way.   There are many scenarios you have good justification to go via such a way anyway. There are often many cases OSM data are not properly tagged and restricted access does not in reality applies to pedestrians, or it can be justified, 

So the decision has been made the risk to allow it is lesser harm than the risk of not allowing it.

It context of BRouter profiles, there are many such occurences of decisions that involve weighing of risks and impacts of allowing what should be denied versus denying what should be allowed.

poutnikl:
As work around, it can be change on BRouter web as custom profile edit to 

`assign accesspenalty switch footaccess 0 9999`

the values costfactor=9999 causes the particular segment is not considered for routing, but BRouter is still aware of it during generation of direction hint ( 10000 cause total ignoration).

As a variant for aventual profile tweak, there can be defined the variable in the global profile segment like

`assign UseHardNOaccess = false`   # used the current default

and in the way profile segment

`assign accesspenalty switch footaccess 0 9999`

could be replaced by 

`assign accesspenalty switch footaccess 0 switch UseHardNOaccess 9999 9998`

vodie:
[...]

PR #866, opened 2025-12-17
profile: Remove hikingroute always having access granted

Totorrr:
Same as in commit e2c68117 and 35f4e331 for bicycle. See their comment.

Should solve issue https://github.com/abrensch/brouter/issues/658. Can you have a look at it @poutnikl?

Extract:

Some hiking & bicycle routes can have portions which are not accessible.

Potential reasons for this:
 - temporary construction work on the ways (sometimes for months),
 - local access change without simultaneous update of bicycle route relation (lack of mapper's knowledge or lack of time to redefine an updated route),
 - ...


afischerdev:

Issue #536, opened 2023-04-20
Feature Request: consider maxheight


francwalter:
Hello

I would love to see a feature where I can consider the maxheight of streets underneath bridges or such.
Often I use BRouter in Locus Maps to get some routing for our car, which is 2.75 m of height (camping car). Sometimes we cannot pass, the maxheight is not high enough and we need to make big drivearounds. 
Is there maybe a possibility to add this as profile?

It could be also useful to have the max weight in a profile, I guess.

Thanks.
Frank

poutnikl:
Looking at brouter.de/brouter/profiles2/lookups.dat, where is exposed the subset of OSM data tags, such info is not currently implemented in BRouter data file, so profile authors have to wait, until that is eventually done. Another question is how reliable are such data included in OSM data for tunnels and bridges,

Totorrr:
Pleases see PR #889 for this. It adds the relevant maxheight keys to `lookups.dat`.

afischerdev:
@Totorrr 
Thanks for your excellent work.
But please hang on a while with the PR #889
I like to point you to the discussion in #878.

There we discuss to add truck data to the lookups.dat and how to do that.

Short summary:

Trucks need more data than maxheight and there are more interests like camper or boat people in this values. 
As you can see in your PR, these changes require a lot of entries for the values. And it will grow with all the others from #878. 

So main question is: do we like to add all this as your main focus is on biking and hiking?

Next question is how to do that:
BRouter can work with index based and value based entries (some test results for this please [see here](https://github.com/abrensch/brouter/issues/878#issuecomment-3907908467)). 

**index based**
- add every relevant values to the lookups (with sub values for other system of measurement)
- add for every index an entry in the profile to calculate with the value (please [see here](https://github.com/abrensch/brouter/blob/46b6657e1f8ce8ec5858c1c6257fcea25b84abde/misc/profiles2/car-vario.brf#L137) and more lines)


**value based** 
- add only one line to lookups for each tag like
  maxheight;000000001 *
- add one line in the profile to get the value, like
```
assign no_height
    switch and not  maxheight=
	              lesser v:maxheight    boat_height true
    false  
```

Totorrr:
Hello @afischerdev,

thank you for pointing the discussion in issue #878, I did not knew it before. 


> **index based**
> 
>  * add every relevant values to the lookups (with sub values for other system of measurement)

[...]

PR #896, opened 2026-03-21
Added expressions for weights and new output options

afischerdev:
This added
- calculation for weight measurements
- direct output for values when they are index based
- error output option (please add -DshowErrors to java call to display it)

The new tests show how this will work
It contains test for 
- value based input and conversion
- index based output direct as value
- mix of index and value base entry on maxheight example
  **Please note**: on mix logic will no output for `unknown` values

This test nw should be independent from  production profiles.


Totorrr:
Hello @afischerdev, what about my reviews?

afischerdev:
@Totorrr 
#897 is merged now.
#536 maxheight and similar tags (like #878) will be not a part of the current update.

Totorrr:
@afischerdev, no, I'm talking about my reviews of **this** PR. See above my comments on the diff.

afischerdev:
@Totorrr 
> this PR. See above my comments on the diff.

Hmm, I don't see any.

Totorrr:
@afischerdev my bad apparently, only I could see the ~commits~ comments as they were not submitted it seems... I just did submit them.

afischerdev:
@Totorrr 
Thanks for the hint - very attentive.
I guess this was changed in [update](https://github.com/abrensch/brouter/commit/4dd1df3d7a94b83c54b511a1f8896f44c993f008) and before it was 10 degrees.
I'll start a change request.

PR #885, opened 2026-02-28
Preparing next Version 1.7.9

afischerdev:
Added a new version for BRouter app, library and profiles, please see [changes](https://github.com/abrensch/brouter/pulls?q=is%3Apr+milestone%3A%22Version+1.7.9%22+is%3Aclosed).

PR #903, opened 2026-03-28
Improve round-trip routing quality

jonnybbb:
## Summary

Adds **round-trip (loop) route generation**: from a single start and a desired
loop size, the engine plans a closed loop back to the start. Introduces a greedy
routed-leg planner, an isochrone-derived candidate pool, a loop quality model
(geometry + road-character + corridor-overlap), and an AUTO strategy competition,
with extensive unit + integration coverage.

Inspired by the Dijkstra-based greedy cycle-route approach ("Efficient
Dijkstra-Based Greedy Algorithm for Cycle-Route Planning", CEUR-WS Vol-3885).

### Engine (`brouter-core/.../router`)
- **GreedyRoundTripPlanner** — iterative routed-leg construction: each step runs a
  real Dijkstra to a candidate waypoint, so legs follow roads, not beelines. Convex
  via placement (angular sweep + unimodal radius) and a heading-persistence term keep
  the loop shape clean.
- **Candidate providers** — IsochroneCandidateProvider (road-native candidates from
  a bounded isochrone expansion), GraphNative/Blended, plus experimental
  Desirability/Capsule providers (off by default), behind RoundTripCandidateProvider.
- **Quality model** — RoundTripQualityGate (STRUCTURAL-bad vs QUALITY-degraded),
  ReuseClassifier, LoopQualityMetrics, RouteChoiceScore, CandidateScorer,
  RoadCharacterScore (road-character appeal), CorridorOverlapIndex (parallel
  same-corridor detection).
- **RoutingEngine.doRoundTrip** — direction/sub-route sweeps + the **AUTO competition**
  (ISO_GREEDY → GREEDY → WAYPOINT → ISOCHRONE last-resort), scoring each, adopting best.
- **Lenient-by-default gate** — a rideable QUALITY-degraded loop is returned with a
  `Warning:` advisory; STRUCTURAL failures are always rejected. `roundTripStrictQuality=1`
  restores hard-reject of QUALITY failures.
- **Data-availability fast-fail** — checks `NodesCache.hasSegmentFor(start)` and fails
  fast with a clear "datafile … not found" instead of a confusing empty result.

### Shared routing code (general routing preserved)
This PR also touches a few core files, but **general (non-round-trip) routing is
behavior-preserving** — verified by method-level diff and pinned by tests:
- **OsmPath** anti-reuse refTrack penalty: the more precise edge-membership test
  (`OsmTrack.containsTraveledSegment`) is **gated behind round-trip mode** via
  `RoutingContext.roundTrip`; general `alternativeidx` routing keeps the historic
  both-endpoints `containsNode` test (pinned by
  `RoutingEngineTest.generalAlternativeRefTrackPenaltyIsHistoric` and, on real tiles,
  `AlternativeRefTrackMembershipTest`).
- **WaypointMatcher.start(...)** gains a trailing way-description arg, threaded through
  MicroCache2/DirectWeaver; consumed only by round-trip via-snapping and null-safe on
  the normal path.
- Additive/no-op for normal routing: `CheapRuler.getScaledBearing` (new),
  `NodesCache.hasSegmentFor` (pure read), `OsmTrack.containsTraveledSegment` + a 0-node
  read guard, `FormatGpx` null-guards for the round-trip merged-track output path.

### API / parameters (`engineMode=4`)
- `roundTripLength=<m>` — desired **total loop length** (precedence over radius; clamp 500 km).
- `roundTripDistance=<m>` — legacy **search radius** (≈ length/2π; clamp 200 km).
- `roundTripAlgorithm=` user tiers **fast** (→waypoint), **balanced** (→greedy, default),
  **quality** (→iso_greedy), plus `auto` (competition) and raw `waypoint|greedy|iso_greedy|isochrone`.
- `roundTripPoints`, `roundTripDirectionAdd`, `roundTripIsochrone=1`, `roundTripStrictQuality=1`,
  `roundTripSteerVias=1` (opt-in dense-area via-steering). Abusive length/radius values are
  clamped, not rejected.

### Tests
- **Unit** (`src/test`, ~41 classes): planner, providers, gate, classifier, metrics,
  param parsing, output, AUTO competition, convex-via geometry — fast, bundled Dreieich
  fixture, no network.
- **Integration** (`src/integrationTest`): real-segment loop-quality matrix with a Leaflet
  HTML report, golden-signature/gold-standard regression, weak-cell/nogo regressions,
  user-via, and a refTrack-membership proof — via a dedicated `integrationTest` task
  (3 GB heap/fork, on-demand tiles), excluded from the standard `build`.

### Docs
- `docs/features/roundtrips.md`.

## What's been tested

```bash
# Fast unit tests (no network; brouter-core:test builds the Dreieich fixture via map-creator)
./gradlew :brouter-core:test :brouter-mapaccess:test :brouter-util:test :brouter-server:test

# Full build + PMD + Checkstyle (does NOT run integrationTest)
./gradlew build

# Slow real-segment suites (downloads tiles on demand; the task IS the gate, no opt-in flag)
./gradlew :brouter-core:integrationTest
#   useful overrides:
#     -Dloop.segments.nodownload=true          # use only locally-present tiles
#     -Dloop.profiles=fastbike,gravel          # filter profiles
#     -Dloop.algorithms=greedy,iso_greedy      # filter algorithms
#     -Dgolden.tests=true -Dgolden.write=true  # recapture golden baselines
#     -Dgreedy.diagnostic=true                 # greedy planner diagnostics
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


afischerdev:
Thanks for your improvements.
After the new version #885 I'll start testing.

jonnybbb:
I am going to provide some actual use cases and test data

jonnybbb:
### Case 2: Mallorca East (dir=90°) — v1.7.8 fails, new succeeds

**v1.7.8**: :x: Routing error — eastward circle points from Vistalegre hit the coast / sparse rural roads, causing a routing failure. No GPX produced.

**New (this PR)**: :white_check_mark: 51.2 km loop, 276m ascend, 0.9% segment overlap. The route successfully navigates the eastern Mallorca countryside despite the proximity to the coast.

Another case where waypoint filtering prevents routing failures in geographically constrained areas (coastlines, islands).

**GPX files:**
- v1.7.8 baseline: _routing failed, no GPX_
- [New (this PR) — mallorca-east_new.gpx](https://gist.github.com/jonnybbb/7efad07d9d9c64b251a000433a37a5c3)

jonnybbb:
### Case 3: Basel West (dir=270°) — back-and-forth improvement

Both versions succeed here, making this a good apples-to-apples comparison of route quality.

| | v1.7.8 | New |
|---|---|---|
| Distance | 42.2 km | 47.1 km |
| Ascend | 268m | 292m |
| Segment overlap | **2.0%** | **1.9%** |

The new version produces a slightly longer, more complete loop while reducing segment overlap. The back-and-forth removal and micro-detour removal trim the retracing artifacts that v1.7.8 leaves in.

**GPX files:**
- [v1.7.8 baseline — basel-west_baseline.gpx](https://gist.github.com/jonnybbb/916b576e292a2f891b6c5ce3532905a4)
- [New (this PR) — basel-west_new.gpx](https://gist.github.com/jonnybbb/6103b48d69f6695747748f77c23a7ee7)

jonnybbb:
@afischerdev I am still testing out things - I will put the PR in ready for review once I am done with testing

afischerdev:
@jonnybbb 
> ready for review once I am done with testing

Fine, thanks for the update.

afischerdev:
Let's start a new dimension of round trips.
With some small remarks for now
- please use `gradlew build` before check in
- great that the available tests are not running in standard process - when running it took more than an hour for me
- the tests contain a reference to profile `mtb-zossebart`, this is not part of the repository, please remove the tests or use standard `mtb` profile
- during a first local test I found some voicehint entries that are not remove during process test2
- One info during process is 
  `greedy round trip: desired distance=50265m, searchRadius=8000m, direction=377`
  What does a direction of 377 mean? I see directions around 0 to 359.

I'll go on with the pre process to find the used way points.

You announced `Generated with Claude Code`. What does that mean to the project? 
- Is it royalty-free and free to use?
- do we have to publish the used prompts?

jonnybbb:
Thanks for the feedback already @afischerdev - I will look into it. I am keeping this still as draft as this is I am not considering this still too early for a proper review. But your feedback is very valuable - I will address it

**Regarding the usage of Claude Code:**

The "Generated with Claude Code" line is a transparency marker — it notes that AI assistance was used on a commit. It is not a license grant from Anthropic and not a claim on the code.

**Royalty-free and free to use?** Yes. Anthropic's Terms of Service assign all rights in outputs to the user: *"Customer … owns its Outputs."* The project's existing license applies as before — no royalties, no Anthropic-side restriction, nothing extra to accept downstream.

**Do we have to publish the prompts?** No. Neither Anthropic nor the trailer require prompt disclosure. The contributor is responsible for the code — meaning it was reviewed, can be explained, and meets the project's standards, same as any hand-written contribution.

The attribution is configurable, so if the project prefers a different convention we can adjust.

afischerdev:
Thanks for your changes
and for the explanations regarding Claude Code

- play back for start direction is fine
- unused voicehints are removed correctly
- tests for effects outside the round trip logic is ok - I missed that last time

I did some tests with shorter distance
```
lonlats=7.5878,47.5581&profile=fastbike&timode=4&roundTripDistance=1000&direction=0&roundTripDirectionAdd=0&roundTripAlgorithm=AUTO&exportWaypoints=1
```
this has no trip inside, only one point. `roundTripLength=5000` brings results

The found points are corrected, but are not included in the list of corrected points `exportCorrectedWaypoints=1`. This is not important for round trips, but could be helpful in the analysis here.

It seems when the round trip algorithm ISOCHRONE is used, the points are placed around the start point and the direction is not used.

Another region - coastal
The call with a direction = 0
```
lonlats=7.270,43.700&profile=hiking-mountain&timode=4&roundTripDistance=5000&direction=0&roundTripDirectionAdd=0&roundTripAlgorithm=AUTO&exportWaypoints=1
```
comes back with the timing (first the routing time, second the time over all)
```
execution time = 0.942 seconds
round trip execution time = 1.851 seconds
```
With a direction of 180 degree
```
execution time = 0.916 seconds
round trip execution time = 21.678 seconds
```

jonnybbb:
Thanks @afischerdev

> this has no trip inside, only one point. roundTripLength=5000 brings results.

We should probably throw an exception here if no length has been specified- WDYT?

Regarding the ISOCHRONE  algorithm - it doesn't yet really bring much benefit. I am investigating.

I will also include your other recommendations.

afischerdev:
> We should probably throw an exception here if no length has been specified- WDYT?

Or perhaps don't be so strict about removing nodes. If, after a reduction, only a single point remains, then the last segment still applies. Or something along those lines. If, for example, you are in the desert and have only one road, then you simply walk back and forth for a while.

I think it is better to come back with something then with nothing.

afischerdev:
Thanks for your updates.
I'm sorry, this build was not finished. Please see the [action results](https://github.com/abrensch/brouter/actions/runs/26631844289/job/78513117632?pr=903)

afischerdev:
Thanks for the update, compiling works now.
But ...
The example above
```
lonlats=7.270,43.700&profile=hiking-mountain&timode=4&roundTripDistance=5000&direction=180&roundTripDirectionAdd=0&roundTripAlgorithm=AUTO&exportWaypoints=1
```
works now in a faster way. But as you see in the timode I would expect voicehint entries. The list is empty. And this is not allowed and will not be intercepted by the format routines.

afischerdev:
>  in VoiceHintList.java Line 31
>  public void assertNonEmptyIfRequested(int trackNodeCount) {

Please do not throw exceptions in this way. No one ever needs that.

The challenge is to have always voicehints.

This [sample](https://brouter.de/brouter-web/#map=17/49.99634/8.70560/standard&lonlats=8.703207,49.996749;8.702445,49.99641) should work without voicehints and exceptions.

afischerdev:
This runs better for round trips. But my trips have lots of voicehints.
The simple sample above doesn't run. May be it is enough to remove the `assertNonEmptyIfRequested `and the calls for it.

afischerdev:
Now the 'normal' requests run as usually.
But the problems with the roundtrips are back.
```
AUTO: failed to write adopted track: java.lang.NullPointerException: Cannot invoke "java.util.List.get(int)" because "t.messageList" is null
```
Please do some real tests before checking in your source. 
Or design a test that is not ignored e.g. in Dreieich with a mini roundtrip.

jonnybbb:
Thanks for taking another look - and sorry for this oversight. I am still testing this feature, which is why the PR is in draft, and I am pushing to my remote branch just to have it available not only locally. I will make sure to create proper automated test cases.
Your feedback is super useful, and I need it to get this done. At the same time, I don't waste your time on unfinished stuff. I will ping you when you can take another look - does this work for you?

afischerdev:
> I will ping you when you can take another look - does this work for you?

Yes, that would be great.
Please keep in mind:
my test routine is always: generate a new lib and use it
First against a collection of old featues

`!!! adding new feature means not to remove old ones`

Second the new features against the description

jonnybbb:
The commits are quite hard to review and confusing - @afischerdev would you prefer a new PR with less and better sliced commits or should  I rebase this PR and force push or keep as is?

afischerdev:
> The commits are quite hard to review and confusing - @afischerdev would you prefer a new PR with less and better sliced commits or should I rebase this PR and force push or keep as is?

I'm not able to review at the moment.

But when you signal a review ready situation, I will download the complete repository and start with this status offline.
So, I think you could leave it as it is.

afischerdev:
@jonnybbb 
I'm back, so we could start a new test.

jonnybbb:
The PR ended quite big ...   I cleaned up the commits. Hope this makes reviewing easier. I appreciate your feedback. 
I have attached a full integration test loop results as html report for an easy to eyeball generated loop review.
[loop-reports.zip](https://github.com/user-attachments/files/29270314/loop-reports.zip)

jonnybbb:
@afischerdev, I think the PR is now ready to review

afischerdev:
@jonnybbb 
thanks for the info.
But please before adding commits, please run locally the command `gradlew build` to see if it works.
If an AI generates your commits, please add this to the prompts.

afischerdev:
@jonnybbb 
Thanks, fine, this one works.

afischerdev:
@jonnybbb 
My tests are complete - for now.
All I found were good results, generated in an acceptable amount of time.
It would be great for me if this PR were finalized and incorporated into the planned new version of the app #940.

What I’m not so keen on is the effort required for the extended round-trip feature.
The library has grown from 338K to 512K (read-only version) for what feels like a minor feature (generating helper points for a round trip), and quite a few classes have been added for this purpose. It isn't clear to me whether all of this is actually necessary to achieve the primary goal. This implies a need to familiarize myself with the new routines, which will take time.

However, those are optimizations that can be addressed later. In my opinion, it is definitely ready for use.

afischerdev:
@jonnybbb 
Thanks for removing some of the classes.
Could you do me a more optical favor? The embedded classes FrontierAxis and IsoAsymmetryBias in RoutingEngine should be moved to the outer scope.

jonnybbb:
@afischerdev Done — FrontierAxis and IsoAsymmetryBias are now top-level classes in btools.router (commit b3ae4e39). Pure mechanical move: same package-private visibility, fields and constructors unchanged; only the class javadoc and the test references were adjusted. Heads-up on what else landed on the branch since your review: two small, measured quality fixes for the loop planner (eb3336f1, 74d9bf2f). The first ensures the routed candidate comparison always sees at least one per-step graph-native candidate (a ranking bias let estimated legs crowd out exactly-known ones); the second replaces the global return-distance guess with per-sector values derived from the isochrone expansion the plan already runs — zero extra searches. Both are always-on code paths: no new request parameters, no system properties, no experimental switches. Full gradlew build plus the 915-case integration suite were green before each push, and the golden signatures were re-captured in-branch.

I will create a follow-up PR that further improves the `ISO_GREEDY` algorithm and better includes the `GREEDY` algorithm.

afischerdev:
@jonnybbb 
Thanks for the cleaning. Great.
My tests are mainly equal to the last version. Only here I have a different (blue = old, red = new lib):
```
lonlats=9.407899,52.270317&engineMode=4&profile=hiking-mountain&format=gpx&timode=4&profile:consider_forest=1&exportWaypoints=1
lonlats=9.551749,52.261752&engineMode=4&profile=hiking-mountain&format=gpx&timode=4&profile:consider_forest=1&exportWaypoints=1
```
testrt
testrt2
But it is not important and does not need to be changed. I only want to inform.


### An other topic:
I plan the docs e.g. aidl:
```
    //  extra round trip parameter
    //  "roundTripDistance" = distance to the round trip points (default 1500)
->  //  "roundTripLength"   = length for the round trip in meters, no default, roundTripDistance is used
-x  //  "roundTripPoints"   = number of used points (default 5)
```
Only the trip length goes out as new parameter.
Others like `roundTripAlgorithm`, `roundTripStrictQuality`, `explicitViaDensify` or similar are not needed for a client to start a round trip calculaltion. `roundTripPoints` isn't needed either.

What do you think?

jonnybbb:
@afischerdev Thanks for testing again! I checked both requests against the old and new libraries.

Because these requests have no fixed `direction`, a single screenshot comparison is noisy. `consider_forest=1` only biases the random start direction; it does not make it reproducible. I ran each request 6 times per library.

Result:

- (9.5517, 52.2618): no real library change. Both libraries return the identical 8759m loop as their usual result (old: 5 of 6 runs, new: 6 of 6). The one remaining old-lib run produced a 9565m variant — that is the random start direction at work, and that is also the difference your screenshot caught.
- (9.4079, 52.2703): real branch change, but acceptable: 9195m -> 7862m, road reuse 3.8% -> 0.8%, no self-crossings, quality score almost unchanged (~0.02 difference). Verified with a pinned `direction`: this change is deterministic, not the random draw. I would leave it as-is.

Using `direction=<angle>`, the loop should be reproducible. If clients want another reproducible loop, use `alternativeidx=<n>` as the round-trip variety seed (`0` = default loop, other values = deterministic variants). So the stable client pattern is: fixed `direction` + chosen `alternativeidx`.

### AIDL docs

I agree with your minimal set, with one addition.

I would document `roundTripLength` as the main size parameter and keep `roundTripDistance` as the legacy radius/default fallback. `roundTripLength` wins if both are set.

I also agree to leave out `roundTripPoints`, `roundTripAlgorithm`, `roundTripStrictQuality`, and `roundTripDensify` from the AIDL docs. `roundTripPoints` is actually not used by the new engine anymore (it is parsed but ignored — the planners choose their own point count), so removing it from the docs matches reality. Without `roundTripAlgorithm`, the client gets `AUTO`, and that is the intended default.

I added the round-trip block:

```java
// "direction"      = initial round-trip bearing; use a fixed angle for reproducible loops, -1/random otherwise
// "alternativeidx" = deterministic round-trip variant seed (0 = default; for round trips any value >= 0 works, it is a seed, not an index)
```

Those two are more useful for clients than the internal tuning knobs, because they give reproducible default loops and reproducible variants.

afischerdev:
@jonnybbb 
New tests are fine.

Docs in your way are fine as well.

jonnybbb:
Perfect - anything else I should look into? Otherwise 👍🏻 from my side to get this released and merge to master. The mentioned planned improvements (https://github.com/jonnybbb/brouter/issues/26) will take more time and I suggest to eventually include them in an upcoming release

afischerdev:
@jonnybbb - thanks for your work.

devemux86:
I did some tests with the new round trip calculation.

The results are better, but it takes a much longer time to calculate a round trip?

afischerdev:
@devemux86 
Yes, it takes a little bit more, better results have a price. 
However, if you compare it with `consider_forest=1`, timing change.

devemux86:
> However, if you compare it with consider_forest=1, timing change.

Do you mean to disable the forest and similar options to improve the calculation time?

jonnybbb:
@devemux86 

  The new default `AUTO` mode spends more work on candidate generation and quality checks, so it can be slower than the old waypoint-only round trip. The trade-off is better loop shape, less same-way-back, and better distance
  quality.

For faster responses you can use `roundTripAlgorithm=FAST` it is faster (and still better than the previous version), but lower quality compared to `AUTO`.

If you have a request where timing is much worse than expected, please share the exact URL/parameters. Then I can check whether it is normal `AUTO` cost or a case we should optimize.

afischerdev:
@devemux86 
> Do you mean to disable the forest and similar options to improve the calculation time?

The other way round. In the past with `consider_*` the area around was scanned - with a higher cost in terms of time. But it was recycled for the next requests.

devemux86:
The [documentation](https://github.com/abrensch/brouter/blob/master/docs/features/roundtrips.md) may need correction, the correct parameter for the round trip direction should be "direction".
(the "startDirection" does not exist in the code)

devemux86:
On phones the "AUTO" takes too much time, like minutes.
Even on desktop larger round trips take minutes.
Users will not wait so much time to get a result.

Also in some tests there are many new created waypoints at the same location.

And the "direction" or "heading" options do not seem to work as expected?
(why do we have 2 options for the same thing?)

Direction seems to be only the start bearing and not the whole round trip position related to the start,
as it usually works in other routers.

For example see the image, it was calculated with direction at "west",
but we get start bearing at west, while the waypoints are in all directions and the round trip is not at west:

jonnybbb:
@devemux86 Thanks, for your useful feedback.

I agree that `AUTO` can take minutes on phones, particularly on very long 170km loops.  This may not be acceptable for the normal interactive flow. For mobile clients, you could experiment with `roundTripAlgorithm=FAST` and keep `AUTO` as the high-quality mode when the user explicitly asks for a better loop and accepts the longer calculation time.

The duplicate waypoints generated at the same location could be a real bug, or at least a missing cleanup step. Could you share the exact URL/parameters for one case? Then I can reproduce it and check whether the points collapse during generation or after snapping to the road network.

About `direction` / `heading`: you are right that the docs need cleanup. `startDirection` is wrong there; the actual request parameters are `direction` and `heading`.

Current behaviour is:
  - `direction` is a soft round-trip start/focus bearing.
  - `heading` hard-forces the first routed leg more strictly.
  - Neither currently means “place the whole loop west/east/north/south of the start”.

So your example is expected with the current meaning: west is used as the starting/focus direction, but the loop still distributes waypoints around the start to form a closed route. If we want “the whole loop should stay mostly west of the start”, that is a different placement constraint and should be handled as a follow-up improvement, not only as a docs fix.

devemux86:
> I agree that AUTO can take minutes on phones, particularly on very long 170km loops.

Long calculation times also happen on desktop.
Phones just exaggerate the situation.

> The duplicate waypoints generated at the same location could be a real bug, or at least a missing cleanup step. 

They are not just duplicate, there are cases with 7-8 waypoints stacked on same location.

> Could you share the exact URL/parameters for one case?

I make random round trip tests on Android emulator on desktop,
not website, so there is not an easy url to generate.

> Neither currently means “place the whole loop west/east/north/south of the start”.

Thanks for the explanation.
I am used to the round trips of other routers, so expected a similar behavior.

devemux86:
> If we want “the whole loop should stay mostly west of the start”, that is a different placement constraint and should be handled as a follow-up improvement, not only as a docs fix.

This is what other routers do, so most users expect it to work the same way.

For example Kurviger: https://kurviger.com/en/plan

Also, this is how the previous version of BRouter worked, so now we are breaking the user experience (?).
This means that the same parameters no longer work in the same way that application developers expect.

afischerdev:
@devemux86 I'm sorry for the trouble.

I probably didn't test thoroughly enough. And I wanted to go forward for the next version. That is never a good situation.

For testing I used `roundTripLength` for 5, 10, and 15 km for the hiking profile (15, 30, and 45 km for the bikers), mostly offline via CLI.

I did a quick Android test with the old and new logic. The results for the 
old library are 8 to 15 seconds 
new library is around 26 to 67 seconds  for a route length of ~ 180 km, same points.

But that doesn't help us.

What about the option of also using the old routines - e.g. `roundTripAlgorithm=OLD`?

Please remember from [above](https://github.com/abrensch/brouter/pull/903#issuecomment-4586824775):
> !!! adding new feature means not to remove old ones

This way, a client developer can decide which variant to use, for example, offering the user the option for a better route but a longer travel time.

devemux86:
Many use BRouter for driving, so round trips of 200 - 300 km are expected. :slightly_smiling_face:

devemux86:
>What about the option of also using the old routines - e.g. roundTripAlgorithm=OLD?

>Please remember from https://github.com/abrensch/brouter/pull/903#issuecomment-4586824775:

>    !!! adding new feature means not to remove old ones

>This way, a client developer can decide which variant to use, for example, offering the user the option for a better route but a longer travel time.

This is indeed the best solution: continue to offer the previous round trips for backward compatibility in apps.

So the new version has available time to improve and cover what most users expect.

jonnybbb:
I can inspect and debug why `FAST` is even slower than the previous version. I also think there is a routig profile topic here. For example, I did a few tests on Palma Mallorca routing 50km overshoots to > 100km, which might be ok for a motorbike but not a road bike. Kurviger also has some detours and road reuse I would expect Brouter AUTO to resolve to get a higher quality loop. We could also use AUTO to route larger loops with FAST waypoint placement and add an explicit QUALITY option to expect longer calculations. The FAST algorithm should have corrected some of the previous loop-generation errors (placement of waypoints in non-reachable segments, detours, overshoots in length).

devemux86:
It's better the many new classes in `brouter-core` that are related to the round trips,
to be included in a separate package `roundtrip` for easier maintenance:
```java
package btools.router.roundtrip;
```

devemux86:
Checking only the `RoutingEngine`, it now contains new API calls
that do not exist in older Android versions that BRouter supports.
e.g. computeIfAbsent, putIfAbsent, sort

So right now new BRouter does run on older Android versions.
This can be easily checked when you open BRouter in IDEA.

- We should either increase the `minSdkVersion` from `23` to e.g. `26`
- Or fix the new code to use API that works on older Android versions

devemux86:
Pull request https://github.com/abrensch/brouter/pull/943 restores the previous implementation of round trips.
It can be used with the parameter `roundTripAlgorithm=OLD`.

A round trip example with `OLD` in the east direction and what users expect:

afischerdev:
We're now dealing with a somewhat unfortunate situation – not exactly ideal for routing enthusiasts.

- There's an update to restore compatibility with the old version. However, this breaks the tests of the new version, likely due to pre-configuration. I can understand that @devemux86 has little interest in fixing this. #943

- There's another update to improve the requested performance. #944

- There's a desire to release soon due to some error messages. #940

- And there's a note that some routines from this PR aren't compatible with API version 23.

What could we do:

- Go back before #903, fix the other problems, publish, and restart roundtrip improvements with a clear situation.

- Add the old roundtrip routine as the default (like #943 by @devemux86 ) by @jonnybbb, find the breaks, rework routines for API level 23, and after publishing, start reworking the roundtrip routines.
  This gives client developers the chance to test the new round trip routines without the need to use them.

What do you think?

jonnybbb:
I am fine reverting 903 - I will have an improved version in #944 very likely tomorrow or Monday - then we can ship this in a new version

devemux86:
> Go back before https://github.com/abrensch/brouter/pull/903, fix the other problems, publish, and restart roundtrip improvements with a clear situation.

My suggestion would also be this. :+1: 

Pull requests from AI must be tested very well by the users who submit them following some simple rules:
- They must not break backwards compatibility on Android

- They must not replace existing working functions breaking user applications
e.g. old round trip algorithm must remain as option, especially as the new one is not production ready

- They must not break / change how existing features work
e.g. round trip direction must work like other routers and old version from day one, as it is what users expect

- If AI adds too many new classes, they better be in separate packages for easy maintenance
e.g. `package btools.router.roundtrip;`

afischerdev:
What to do next:
- look after #941
- replace entries in doc e.g. link to roundtrip, revisions, ...
- publish some rules for adding new Pull Requests that a developer can look at before push it. And tester can look at it on start testing

jonnybbb:
@devemux86 can you share at least some of your use cases so that I can cover them in some automated test cases.

devemux86:
I don't have specific use cases and I do all tests on Android emulator (not website).

I've seen all the problems right away in the first few random tests.

devemux86:
[Bikerouter](https://bikerouter.de) has a nice implementation of round trips using BRouter.

Round trip documentation: https://docs.bikerouter.de/en/roundtrip-planner/

They also use the correct implementation of direction for round trips:

settings

PR #920, opened 2026-04-16
Support large routing requests via POST and PUT

pcace:
## Summary

This change adds request-body support for standalone `/brouter` routing requests so large `nogos`, `polylines`, and `polygons` no longer have to fit into the URL.

This PR is based on a clean branch from `upstream/master` and contains only the standalone request-body support change.

## What changed

- accept routing parameters via `POST` and `PUT` request bodies for `/brouter`
- keep existing `GET` query parameter support unchanged
- add request-body parsing safeguards based on `Content-Type` and `Content-Length`
- add a configurable `maxRequestLength` server property
- raise the standalone startup script defaults to support bodies slightly above 5 MiB
- document the request-body workflow and file-based `PUT` uploads
- add tests for `POST` body routing and large polygon `PUT` requests

## Notes

- supported routing body content types are `application/x-www-form-urlencoded` and `text/plain`
- the standalone script now exposes the relevant JVM and request-size settings through environment variables

## Validation

- built successfully in the project Docker build
- verified routing with `GET` and `PUT` locally
- verified a real `PUT` request with a body slightly above 5 MiB against a deployed server

### DIFF
--- a/brouter-server/src/main/java/btools/server/RouteServer.java
+++ b/brouter-server/src/main/java/btools/server/RouteServer.java
@@ -38,6 +38,8 @@ import btools.util.StackSampler;
 
 public class RouteServer extends Thread implements Comparable<RouteServer> {
   public static final String PROFILE_UPLOAD_URL = "/brouter/profile";
+  public static final String PROFILES_URL = "/brouter/getprofiles";
+  public static final String ROUTING_URL = "/brouter";
   static final String HTTP_STATUS_OK = "200 OK";
   static final String HTTP_STATUS_BAD_REQUEST = "400 Bad Request";
   static final String HTTP_STATUS_FORBIDDEN = "403 Forbidden";
@@ -83,6 +85,8 @@ public class RouteServer extends Thread implements Comparable<RouteServer> {
       String agent = null;
       String encodings = null;
       String xff = null; // X-Forwarded-For
+      String contentType = null;
+      int contentLength = -1;
 
       // more headers until first empty line
       for (; ; ) {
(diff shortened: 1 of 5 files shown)


EssBee59:
==>  no longer have to fit into the URL

Hello pcace!

Could you please first explain the need for implementing POST requests?
(I am missing defects about that)
The URL max size has been extended again and again in the last years, so that I thought, the GET requests support any real routing request currently?

pcace:
Hi, sorry that it took so long to reply. 
The reason for that change is a problem i am facing in roadbooknavi.de it allows to create quite long routes. But i also want to have the option to for example avoid all nature reserves. To do that, i query data around the route via an overpass endpoint and use the output as a geojson as nogo areas. Sometimes this is wayyyy bigger the the current limitation. To completely get rid of the problem i thought this pr could solve it permanently.

I am currently using it in my fork successfully since the pr and never had an issue with it

Cheers and thanks for reply!

afischerdev:
Thanks for your answer. 
We will not enable this at the moment. We have several server load problems with unknown users and heavy traffic. But we keep the PR open.

I had a look at the roadbook tulips, nice feature. 
But it looks like you will send your data extracts to BRouter and like to get it back with the route. Why not mixing the data from export and route at home?

Is there a way for you to do the routing with the Android app?

Could you provide us with such an export json file?

pcace:
Hi, thanks for your reply! I have a geojson attached as an example. its the nature reserves in brandenburg state germany. There is no way i could use this as nogo areas using the current method because its just wayy to big url then. 

[brb nogo.json](https://github.com/user-attachments/files/31253531/brb.nogo.json)

I think i dont quite get your point here: 

`But it looks like you will send your data extracts to BRouter and like to get it back with the route. Why not mixing the data from export and route at home?
`

basically yes - i get the extracts from overpass osm, and use them in brouter as nogo areas. but what do you mean with mixing data?

Thansks a lot!

afischerdev:
As I said, I had a look on the tulips like this 
Sample Tulip<BR><small>Turn right after
the bridge</small>
And so I started thinking about adding entries for POIs.
What you send is a nogo file for your area - that is fine for me.

pcace:
Thanks for the clarification! I hope i understand your point now correctly: by "mixing the data at home" – I'm actually already doing that in a sense: I run my own BRouter instance, so the POST/PUT support would only be enabled there, not on the public server. That's exactly the reason for this PR: it's not meant to change the public brouter.de behavior, it just makes it possible for self-hosted instances to accept large nogo datasets without URL limits.

The Android app isn't an option for a web app like RoadbookNavi, since routing happens server-side in the backend.

what file do you mean here: ```Could you provide us with such an export json file?``` if you mean finished roadbook files - you can just make one on roadbooknavi.de and export it as you like! json, openrallyGPX, PDF, etc etc.

afischerdev:
@pcace 
Thanks for the clearing.

> it's not meant to change the public brouter.de behavior

Not exactly. When I run your source without any changes, it starts with POST/PUT support.
Only when I set the new parameter -DmaxRequestLength=0 I get a `400: Bad Request`
I would prefer to have another variable e.g. `usePOSTRequests` with default `false`. So the server start scripts need a change, when you like to use the new features.

pcace:
Hi, thanks for the feedback!
i have worked on it again: 

- Request bodies are now opt-in: usePOSTRequests defaults to false; POST/PUT to /brouter returns 405 Method Not Allowed until -DusePOSTRequests=true is set. No behavior change for existing deployments, and the public server is unaffected unless an operator explicitly enables it.
- server.sh change is minimal now (just the opt-in flag), the env-var refactoring was dropped. When enabled, maxRequestLength (default 1000000) caps the body size, e.g. -DmaxRequestLength=6291456.
- Trimmed unrelated scope (OPTIONS/CORS handling for /brouter, getprofiles additions) and added a test asserting the disabled-by-default behavior.
- Branch rebased onto current master; all tests green.

afischerdev:
@pcace 
Thanks for the new version. It works fine.

afischerdev:
@pcace 

pcace:
Thanks a lot!!

PR #923, opened 2026-04-24
Update for Gradle build

afischerdev:
- the final build process for the release revealed that the Gradle version is somewhat outdated. It was version 9.4.1, which generated a bug for `checkstyle WhitespaceAfter` compared to version 8.13.
- to make the review process a little bit clearer a change to the next version preview
- a small jar file was added for `read only` or `routing only` users


afischerdev:
I'll go back to the gradle 8.13 to avoid problems with IDEA.
The problem started with a break in githup actions. So this needs the update.

jonnybbb:
@afischerdev should I try creating a Gradle 9.6.1 upgrade

devemux86:
@jonnybbb We remain compatible with IDEA that currently supports **Android Gradle plugin 9.0.1**.

This just needs **Gradle 9.1.0**: https://developer.android.com/build/releases/agp-9-0-0-release-notes

PR #938, opened 2026-06-30
BInstallerView: appearance and other improvements

devemux86:
- Make the grid visible at all zoom levels

- Use more visible color and width on the grid

- Use semi-transparent color with fill on the tiles

- Start at map center


afischerdev:
@devemux86 
Great, many thanks for the improvement.

afischerdev:
@devemux86 
Thank you very much for the collaboration.

Issue #931, opened 2026-06-05
Attempt to invoke virtual method 'boolean java.lang.String.contains(java.lang.CharSequence)' on a null object reference


noahsmindfuck:
I just updated BRouter from 1.7.8 to 1.7.9 on GrapheneOS, original install via FDroid.
The error in the title appears when switching between route profiles in Osmand.

To reproduce:
1: calculate a new route in Osmand that I have not calculated before - works
2: switch to another BRouter Profile in Osmand - works
3: switch back, eg. toggling between route profiles, the error above shows up and repeats until I do step 1 again. Repeating the request in BRouter app also reveals the same error.

Probably its with something in Osmand, probably with the Brouter-Osmand API interction.

Osmand version 5.3.10, versionCode 531003
BRouter Version 1.7.9 versionCode 56
Android 16, GrapheneOS Kernel 6.1.174-android14-11-gf4e508bdefd0

noahsmindfuck:
It happens when I click on <repeat:fastbike>, the original request comes from Osmand.
<img height="400" alt="Image" src="https://github.com/user-attachments/assets/091e8da5-0c8d-406d-b1f6-3b732ad3a85a" />
<img height="400" alt="Image" src="https://github.com/user-attachments/assets/4810bdfe-c25b-4ceb-9f2d-4dd873cdd3ce" />

EssBee59:
I can not follow, what you are exactly doing:
If you need as Osmand user to switch between many profiles in Brouter, I recommend to use the "new way" for adressing a brouter-profile directly in Osmand (without a change in brouter!)

See documentation:
https://osmand.net/docs/user/navigation/routing/brouter/#osmand-version-471

noahsmindfuck:
Hi essbee59
Thank you for your reply.
I wrote down the steps to reproduce in my initial post.
 The screenshot of the error within the brother app is to show that that the error is not originating in osmand but in bouter.

afischerdev:
@noahsmindfuck 
I tried out your description to reproduce the error message, but without success.
The repeat function becomes available after an app has performed a route calculation using BRouter. This process writes the file `timeoutdata.txt` to the "modes" folder within the BRouter media directories. This file contains the parameters used for routing, allowing BRouter to independently recalculate the route. There may be an error in this context. 
When you have this the next time could you place the text file here. At the first time of recalculation there is also a raw data file of the last calculation. Could your please add that as well. Only present before recalc.

afischerdev:
[...]

PR #939, opened 2026-07-04
Added write/read the way point name

afischerdev:
We sometimes get an NPE when the name is missing. Please see #931


afischerdev:
While experimenting with the new round-trip function, I noticed that the `<repeat profile>` function throws an exception when only a single waypoint is present.

Note:
The check assumes that the ability to repeatedly execute round trips within the app is not required.

PR #943, opened 2026-07-11
Round trips: restore previous implementation as an option

devemux86:
Use the parameter `roundTripAlgorithm=OLD`.

### DIFF
--- a/brouter-core/src/main/java/btools/router/RoundTripAlgorithm.java
+++ b/brouter-core/src/main/java/btools/router/RoundTripAlgorithm.java
@@ -49,7 +49,7 @@ public enum RoundTripAlgorithm {
    * recommended choice for a best-quality loop.
    */
   public static RoundTripAlgorithm fromString(String s) {
-    if (s == null) return AUTO;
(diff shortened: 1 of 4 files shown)


afischerdev:
@devemux86 
Thanks for the update. It will help to publish a new version soon.
I would recommend to set default to OLD.
```
public RoundTripAlgorithm roundTripAlgorithm = RoundTripAlgorithm.OLD;
```
So all clients have the old situation with the first contact and when the like to get other results, there are the caller options.

devemux86:
>  I would recommend to set default to OLD.

Good point, I pushed a commit with the update.

devemux86:
> It will help to publish a new version soon.

@afischerdev Please see my comment here: https://github.com/abrensch/brouter/pull/903#issuecomment-4944330079

New BRouter does not run on older Android versions anymore.

devemux86:
I reverted the set default to `OLD` as all new round trip tests are broken.

@afischerdev It is best to revert BRouter to the previous **working** status.

PR #944, opened 2026-07-11
Round-trip follow-up to #903: BALANCED/QUALITY tiers, context-aware AUTO, planner bug fix, performance budget suite

jonnybbb:
Follow-up to #903, addressing the on-device timing reports in that thread (old logic 8–15 s vs new 26–67 s for ~180 km) and completing the round-trip mode ladder.

## What's in it

**1. Bug fix: NPE on re-detailed guide tracks + error-path contract** (`c361d238`)
`OsmTrack.copyDetours` gutted its *source*: the `FrozenLongMap` constructor moves the source map's arrays out, so a second guided detail pass over the same track object crashed in `registerDetourForId` — destroying an otherwise valid loop mid-detail. Current master ships this. Second layer: an exception escaping `doRoundTrip` could surface as a 0-node "success" track with no error text; the catch now enforces the track-XOR-error contract.

**2. Iso-pool health model + AUTO competition dedup** (`e66e825d`, fork issue jonnybbb/brouter#26)
ISO_GREEDY carries a per-plan health tracker for its candidate pool; a thin/bunched/losing pool is demoted mid-plan down to graph-native candidates, absorbing the separate plain-GREEDY run. AUTO starts GREEDY only when the ISO result proves the comparison still useful — removing the duplicate planner run from the common case. The internal comparison branch runs the same ladder order as the recursion it replaces (your undershoot sentinel caught a contraction regression from the fewer-steps-first order during validation — exactly what it was built for; fixed and documented in the commit).

**3. BALANCED tier** (`0fa1b663`, fork issue jonnybbb/brouter#27)
`roundTripAlgorithm=BALANCED`: one graph-aware planning run under a hard `min(request budget, 8 s)` per-slice clamp with reduced per-step search (top-K 2/3), no retry ladders, geometric fallback only when no loop closes at all. Degraded results return with the lenient `Warning:` disclosure. This is the recommended interactive default for phones.

**4. QUALITY tier + context-aware AUTO** (`94cb82d7`)
The full ladder is now `FAST < BALANCED < AUTO < QUALITY`. QUALITY = the competition at max effort (both planners always, top-K 4/6, doubled plan budget) — deliberately *not* a forced-ISO_GREEDY alias, since the radial planner still wins ~a quarter of matrix cells. AUTO resolves its effort from request context and logs the decision: a request budget ≤ 10 s or `memoryclass` ≤ 48 resolves to BALANCED-grade bounded effort (the direct answer to the phone timings); profile class is read from the profiles' own `validForFoot/Bikes/Cars` globals (motorized profiles get a provisional-quality advisory — no loop corpus covers them yet); length class is classified and logged so future tuning lands on evidence.

**5. Performance budget suite** (`3d1f00db`) — see metrics below. Work-unit ceilings (`linksProcessed`, routed candidates) and the BALANCED latency contract, asserted over region anchors; wall-clock recorded to `build/reports/perf/roundtrip-perf.jsonl` for trend-watching, never asserted. Includes a `memoryclass=32` axis.

**6. Build: pin the build JDK to 17** (`21a1e486`) — PMD 7.0.0 mass-produces `LooseCoupling` false positives under newer JDKs (verified A/B: plain master fails 7 pmd tasks under Java 25, is green under 17). Toolchain + daemon-JVM criteria make `./gradlew build` green from any environment; bytecode stays Java 11. Happy to split this into its own PR if preferred.

## Performance metrics (M-series dev machine, local segments; wall-clock is machine-relative, ratios and work units are the signal)

**Tier ladder, Basel start, fastbike, fixed direction:**

| Request | FAST | BALANCED | AUTO | QUALITY |
|---|---|---|---|---|
| 180 km | 2.2–3.5 s | 6.6–6.8 s, 172.4 km | 11.3–11.7 s, 175.8 km | — |
| 15 km | 0.4 s | 1.2 s | 1.5 s (identical loop to BALANCED) | — |

**24-cell sweep** (Freiburg/Basel/Nice/Mallorca/Girona/Garmisch × 50/100 km × fastbike/gravel): BALANCED's worst cell = 8.04 s (its cap held everywhere) vs AUTO's worst = 20.7 s; total time −27%; mean distance-to-target miss 9.8% vs AUTO's 8.2%; 1/24 cells returned a disclosed `Warning:`. Budget-exhaustion contract: a 250 km request under BALANCED returns a disclosed 147.7 km best-effort loop in 7.0 s instead of nothing.

**Work units** (`linksProcessed` = Dijkstra link expansions, deterministic per request — the `memoryclass=32` twins reproduce byte-identical counts):

| Cell | FAST | BALANCED | AUTO | QUALITY |
|---|---|---|---|---|
| Basel 15 km fastbike | 91 k | 4.04 M | 4.58 M | — |
| Basel 100 km fastbike | — | 348 k | 867 k | 627 k (routes 94.1 km vs AUTO's 87.5 km on the 100 km ask) |
| Garmisch 50 km gravel | 105 k | — | 167 k | — |

(The 15 km dense-urban cell costing 11× the 100 km loop is a known follow-up optimization target the suite surfaced — every retry rung fits the small-scale budget.)

**A practical note for on-device timing comparisons:** rerunning a CLI command into the same output file advances to the *next alternative route* (`doRouting`'s output-file cycling), so repeated timings against one output file measure different routes. The perf suite runs with no output file; worth ruling out in manual old-vs-new comparisons too.

## Validation

- Full unit suite (707 tests), golden signature test, and the undershoot sentinel green.
- `./gradlew build --rerun-tasks` fully green — PMD, checkstyle, all module tests, no skipped gates.
- Full loop-quality integration matrix: green on this exact head — 933 tests, 0 failures, 273 terrain-scoped skips (40 min run, includes the new perf-budget cells).

Reviewer pointers: `docs/features/roundtrips.md` (ladder + context rules), `RoundTripEffortPolicy` (the presets and resolution), `RoundTripPerfBudgetTest` (the metrics harness).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

### Why FAST regressed vs. the pre-#903 round trip

The old (pre-#903) routine placed a geometric narrow-arc ring (`buildPointsFromCircle`) and routed — nothing else. #903's FAST/WAYPOINT path added two node-matching passes the old routine never had: a 24×3 reachability **probe**, then a per-via candidate-snapping pass (`validateAndAdjustWaypoints`, ~21 candidates/via). Those ~160–180 extra graph matches per request scale with loop radius and were the **entire** slowdown — measured ~1.8–2.3× the old routine, growing with loop size (matching the on-device reports). The routing itself was unchanged; only placement got heavier.

### What the rework changes (`-Droundtrip.fast.optimized`, default on; AUTO/QUALITY/ISOCHRONE untouched)

- **Reuse the probe's snapped nodes as vias** — the probe already road-snaps points per direction; use those directly instead of re-generating geometric points and re-matching them. Eliminates `validateAndAdjustWaypoints` (and its cache reset) for FAST.
- **Trim the probe grid** 24×3 → 12×2.
- **Reachability guard** — each via is tested with the same bounded `findTrack` island primitive the router uses; vias stranded on a small disconnected component are dropped before routing (best-effort; a budget timeout keeps the via). Falls back to the circle path if fewer than 3 reachable vias survive.
- **Dedup** coincident vias — fixes the stacked-waypoint report.

### Performance matrix

Warm A/B in a single JVM (real `E5_N45` tile, `car-fast`/`fastbike`, fixed `direction`); machine-relative M-series dev box — ratios are the signal, not absolute ms.

| Loop | old (pre-#903) | FAST before rework | **FAST optimized** | vs. before | vs. old |
|---|---|---|---|---|---|
| 20 km bike | 211 ms | 247 ms | **229 ms** | 1.08× | ≈ matched |
| 50 km bike | 198 ms | 575 ms | **391 ms** | 1.47× | ~2× |
| 180 km car | 663 ms | 2150 ms | **~830 ms** | **~2.6×** | +25% |

The gain grows with loop size (removed overhead scales with area); at the 180 km motorized scale FAST is back to roughly the old routine's speed.

### Bugs fixed alongside

- **Stacked waypoints** (multiple bearings snapping to one node): Karlsruhe 180 km 5–8 stacked positions → **0**.
- **Island robustness**: Basel 180 km car **0/8 → 8/8** directions form (better than the pre-rework 5/8) — the guard drops the islanded via instead of failing the loop.

707/707 `:brouter-core:test` green.


afischerdev:
@jonnybbb 
Based on the experience with our last pull request, I have put together a developer guide.
Therefore, I would like to ask you to close this PR and follow the guide for a new one.  To go back to a clear repository
Thank you for your effort.

jonnybbb:
@devemux86 referring to [your comment](https://github.com/abrensch/brouter/pull/948#issuecomment-4966064974)

**Graphhopper** uses an [initalHeading](https://github.com/graphhopper/graphhopper/blob/f866a9c27810911bafa0eb8d0b75a9416d1de48a/core/src/main/java/com/graphhopper/routing/RoundTripRouting.java#L52) that determines the first leg direction 

The **openrouteservice** doesn't even have a direction, but only length, seed and points

I don't have access to the source of other roundtrip routing services.

devemux86:
GraphHopper does not have full round trip functions.
Kurviger, which is derived from GraphHopper, has proper round trip functions and is the standard.

OpenRouteService website does not offer all the round trip options.
OpenRouteService API allows to set the direction of the whole round trip (option bearings).

Anyway, this is the correct way to use the direction option in the round trips.
We must follow what BRouter and other routers already offer as this is what users expect.

PR #946, opened 2026-07-13
Add surface=laterite to lookups.dat (mirror existing clay/mud handling)

julcnx:
## Summary

`surface=clay` and `surface=mud` are already recognized values in `misc/profiles2/lookups.dat`, but `surface=laterite` is not. Untagged, it currently gets encoded as the generic `unknown` surface bucket, so it's indistinguishable from other unrecognized surface values in any profile.

[`surface=laterite`](https://wiki.openstreetmap.org/wiki/Tag:surface%3Dlaterite) is a natural (lateritic soil) road surface, common on unpaved roads across tropical regions, and is tagged in the same spirit as `ground`/`dirt`/`clay`/`mud` (approved via OSM's proposal process ([Proposal:Surface=laterite](https://wiki.openstreetmap.org/wiki/Proposal:Surface%3Dlaterite), 29-0-0).

## Changes

- `lookups.dat`: add `surface;0000000001 laterite`, alongside the other low-frequency surface values (`rock`, `stone`), using the same placeholder count convention for values that are rare in the German reference dataset the counts are drawn from.
- `gravel.brf`: add `laterite` to the two switch lists where `clay` and `mud` already appear together (the wet-conditions smoothness penalty, and the `is_paved` exclusion list).
- `vm-forum-liegerad-schnell.brf` / `vm-forum-velomobil-schnell.brf`: same addition to the `surfacepenalty` switch list where `clay` and `mud` already appear together, plus the matching German comment update.

No other profiles reference `clay` and `mud` together, so no other files were touched. `mtb.brf` references `mud` alone (without `clay`) in a couple of spots for MTB-specific preference weighting — I left those alone since that's a distinct judgment call about how laterite specifically rides, not a mechanical parity fix.

## Testing

I wasn't able to run the Gradle test suite locally (this repo requires Java 17+, and I only had 16/8 available here), so `IntegrityCheckProfileTest` / `EncodeDecodeTest` haven't been verified locally. The change is a single new lookup-table line plus four one-token switch-list additions, mirroring existing entries, but please let CI confirm and flag if anything about the lookup-table format needs adjusting.

### DIFF
--- a/misc/profiles2/gravel.brf
+++ b/misc/profiles2/gravel.brf
@@ -124,7 +124,7 @@ assign smoothnesspenalty
         switch highway=cycleway                                                                1.2
         # handle non-specific tags last
         switch surface=unpaved                                  switch not assume_wet_conditions 1.6 3.0
-        switch surface=ground|grass|dirt|earth|mud|clay|sand    switch not assume_wet_conditions 2.6 4.0
+        switch surface=ground|grass|dirt|earth|mud|clay|laterite|sand    switch not assume_wet_conditions 2.6 4.0
         # default smoothness
                                         switch not assume_wet_conditions 1.6 3.2
 
@@ -168,7 +168,7 @@ assign tracktypepenalty
                                1.0
 
 assign is_paved
-    and not surface=ground|grass|dirt|earth|mud|clay|sand|pebblestone|gravel|fine_gravel|compacted
+    and not surface=ground|grass|dirt|earth|mud|clay|laterite|sand|pebblestone|gravel|fine_gravel|compacted
     or highway=primary|primary_link|secondary|secondary_link|tertiary|tertiary_link
     or surface=asphalt|concrete|paved|paving_stones|sett|cobblestone|metal
     or highway=road|unclassified|residential|living_street|service
--- a/misc/profiles2/lookups.dat
+++ b/misc/profiles2/lookups.dat
@@ -61,6 +61,7 @@ surface;0000004398 clay
 surface;0000003760 metal
 surface;0000000001 rock rocks rocky
 surface;0000000001 stone
+surface;0000000001 laterite
 
 lit;0000809223 yes
 lit;0000000001 no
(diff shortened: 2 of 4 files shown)


afischerdev:
Thanks for your suggestion.

Two steps are required to complete your task:
- a new lookups.dat
- entries in profiles that use the new tag value

I will create a PR for a new `lookups.dat` and check for any other pending requests regarding the `lookups.dat`.
The data will then be generated, and once it is online, the profiles can be checked in.
Please remove the `lookups.dat` update from here; I will create a separate PR for it.

julcnx:
Thanks for the quick response and for taking this on.

I've removed the `lookups.dat` change from this PR (just the added `surface;0000000001 laterite` line), so it now only touches the profile files (`gravel.brf`, `vm-forum-liegerad-schnell.brf`, `vm-forum-velomobil-schnell.brf`). Happy to leave this open until the new `lookups.dat` is online, or close/reopen later if that's easier for you to track, whichever you prefer.

afischerdev:
Thanks for your change.
The profiles are checked in when the data is online.
At the moment you can test it on the test server please see #951

afischerdev:
Sorry, my fail, the workflow run was to early. I'll repeat it later.

julcnx:
Just checking in, it's been about a week since the last update here. I know #951 needs the workflow run repeated before the new `lookups.dat` can go online and this can be checked in, no rush from my side, just wanted to bump this in case it slipped off the radar. Happy to help test once the data's live.

afischerdev:
@julcnx 
#951: the others need also a chance for a review or test.
You could test against the [test server](https://brouter.de/brouter-test). This is already working there.

julcnx:
> You could test against the [test server](https://brouter.de/brouter-test). This is already working there.

Tested this on the test server (brouter.de/brouter-test) and it's resolving correctly: `surface=laterite` shows verbatim in the route's WayTags for the laterite segments I crossed, with CostPerKm around 10481, roughly double a plain `surface=unpaved` path (~5241/km) on the same route and about 2.8x a paved `surface=concrete` segment (~3690/km). Looks like it's behaving as intended, thanks for getting this checked in.

afischerdev:
@julcnx 
>  and it's resolving correctly
Thanks for the report.

PR #948, opened 2026-07-13
Added Developer Guide

afischerdev:
Based on discussions in several pull requests, here is a first version of a developer guide.

### DIFF
--- a/docs/developers/developer_guide.md
+++ b/docs/developers/developer_guide.md
@@ -54,11 +54,16 @@ There are thus three levels:
 
 - there are several docs, please have a view into the doc folder
 
+- the main source is in poor Java
+  A new feature should follow this way and not add more dependencies.
+  If it is unavoidable, please discuss this before adding it.
+
 
 ### Testing
 
 - keep in mind this software is used as a server and as an Android service
-  That means it should return for on the same input the same output as before.
+  That means it should return on the same input the same output as before when a new feature is added.
+  This will not work if the task is to change the output.
   New features normally have new parameter to activate them.
 
 - running `gradlew clean build` in your working copy generates several reports.
@@ -72,8 +77,23 @@ There are thus three levels:
 
 - new features must work like other routers and old version from day one, as it is what users expect
 
-- if AI adds too many new classes, they better be in separate packages for easy maintenance
+- if the pull request adds too many new classes, they better be in separate packages for easy maintenance
 
+- BRouter operates as a service: a client sends a request, BRouter responds, and the client proceeds to use the result.
+  This means that timing should be monitored whenever the code is modified or extended.
+
+### Using AI
+
+- feel free to use AI for your development
+
+- the description of a new feature should be done by the developer
+
+- you should understand each line the AI generates
+  Means you have to review all code before check in.
+
+- keep your AI outside the repository
+
+- check if AI licence meets BRouter [MIT licence](https://github.com/abrensch/brouter/tree/master?tab=MIT-1-ov-file)
 
 ### Review
 
--- a/docs/developers/developer_guide.md
+++ b/docs/developers/developer_guide.md
@@ -1,0 +1,82 @@
+---
+parent: Developers
+title: Developer Guide
+---
+
+
+### Starting point
+
+A Git account is required to participate in this project. This allows you to create a personal fork of the BRouter project and set up a development branch, which is then cloned to your development PC. From there, a working copy is created for actual development.
+
+There are thus three levels:
+- BRouter master
+
+- Personal BRouter clone for the new feature/error handling
+  this maintains the connection to the master
+
+- Personal BRouter working copy
+  This is the area where the actual work takes place.
+  Here you have everything you need for your new feature, including helper classes, extra resources, and more.
+
+
+#### What does that mean
+
+- the local branch and the working copy have no contact
+  Working copy has no connect to git.
+
+- all work is done in the working copy.
+  No one is interested to see what you are doing there.
+
+- all testing is done in the working copy.
+  No one is interested to see what you are doing there.
+
+- when you find your new routines/changes are fine, copy only the changed or new files to the local branch
+  Test if this is compileable by `gradlew clean build`.
+
+- most new features need a test routine as well
+  Generate a test for general functionality like test for null, 0, -1, +1, Integer.Max Value something like that, to signal your function is fine.
+  Do not add more resources.
+  Test if this is compileable by `gradlew clean build`.
+
+- before the final check in for a pull request, check if a doc is needed
+
+
+#### Some useful rules for a start
+
+- plan the new features as small as possible.
+  A discussion before could be helpful.
+  Save resources, means: save project space on git, save running actions on git, save time for reviews.
+
+- implement only for your plan
+  New ideas can be done later in a new plan.
+  You may find errors along your way that have nothing to do with your plan.
+  Leave it as it is. May be collect it for a later review and a new plan.
+
+- there are several docs, please have a view into the doc folder
+
+
+### Testing
+
+- keep in mind this software is used as a server and as an Android service
+  That means it should return for on the same input the same output as before.
+  New features normally have new parameter to activate them.
+
+- running `gradlew clean build` in your working copy generates several reports.
+  Check them regularly.
+
+- new features must not break backwards compatibility on Android
+
+- new features must not replace existing working functions breaking user applications
+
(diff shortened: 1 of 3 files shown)


docs/developers/developer_guide.md
```diff
@@ -0,0 +1,82 @@
...
+  You may find errors along your way that have nothing to do with your plan.
+  Leave it as it is. May be collect it for a later review and a new plan.
+
+- there are several docs, please have a view into the doc folder
+
+
+### Testing
+
+- keep in mind this software is used as a server and as an Android service
+  That means it should return for on the same input the same output as before.
```
jonnybbb: How about this: if the change is to improve a feature, the output should change intentionally. I would argue that a change should not break existing test cases without discussing with the maintainers beforehand

docs/developers/developer_guide.md
```diff
@@ -0,0 +1,82 @@
...
+  New features normally have new parameter to activate them.
+
+- running `gradlew clean build` in your working copy generates several reports.
+  Check them regularly.
+
+- new features must not break backwards compatibility on Android
+
+- new features must not replace existing working functions breaking user applications
```
jonnybbb, line 74: How do you know? This should be covered by existing test cases

docs/developers/developer_guide.md
```diff
@@ -0,0 +1,82 @@
...
+
+- new features must not break / change how existing features work
+
+- new features must work like other routers and old version from day one, as it is what users expect
```
jonnybbb, line 78: This is very vague and hard for someone to figure out, and also debatable, like this kurviger.com example https://github.com/abrensch/brouter/pull/903#issuecomment-4936830938 - this is meant to be used for motorcycles, but this repo doesn't even have a motorcycle profile.

docs/developers/developer_guide.md
```diff
@@ -0,0 +1,82 @@
...
+
+- if AI adds too many new classes, they better be in separate packages for easy maintenance
```
jonnybbb: I would make this AI independed and try to define some general architectural rules on how the codebase is structured. 
AI-generated code must include the coding agent version and the model that has been used in any commit.

docs/developers/developer_guide.md
```diff
@@ -0,0 +1,82 @@
...
+  New features normally have new parameter to activate them.
+
+- running `gradlew clean build` in your working copy generates several reports.
+  Check them regularly.
+
+- new features must not break backwards compatibility on Android
+
+- new features must not replace existing working functions breaking user applications
+
+- new features must not break / change how existing features work
```
jonnybbb, line 76: Contributors assume this is covered by the test cases

docs/developers/developer_guide.md
```diff
@@ -0,0 +1,82 @@
...
+### Testing
+
+- keep in mind this software is used as a server and as an Android service
+  That means it should return for on the same input the same output as before.
+  New features normally have new parameter to activate them.
+
+- running `gradlew clean build` in your working copy generates several reports.
+  Check them regularly.
+
+- new features must not break backwards compatibility on Android
```
jonnybbb, line 72: Can you add some basic (manual) tests  that are expected to work, and how to verify this

docs/developers/developer_guide.md
jonnybbb: I don't get the correlation between the two. It is crucial that the code must work in both environments, and the crucial information here to be aware of is resource availability and  the compatibility

jonnybbb (commented):
Thanks for putting this together. As someone mostly unfamiliar with the entire codebase I found the rules too vague and less actionable. I would also expect to have a better safety network with proper tests that cover those rules and contributors can rely on.

docs/developers/developer_guide.md
afischerdev: No, we don't know our clients and what they do with the output.
So we can't build test for them. 
But when we plan a new feature we add it to the old ones and do not replace:
```
switch(do) {
  switch feature1: ...
  switch feature2: ...
  switch featureNew: ...
}
```
But you are right, it is not clear that his means the old features.

docs/developers/developer_guide.md
afischerdev: As before, we don't know what clients do. But we don't want to ruin their work. The goal is not to replace old functions; we are adding functions.

docs/developers/developer_guide.md
afischerdev: You are right, that is independed from AI.

> AI-generated code must include the coding agent version and the model that has been used in any commit.

What you do in your working copy is yours. 
You could add a link to the AI, but when we need to see all the mess like in #903 or #944 or #942, I would recommend to exclude AI from this project.

jonnybbb: Shall I still continue contributing with the help of AI or do should I keep the work in a separate fork and not contributing back?

afischerdev: You started to bring a new roundtrip logic from a text to code. When AI can do that for you, that is a fine thing. 
But whatever an AI do for you, you have to check it for all the things that belongs to a project before commiting and leave this work for others.

Please continue your work with fresh ideas. 
Do all you like to do in your working copy and add only final and working results to the repository.

docs/developers/developer_guide.md
jonnybbb: I fully agree to backwards compatibility - I think we better use this term - this also opens up for keeping the API stable but improve the actual result

docs/developers/developer_guide.md
jonnybbb: Perfect - my intent to open the PR as draft was mainly to allow others to see there is some ongoing work I never pushed back to the repository (I do not have permissions ) just opened a draft PR - but I am fine to even open a draft only when I am considering it done - I think for new contributors early feedback is helpful - as was yours - I never intended to leave work for others. But the contribution guidelines should be cristal clear of the expectations and have a mature test coverage otherwise new contributors do have any chance to get it right

afischerdev: I'm sorry to say, no, not perfect. May be the intention was good, but the result was terrible. Please see [actions list](https://github.com/abrensch/brouter/actions) and your PR #944. Full of commits, fault or ok. Nobody needs your working results, except you. Or [deployments list](https://github.com/abrensch/brouter/deployments). That is what I mean with `save resources`. Keep your AI outside the repository.

jonnybbb: I am wondering why those builds are  triggered in the first place for draft PRs - I see your point why this resulted in some frustration on your side. 
What I meant with "perfect" was that you agreed that I should continue my work with fresh ideas - not that it was perfect

afischerdev: Yes, it would be a shame if this beautiful idea for new round trips were abandoned.

jgillich: AI is a great tool but it's important that humans remain in control. Cleaning up the test bloat probably would've taken just a single prompt. I really like Maplibre's AI policy, perhaps it could serve as a template https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md

I do agree that draft PRs are useful, early feedback can help so you don't end up building the wrong thing. If they create a burden on the infrastructure, they should be excluded

jonnybbb: @afischerdev here is a PR https://github.com/abrensch/brouter/pull/950 to exclude draft PRs from the CI pipeline

docs/developers/developer_guide.md
jonnybbb: This was more referring to 
> must work like other routers

Which is hard to figure out in particular, if they have a different target group, e.g., motorcycles/cars vs bicycles/pedestrians -

devemux86:
Round trips work the same on all routers in all travel modes:

- **The direction parameter is not the direction of departure, but the direction of the entire round trip.**

e.g. BRouter, GraphHopper, Kurviger, OpenRouteService, RouteYou, ...

This is how the existing direction parameter currently works and must continue to work
to avoid breaking application compatibility and user expectations and experience.

For example a correct round trip with direction East:

docs/developers/developer_guide.md
devemux86: > As before, we don't know what clients do. But we don't want to ruin their work. The goal is not to replace old functions; we are adding functions.

:+1: 

Like the last round trip with AI, it replaced incorrectly the existing round trip instead of add new algorithm,
it worked differently and incorrectly (i.e. direction) and it was very slow, leading to a poor user experience.

PR #951, opened 2026-07-15
Update lookups.dat Version 11.3

afischerdev:
Before we do a greater update on the lookups.dat here a small fix for some wishes:

- #928 seperate forestry and agricultural
- #946 add surface=laterite
- #900 add cycleway=asl
- #937 rework maxspeed
  `rural `and `urban` entries now return a country specific value
  In version 11.3 the tags `rural` and `urban` are still in the lookups.dat to prevent failures in the profiles - the standard profiles do not include them.
  This will change in `version 12.1`, please therefore remove `rural` and `urban` from your profile - if you use it today.

The test server already uses this lookups.dat, so please test your wishes there
Examples:
Way with [rural in DE](https://www.openstreetmap.org/way/70483382)
[Route](https://brouter.de/brouter-test/#map=14/49.7733/8.7577/standard&lonlats=8.770349,49.776442;8.740653,49.771897&profile=car-eco)
Way with [rural in FR](https://www.openstreetmap.org/way/279319472)
[Route](https://brouter.de/brouter-test/#map=12/47.7024/7.0544/standard&lonlats=7.039089,47.721115;7.098701,47.715016&profile=car-eco)

PR #956, opened 2026-07-30
Alias access=unknown to destination via lookups.dat

julcnx:
access=unknown (https://wiki.openstreetmap.org/wiki/Tag:access%3Dunknown, a real, documented OSM convention meaning access conditions weren't verified and need survey) isn't referenced anywhere in `defaultaccess` in `trekking.brf`, `mtb.brf`, or `skating.brf`. It falls through to the `else true` branch and is treated identically to `access=yes`.

This adds `unknown` to the existing `access=private|no` false-branch in both the way-context and node-context `defaultaccess` assignments in all three files.

I looked for a softer option first (these engines have a graduated cost tier for surface/tracktype quality via `probablyGood`/tracktype grade, 1.0x-5.0x), but there's no equivalent graduated tier for access legality, it's strictly binary (full allow vs. cost 10000/blocked). Extending the existing block-list was the smallest change available without inventing new profile logic. I recognize that's a blunter outcome than the tag's actual meaning ("unverified", not "known restricted"), noted here so it's a deliberate, visible tradeoff rather than a silent one. Since these are user-editable text profiles, anyone who wants different handling can adjust their own copy.

Not touched: `car-vario.brf` already effectively blocks `access=unknown` on tracks by default (it only allows tracks via an explicit `access=yes|permissive|designated|destination` allowlist or `use_offroad`), so no change needed there. Also not touched: `hiking-mountain.brf`, `fastbike.brf`, `fastbike-verylowtraffic.brf`, `shortest.brf`, `vm-forum-liegerad-schnell.brf`, `vm-forum-velomobil-schnell.brf`, these also define their own `defaultaccess` but with different logic I didn't want to guess at editing blind.

I didn't find an existing isolated unit test exercising `defaultaccess`'s private/no handling to extend (`brouter-expressions`'s test profiles don't cover access logic), and building/running the full routing-integration tests wasn't something I could do in my environment, so this is untested beyond manual review of the `.brf` syntax against the surrounding unchanged lines. Flagging that plainly rather than claiming it's verified.

This PR (code and description) was drafted with the help of an AI coding assistant (Claude, Anthropic, Claude Sonnet 5), based on manual reading of the profile files listed above. Please review accordingly.

### DIFF
--- a/brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java
+++ b/brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java
@@ -130,6 +130,7 @@ public class OsmCutter extends MapCreatorBase {
     }
 
     if (n.getTagsOrNull() != null) {
+      generatePseudoTags(n.getTagsOrNull());
       int[] lookupData = _expctxNode.createNewLookupData();
       for (Map.Entry<String, String> e : n.getTagsOrNull().entrySet()) {
         _expctxNode.addLookupValue(e.getKey(), e.getValue(), lookupData);
@@ -149,6 +150,14 @@ public class OsmCutter extends MapCreatorBase {
 
 
   private void generatePseudoTags(Map<String, String> map) {
+    // access=unknown collides with the reserved lookup index used for
+    // absent/unrecognized values (see BExpressionContext#addLookupValue),
+    // so it decodes identically to no access tag at all. Rewrite it here
+    // so lookups.dat can alias it to a distinct, non-fallback value.
+    if ("unknown".equals(map.get("access"))) {
+      map.put("access", "value_unknown");
+    }
+
     // add pseudo.tags for concrete:lanes and concrete:plates
 
     String concrete = null;
--- a/misc/profiles2/lookups.dat
+++ b/misc/profiles2/lookups.dat
@@ -135,7 +135,7 @@ access;0002688349 private restricted residents employees
 access;0000319927 yes public
 access;0000144799 no
 access;0000140215 permissive
-access;0000108802 destination visitors
+access;0000108802 destination visitors value_unknown
 access;0000099899 agricultural forestry agricultural;forestry
 access;0000039934 designated official
 access;0000011813 customers
@@ -939,7 +939,7 @@ access;0000078544 private
 access;0000014933 yes public
 access;0000014456 no
 access;0000008670 permissive
-access;0000006316 destination customers delivery
+access;0000006316 destination customers delivery value_unknown
 access;0000000973 agricultural forestry
 access;0000000942 designated official
 access;0000000001 permit
--- a/misc/profiles2/mtb.brf
+++ b/misc/profiles2/mtb.brf
@@ -219,7 +219,7 @@ assign any_cycleroute
 
   assign defaultaccess =
          if access= then not motorroad=yes
-         else if access=private|no then false
+         else if access=private|no|unknown then false
          else true
 
   assign bikeaccess =
@@ -559,7 +559,7 @@ assign any_cycleroute
   assign defaultaccess =
          if ( access= ) then  if allow_default_barrier_restriction then not barrier=gate # add default barrier restrictions here! - modified v2.5.15, conditioned in 2.6
                                                                    else true
-         else if ( access=private|no ) then false
+         else if ( access=private|no|unknown ) then false
          else true
 
   assign bikeaccess =
--- a/misc/profiles2/mtb.brf
+++ b/misc/profiles2/mtb.brf
@@ -219,7 +219,7 @@ assign any_cycleroute
 
   assign defaultaccess =
          if access= then not motorroad=yes
-         else if access=private|no|unknown then false
+         else if access=private|no then false
          else true
 
   assign bikeaccess =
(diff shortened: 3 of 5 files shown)


mcliquid:
Why should a value listed as "unknown" be equated with "no"? Just because it's unknown doesn't automatically mean it's "no." I think that assumption is completely wrong. To me, the "unknown" value is actually a troll tag. "Unknown" is the same as if the tag were completely missing—in that case, we don't automatically assume it's "no" either.

lpirl:
Agree with unknown != no. [OSM WIki](https://wiki.openstreetmap.org/wiki/Tag%3Aaccess%3Dunknown) does not agree with `access=unknown` being a troll. Known unknowns might not matter for some uses (e.g,, routing) but might matter for others  (e.g,, judging data quality).

julcnx:
> Why should a value listed as "unknown" be equated with "no"?
> Agree with unknown != no.

Fair, and you're right, blocking outright was the wrong instinct here.

`car-vario.brf` already has a tier that fits better: `access=destination` isn't blocked there, it's allowed but cost-multiplied 64x (`caraccess_destination`, line 286), so the router still reaches it if that's genuinely where you're going, without treating it as a through-route. That's a much closer match to what `access=unknown` actually means, especially since a lot of it comes from unverified imagery tracing, where nobody's confirmed on the ground whether a road or track is passable or even connects at the far end.

Revised plan: add `unknown` alongside `destination` in `car-vario.brf`'s `access` switch and `caraccess_destination`, reusing the existing mechanism directly. For `trekking.brf`/`mtb.brf`/`skating.brf`, `defaultaccess` is strictly boolean, no cost tier exists to reuse, so matching this for bike/foot means adding a small one rather than reusing something already there. Wanted to check that's the right direction before touching it.

afischerdev:
There is not only the wiki `unknown` value.

If you look at the lookups.dat file, you will notice that BRouter does not use all available values.
All unused but available values are marked as 'unknown'. So a value `access=motor_vehicle` will produce an `access=unknown` for BRouter. You could test this with the class [EncodeDecodeTest](https://github.com/abrensch/brouter/blob/master/brouter-expressions/src/test/java/btools/expressions/EncodeDecodeTest.java)

I would recommend not to use the value `unknown`.

julcnx:
> I would recommend not to use the value `unknown`.

You're right, and digging further confirmed it's worse than a `.brf`-matching issue: `"unknown"` is the hardcoded fallback name BRouter gives every unrecognized value for a given key, so a `lookups.dat` entry literally named `unknown` collides with that fallback instead of getting its own index. There's no way to make it distinct at the profile/`lookups.dat` level.

I'd still like to find a way to support it though, `access=unknown` is a documented, actively-used tagging convention ([wiki](https://wiki.openstreetmap.org/wiki/Tag:access%3Dunknown)), and it already gets distinct handling in OSRM, GraphHopper, and Organic Maps, so BRouter falling through to `access=yes` behavior is increasingly the outlier.

Proposed direction: rather than touching the reserved fallback, add a pseudo-tag at map-creation time (same mechanism already used for `concrete:lanes`/`concrete:plates` in `generatePseudoTags`, `OsmCutter.java`), so a literal `access=unknown` gets rewritten to a new synthetic tag like `accessunknown=yes` before it ever reaches the generic-unknown encoding step. That gives it a genuinely distinct index in `lookups.dat` with no collision risk, and `car-vario.brf`/`trekking.brf`/`mtb.brf`/`skating.brf` can key off it safely.

Wanted to check this direction is acceptable before writing it up as a diff, since it touches `OsmCutter.java` rather than just profile text. Let me know if you'd rather I take a different approach.

afischerdev:
@julcnx 
`generatePseudoTags` could be a solution.
What about:
Move an incoming `unknown` to `value_unknown` and use it as alias
```
access;0000319927 yes public value_unknown 
access;0000144799 no
```
You just have to decide which is correct: yes/no/diverse.
No need to change the profiles.

julcnx:
Thanks, that's a lot simpler than touching OsmCutter.java. One thing I want to make sure I understand though: if value_unknown gets folded into the yes bucket (or no), does that mean profiles can only treat it identically to whichever one it's aliased to, or is there a way to give it its own distinct encoded value even while sharing the alias line? The reason I'd asked for a separate index in the first place is that car-vario.brf already has a cost-multiplied "destination" tier that fits access=unknown better than a flat yes or no, and I'd lose access to that if it collapses into a binary bucket. If diverse isn't on the table at the lookups.dat level, yes is probably the safer default of the two, but happy to defer to whatever's simplest to maintain on your end.

afischerdev:
@julcnx 
Yes, we could add this to destination as an alias like this
```
access;0000108802 destination visitors value_unknown
```

afischerdev:
@julcnx 
Looks good for me. 
If you don't mind, I would like to keep the lookups.dat updates on one place #951. So could you please remove lookups.dat here, I'll add this on #951.

Was it intended to close this?

julcnx:
Sorry for the noise, closed this by accident renaming the branch on my fork. Continued as #961, which now carries just the OsmCutter.java rewrite of access=unknown to value_unknown, per afischerdev's request below I dropped the lookups.dat change there so it stays consolidated in #951.

PR #957, opened 2026-08-02
Add configurable unpavedPenalty parameter to trekking.brf

sebastienaubry:
The non-probablyGood cost values for track/road/path/footway ways (by tracktype grade) were previously hard-coded (1.3, 2.0, 3.0, 5.0, 5.0). This introduces a single unpavedPenalty multiplier so users can tune how strongly unpaved/low-quality surfaces are avoided, without editing the profile source.

Default value (1.0) reproduces the exact previous cost values, so this is a non-breaking change for existing users of the profile.


afischerdev:

PR #960, opened 2026-08-05
New Roundtrip Waypoint Type and new parameter for lonlats

afischerdev:
At the moment round trips are fixed to `engineMode=4`.
This enables the round trip on the `engineMode=0` as well. This is done by the call
`... lonlats=8.784465,49.779428,r  ...
`
and now we can build combinations for this
```
... lonlats=8.7806257,49.7698181;8.784465,49.779428,r ...
... lonlats=8.784465,49.779428,r;8.7806257,49.7698181 ...
... lonlats=8.7806257,49.7698181;8.784465,49.779428,r;8.7806257,49.7698181 ...
```
It should not be more then one round trip in the combinations.
To make it clearer, a new parameter is added: `roundTripStartDirection`
No reuse of `direction` parameter from now.

`engineMode=4` is still working as expected.
`buildPointsFromCircle` is still the entry point for generation of round trip points.
In the future it would be nice to have a call for BRouter Web like this:
https://brouter.de/brouter-web/#map=14/49.7801/8.7917/standard&lonlats=8.784465,49.779428,r

### DIFF
--- a/brouter-core/src/main/java/btools/router/RoutingContext.java
+++ b/brouter-core/src/main/java/btools/router/RoutingContext.java
@@ -220,6 +220,7 @@ public final class RoutingContext {
   public Integer startDirection;
   public boolean startDirectionValid;
   public boolean forceUseStartDirection;
+  public Integer roundTripStartDirection;
   public Integer roundTripDistance;
   public Integer roundTripDirectionAdd;
   public Integer roundTripPoints;
--- a/brouter-core/src/main/java/btools/router/RoutingEngine.java
+++ b/brouter-core/src/main/java/btools/router/RoutingEngine.java
@@ -171,6 +171,12 @@ public class RoutingEngine extends Thread {
 
     switch (engineMode) {
       case BROUTER_ENGINEMODE_ROUTING:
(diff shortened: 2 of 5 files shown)


devemux86:
@afischerdev Do you have a screenshot to see what the expected result is with a round trip and multiple waypoints?

devemux86:
> In the future it would be nice to have a call for BRouter Web like this:
https://brouter.de/brouter-web/#map=14/49.7801/8.7917/standard&lonlats=8.784465,49.779428,r

We also need the round trip distance and direction in the website parameters.

devemux86:
>To make it clearer, a new parameter is added: `roundTripStartDirection`
No reuse of `direction` parameter from now.

>`engineMode=4` is still working as expected.

@afischerdev This seems like a breaking change, as `direction` will no longer work with `engineMode=4`?

afischerdev:
@devemux86 
Yes, you are right, `direction` recall for engineMode=4
test_rt1

And the screen from the first sample
test_rt

> We also need the round trip distance and direction in the website parameters.

Yes, absolutely.

PR #961, opened 2026-08-06
Alias access=unknown to destination via lookups.dat

julcnx:
Continuation of #956, which got closed by an accidental branch rename on my fork rather than any decision to abandon it. Same commit, same content, just a fresh PR since GitHub doesn't migrate an open cross-fork PR when the head branch is renamed.

See #956 for the full discussion with @afischerdev, this replaces the original "block access=unknown outright" approach with the alias-to-destination approach worked out there.

### DIFF
--- a/brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java
+++ b/brouter-map-creator/src/main/java/btools/mapcreator/OsmCutter.java
@@ -130,6 +130,7 @@ public class OsmCutter extends MapCreatorBase {
     }
 
     if (n.getTagsOrNull() != null) {
+      generatePseudoTags(n.getTagsOrNull());
       int[] lookupData = _expctxNode.createNewLookupData();
       for (Map.Entry<String, String> e : n.getTagsOrNull().entrySet()) {
         _expctxNode.addLookupValue(e.getKey(), e.getValue(), lookupData);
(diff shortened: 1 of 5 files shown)


julcnx:
Done, dropped the lookups.dat change, this PR now only carries the OsmCutter.java rewrite of `access=unknown` to `access=value_unknown`. It's inert on its own until `lookups.dat` has the `value_unknown` alias on the `destination` line in both the way-context and node-context `access` blocks, whichever of this or #951 lands second will make it live.

afischerdev:
Thanks for the remove. 
I'd like to make a generation test for the class update and the new lookups.dat. But this will take a while.

afischerdev:
I made some tests with unknown value.
In the way you suggested it works and the output for this [sample way](https://brouter.de/brouter-web/#map=17/49.86804/8.94162/standard&lonlats=8.940808,49.868238;8.943998,49.867257&profile=hiking-mountain) is
```
...", "highway=track surface=grass estimated_forest_class=5 tracktype=grade4 access=designated motorcar=no motorcycle=no route_hiking_rwn=yes route_hiking_lwn=yes", ...
```
One test was with a poor unknown used as alias
`access;0000006316 destination customers delivery unknown
`
This did not work as expected and delivered `access=unknown` as before.
There are some other unknown values in the lookups.dat. I will remove them, as less is often more.

In the next step, we prepare the test server for the new lookups.dat file.

afischerdev:

afischerdev:
When you now try the sample above on [testserver](https://brouter.de/brouter-test/#map=15/49.8680/8.9416/standard&lonlats=8.940808,49.868238;8.943998,49.867257&profile=hiking-mountain) you will find `access=destination`

PR #963, opened 2026-08-08
route generator

francis-bunker:
Added a loop generator mode using the BROUTER graph.  It creates routes quickly, working directly with the graph data to get good route info.  See demo at https://outdoorroutemap.com/map/ and click the suggest route button

### DIFF
--- a/brouter-core/src/main/java/btools/router/FormatJson.java
+++ b/brouter-core/src/main/java/btools/router/FormatJson.java
@@ -4,8 +4,11 @@ import java.io.BufferedWriter;
 import java.io.StringWriter;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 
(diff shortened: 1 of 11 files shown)


devemux86:
Is this about round trips without a user choice of direction?

devemux86:
The new method in `FormatJson` should work like the existing `format` for routing / round trips.

It should contain all the needed data, e.g. `voicehints`, `messages`, etc.
Essentially copy the existing `format` to a new method and add the extra properties.

afischerdev:
@francis-bunker 
Thanks for your ideas.
I see two major aspects in your development:
- New round-trip routing logic
- Export of more than one route

### New export
Multi route export is a fine idea but as @devemux86 explained there are some problems with it.
We have several export formats. E.g. a combination of way point with the routing hints and a track for the line. It would be hard to find the final route in it.
We might consider the multiple export simply as a presentation format that displays the available route suggestions. A client can use it for this purpose, and when the user selects a route, they receive the standard BRouter export in the selected format.

### New round trip algorithm
I'll take a closer look at that in the next few days. It seems to be using a new pathfinding method.

PR #973, opened 2026-09-06
Improve docker image and workflow

Quadrubo:
The published image is amd64 only, runs as root on a JDK image from 2021 and server.sh hard-codes the JVM options, the port and one thread. This PR builds the image for amd64 and arm64, makes the server settings configurable through the environment and slims the image down.

Builds with these changes have ran on my fork: https://github.com/Quadrubo/brouter/actions/runs/34043044688

## Added

- arm64 image
- `server.sh` reads `JAVA_OPTS`, `PORT` and `MAXTHREADS` from the environment. Defaults stay the same as before.
- README section for the published image and its variables

## Changed

- Runtime image changed to `eclipse-temurin:17-jre-jammy` from the deprecated `openjdk:17.0.1-jdk-slim`. Results in the image size dropping from 410 MB to 265 MB.
- `server.sh` execs java so that java is the container's main process
- Uses the current action versions in the workflow, drops unused install steps and the actions layer cache

## Breaking Changes

- The container runs as user 1000 instead of root. Therefore a mounted `/customprofiles` must be writable by that user if the profile upload is used. Segment files only need to be readable.
- `/profiles2` now only contains `misc/profiles2`. The old image copied the whole `misc` tree including scripts and utils which are unnecessary in the image.
- The image uses JRE instead of JDK
- `server.sh` picks up `JAVA_OPTS` and `PORT` from the environment. Anyone with those variables already set for something else replaces BRouter's defaults. I could rename them to `BROUTER_JAVA_OPTS`, `BROUTER_PORT` and `BROUTER_MAXTHREADS` if you want
- Building the image now requires BuildKIt, this has been the default for Docker since version 23

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,10 +1,5 @@
 name: Docker
 
-# This workflow uses actions that are not certified by GitHub.
-# They are provided by a third-party and are governed by
-# separate terms of service, privacy policy, and support
-# documentation.
-
 on:
   workflow_dispatch:
   schedule:
@@ -17,11 +12,8 @@ on:
     branches: [ "master" ]
 
 env:
-  # Use docker.io for Docker Hub if empty
   REGISTRY: ghcr.io
-  # github.repository as <account>/<repo>
   IMAGE_NAME: ${{ github.repository }}
-
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


afischerdev:
@Quadrubo 

PR #975, opened 2026-09-11
Add via ferrata difficulty limit to hiking profile

### LINKED ISSUE #974: Allow hiking profile to limit via ferrata difficulty
## Problem

`highway=via_ferrata` is available in `lookups.dat`, but `via_ferrata_scale` is not preserved in routing data. The standard `hiking-mountain.brf` profile therefore cannot distinguish an easy ferrata from a difficult one. This also affects ferrata sections mapped as `highway=path` with `via_ferrata_scale=*`.

This is a follow-up to #517: that issue made the highway type visible, but not its documented difficulty scale.

## Proposed behavior

Add a `via_ferrata_scale_limit` profile parameter with values `-1` and `0` through `6`:

- `-1`: exclude all recognizable ferratas (default).
- `0`–`6`: allow only documented OSM grades up to the selected whole grade.
- Treat `N-` as grade `N`, and require limit `N+1` for `N+`.
- Exclude missing and unrecognized grades instead of guessing their meaning.
- Recognize both `highway=via_ferrata` and any way carrying `via_ferrata_scale`.
- Apply the exclusion to flat, uphill, and downhill cost factors.
- Keep the existing costs of allowed ways; the parameter is a ceiling, not a preference.

The documented OSM values are `0` through `6` with optional `+`/`-`. Nonstandard decimals and regional letter grades would remain unknown and fail closed.

## Routing-data compatibility

This requires appending `via_ferrata_scale` to `lookups.dat`, incrementing the minor lookup version, and rebuilding `.rd5` data. Existing lookup indices must stay unchanged.

PR #951 already targets lookup version 11.3, so the final minor-version change and merge order need to be coordinated with it.

## Validation

The implementation has profile-level tests for all whole grades, `+`/`-` boundaries, missing and unknown values, non-ferrata regressions, both highway mappings, both directions, and all three cost factors.

I also rebuilt a small `.rd5` from current OSM data around Alpinpark Gesäuse-Johnsbach:

- Basic: way 687551238, grade `3+`.
- Luis: way 687551236, grade `4+`.
- X-Tream: way 687551237, grade `5`.

With the same endpoints, limit `3` routes over ordinary paths, while limits `4` through `6` use Basic. Starting on the Luis/X-Tream branch, limit `5` permits both and BRouter chooses the lower-cost Luis branch; a start point inside X-Tream verifies that grade `5` becomes routable only at limit `5`.

Would this scope and default behavior be acceptable for `hiking-mountain.brf`?

shadow-identity:
Closes #974

## Summary

- Use the documented `via_ferrata_scale` values supplied by lookup version 11.3.
- Add a configurable maximum ferrata grade to `hiking-mountain.brf`.
- Treat the optional `-` and `+` modifiers as aliases of their whole-grade group.
- Recognize both `highway=via_ferrata` and `highway=path` sections carrying the scale tag.
- Fail closed for missing or unrecognized grades.
- Document the parameter and cover its boundaries with profile-level tests.

## Behavior

The new `via_ferrata_scale_limit` parameter accepts `-1` and whole values from `0` through `6`:

- `-1` excludes all recognizable ferratas and is the default.
- A whole grade is allowed when it is at or below the selected limit.
- `N-`, `N`, and `N+` are treated as the same whole-grade group.
- Missing, nonstandard, and unknown values are excluded rather than guessed.

The limit is applied to `costfactor`, `uphillcostfactor`, and `downhillcostfactor`. Allowed ferrata ways retain their existing profile cost, so the option controls eligibility and does not prefer more difficult variants.

## Compatibility

The required `via_ferrata_scale` entries are managed centrally in PR #951 for lookup version 11.3. As requested in review, this PR does not modify `lookups.dat` and depends on #951. Routing data must be rebuilt before the new tag is available to the profile.

Until #951 reaches `master`, parsing `hiking-mountain.brf` against the current 11.2 lookup fails because the tag is intentionally not present there yet.

## Testing

- `./gradlew clean build --no-daemon` — passed locally with OpenJDK 17 for all modules enabled without an Android SDK, using the exact lookup aliases proposed for #951.
- 15 new `ViaFerrataProfileTest` methods — passed with zero failures and errors.
- Checkstyle and PMD — passed.
- Real OSM end-to-end check — rebuilt `.rd5` and ran 15 routes around Alpinpark Gesäuse-Johnsbach using Basic `3+`, Luis `4+`, and X-Tream `5`.

The real-data matrix confirmed that Basic `3+` becomes eligible at limit `3`, Luis `4+` at limit `4`, and X-Tream `5` at limit `5`. The OSM extract and generated routing artifacts are not committed, in keeping with the developer guide's request not to add test resources.

The local environment had no Android SDK configuration, so the Android app module was not included in the local build; the configured upstream CI build remains the Android verification point.


afischerdev:
@shadow-identity 
Please remove the lookups.dat from your PR. The lookups.dat changes for all are collected in #951

shadow-identity:
Thanks. I removed `lookups.dat` from this PR and adapted the profile, tests, and documentation to the grouped whole-grade aliases proposed for #951. The PR now depends on #951.

Modifiers are handled as aliases of the corresponding whole grade, and the grade handling corresponds to the lookup groups in your list.

afischerdev:
we use teh new lookups.dat v11.3 for new data on test server
Please have a look at this [route](https://brouter.de/brouter-test/#map=18/47.54421/14.58383/standard,Waymarked_Trails-Hiking&lonlats=14.582946,47.544718;14.584391,47.543994). When you use `assign processUnusedTags 1` in the profile you will see the new `via_ferrata_scale` values.

PR #977, opened 2026-09-19
Make HTTP server memory class configurable

flamezor:
While self-hosting BRouter I ran into `memory limit reached` on some long and complex routes.

`ServerHandler.readRoutingContext()` currently hardcodes the memory class:

```java
rc.memoryclass = 128;
```

This change makes it configurable through the `memoryClass` JVM system property while keeping `128` as the default:

```java
rc.memoryclass = Integer.getInteger("memoryClass", 128);
```

This keeps the current behavior unchanged unless the property is explicitly set.

For example:

```text
-DmemoryClass=256
```

### Testing

I built **BRouter 1.7.11-beta** with this change and tested the same long-distance route using:

- the same routing profile
- the same segment data
- the same Docker image
- a 2 GB JVM heap

With no `memoryClass` property specified, the request terminated with:

```text
memory limit reached
```
The same happened explicitly with:

```text
-DmemoryClass=128
```

With:

```text
-DmemoryClass=256
```

the same request completed successfully after ~9 min and returned:

- a 4.58 MB GeoJSON response
- a route of about 2354 km

The default behavior remains unchanged when the `memoryClass` property is not specified.


afischerdev:
Please add a short description in the `misc/scripts/standalone/server.sh` how to use this new parameter.
But don't set this as default.

